PR #2, opened 2017-03-28
Hitting barracks

GoldenGuy:
Arent this should be there? If other classes can damage it, why builder cant?

### DIFF
--- a/Entities/Characters/Builder/BuilderHittable.as
+++ b/Entities/Characters/Builder/BuilderHittable.as
@@ -20,6 +20,7 @@ const string[] builder_alwayshit =
 	"building",
 	"quarters",
 	"storage",
+	"barracks",	
 };
 
 //fragments of names, for semi-tolerant matching


fzzle:
The barracks is obsolete and shouldn't be taken into consideration (in hacks like these).

GoldenGuy:
I though i saw some maps with barracks. Ok

furai:
Still I don't think there's anything wrong with adding it for the consistency sake. @1bardesign what do you think?

1bardesign:
Adding the barracks would add another string comparison, but that's about it. It's not crazy expensive but it's not free.

I agree with fuzzle that it's an obsolete building not found in vanilla gamemodes - modders using it in a gamemode can add it pretty easily to the builder always hit code.

I don't have a strong opinion either way though. The builder code is already laden with hacks.

1bardesign:
Gonna reject this for now I'm afraid - as said above people who need it for mods can add it themselves. I don't see much direct harm in adding it but I'd rather not pull in yet another slight hit to performance if I can help it, sorry!

PR #3, opened 2017-03-31
Mine fix

fzzle:
Fixes the notorious frame issues:
- Picking up seemingly primed mines.
- No frame set on initialization (unprimed on join).


fzzle:
@1bardesign, it wasn't the first time I fixed it. Somehow the issue found its way back. 😉

PR #5, opened 2017-05-03
fixed amount spelling from ammount to amount

calvinloveland:
Really just wanted to get comfortable with pull requests.
Shouldn't be a controversial fix.


1bardesign:
(it doesn't look like I've been set up with write permissions github-side, so it might take a while for the changes to show up as merged)

PR #8, opened 2017-07-01
added support for dlc heads and removed runnerhead.as include in quarters.as

makmoud98:
this just simplifies the code. my main motive for this is to be able to add hooks like onInit and onCommand into runnerhead. before this change it was not possible because the compiler thinks there is already a function with the same name and parameter present (since runnerhead is included at the top). this change will allow me to update the playerheads mod without having to make changes to the quarters script. this helps to keep the playerheads mod to be as compatible with other mods as much as possible. thanks for reading

### DIFF
--- a/Entities/Industry/CTFShops/Quarters/Quarters.as
+++ b/Entities/Industry/CTFShops/Quarters/Quarters.as
@@ -1,4 +1,4 @@
-﻿// Quarters.as
+// Quarters.as
 
 #include "Requirements.as";
 #include "ShopCommon.as";
@@ -6,7 +6,6 @@
 #include "CheckSpam.as";
 #include "StandardControlsCommon.as";
 #include "CTFShopCommon.as";
-#include "RunnerHead.as";
 
 s32 cost_beer = 5;
 s32 cost_meal = 10;
@@ -268,6 +267,7 @@ void onAttach(CBlob@ this, CBlob@ attached, AttachmentPoint@ attachedPoint)
 	attached.AddScript("WakeOnHit.as");
 
 	string texName = default_head_path;
+	u16 frame = 0;
 	CSprite@ attached_sprite = attached.getSprite();
 	if (attached_sprite !is null && getNet().isClient())
 	{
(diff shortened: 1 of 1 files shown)


fzzle:
Nice modification. I wanted to do this for some time. Implementation scripts (w/ hooks) shouldn't be used as common scripts (included). It also removes the head gib created for quarters  (`onDie` in RunnerHead.as).

Entities/Industry/CTFShops/Quarters/Quarters.as
```diff
@@ -268,6 +266,7 @@ void onAttach(CBlob@ this, CBlob@ attached, AttachmentPoint@ attachedPoint)
 	attached.AddScript("WakeOnHit.as");
 
 	string texName = default_head_path;
```
fzzle: Compile error: `default_head_path` is still used.
Could set it to `'Heads.png'`.

Entities/Industry/CTFShops/Quarters/Quarters.as
```diff
@@ -277,6 +276,7 @@ void onAttach(CBlob@ this, CBlob@ attached, AttachmentPoint@ attachedPoint)
 		if (head !is null)
 		{
 			texName = head.getFilename();
+			frame = head.getFrame();
```
fzzle: `head.getFrame();` returns the frame of the spritesheet currently used in the animation. If I idle the spritesheet frame is offset + 0. However, if I (for example) slash the moment I enter the bed the spritesheet frame is offset + 1. Then `anim.AddFrame(frame + 2)` adds the spritesheet frame at offset + 3 (a blank frame).

Replace line 279:
```
Animation@ animation = head.getAnimation('default');
if (animation !is null)
{
  frame = animation.getFrame(2);
}
```

Replace line 299:
```
anim.AddFrame(frame);
```

1bardesign:
Hey; I'd love to go through and merge/close the current batch of pull requests next Monday (ie in a week) - if you could get this ready to go with the required changes as a PR I'd be happy to merge it in 👍

1bardesign:
Closing this now; went with https://github.com/transhumandesign/kag-base/pull/13 instead

PR #9, opened 2017-07-02
Minor changes in CTFshops

HkgStudio:
Deleted some useless stuff.


1bardesign:
This seems to be just testing out the PR system as the changes aren't extensive.

In my eyes these changes aren't warranted - the removed code is to allow other scripts (or a tcpr app, or the server owner via rcon) to modify which config is loaded by writing a single rules var; so it's not useless and I'd rather not merge it in; not only that but it occurs in more shops than just the quarters!

It's good to have it scrutinised though - thanks for your time.

PR #10, opened 2017-07-09
Water nerf

AsuMagic:
This is a fairly minor nerf: Water bombs and water arrows now raycasts (through tiles but not blobs) and splashes in a radius rather than in a box, but the radius is the same. This should be sort of more coherent with the rest and would be a little less painful to fight against.

There is a minor cleanup in Bomb.as but it should be rewritten at some point anyway.


AsuMagic:
Looks like (and according to fzzle) raycast checks would be quite broken because waterbombs could explode within tiles. PR should be closed for now.

fzzle:
There are side-effects of raycasting. These can be prevented by the calculation of a better source position.

1bardesign:
No worries - revisiting this kind of thing is valuable even if it doesn't get merged in.

Sidenote: Things like `Bomb.as` being messy and not making a huge amount of sense are just part of the legacy of KAG - a rewrite would be worth doing if it provided real benefits but if it just provides the same functionality in a nicer wrapper it's not a great use of time; if you want to do it by all means, but just be mindful of spending too much time on it as it can be discouraging.

PR #14, opened 2017-08-06
Community ctf maps and tth map added. Mapcycles updated.

Panky123:
I put some explanations of my actions in the description instead of the top. Sorry.


1bardesign:
Accepting this and the other maps PR https://github.com/transhumandesign/kag-base/pull/15 now - we need to standardise the maps naming though and clear out the existing archive "soon".

PR #15, opened 2017-08-06
Accepted TDM maps

fzzle:
Accepted community TDM maps.

I decided not to put them into a month / year folder as Git keeps track of it and I find it inconsistent with how it's been handled beforehand. If this request is accepted I'll refractor all the map folders and cycle files to follow the same convention.


1bardesign:
Accepting this and the other maps PR https://github.com/transhumandesign/kag-base/pull/14 now - we need to standardise the maps naming though and clear out the existing archive "soon".

PR #17, opened 2017-08-15
More translations (#1)

AsuMagic:
This is some initial work to bring in a much better translation support to KAG. Some various gamemode things have been translated, and a large chunk of the ingame help has been translated as well. It also removes the legacy `.irrst` translations.
This compiles and should generally work well, but considering the large amount of changes and code affected this will need to be throughfully tested.
All of the new translation support comes with french translations in. It might need to be reviewed by someone else than me but I think it should be good quality.


1bardesign:
I'll get onto reviewing this after I've sorted out the internal repo situation :) if you've got more to add to this PR keep the commits coming, would love to get the translation stuff into the mainstream, or at least to the point it's worth having GUI for it :)

PR #20, opened 2017-09-24
[Delete] Old Greg

Enderzilla747:
Forgot to delete old Greg when I suggested my new one, oops.


1bardesign:
Please don't submit multiple pull requests for the same issue or feature - add more commits to the original.

I'm not sure I'm keen on pulling either of these in wholesale, please get in touch on discord or the forums early next week so we can go through why you made various changes and why you feel they're actually necessary.
The gargoyles are meant to be stone gargoyles, not fleshy zombies.

PR #25, opened 2017-10-13
Update KickAFK to move people with kick immunity to spectator

Mazey:
Instead of leaving people with kick immunity, like admins, in-game, move them to spectator.


Mazey:
I completely overlooked that. Cheers

1bardesign:
Looks good, definitely better than having them afk on a team.

PR #26, opened 2017-10-13
Slight modability improvement for custom blocks

ANybakk:
Just a very small modability improvement, that lets modders not just create custom blocks that serve no purpose, but can be harvested by a builder's pickaxe or drill. Without messing with vanilla scripts needlessly.

### DIFF
--- a/Entities/Materials/MaterialCommon.as
+++ b/Entities/Materials/MaterialCommon.as
@@ -7,6 +7,8 @@
 // the server-side. An example can
 // be found in Log.as
 
+#include "CustomBlocks.as";
+
 namespace Material
 {
   const float MERGE_RADIUS = 20.0f;
@@ -224,6 +226,10 @@ namespace Material
     else if (map.isTileGold(type))
     {
       createFor(this, 'mat_gold', 4 * damage);
+    }
+    else
+    {
+      MaterialFromCustomTile(this, type, damage);
     }
   }
 }
--- a/Scripts/MapLoaders/CustomBlocks.as
+++ b/Scripts/MapLoaders/CustomBlocks.as
@@ -9,6 +9,8 @@
  *		Note: don't modify this file directly, do it in a mod!
  */
 
+//#include MaterialCommon.as;
+
 namespace CMap
 {
 	enum CustomTiles
@@ -18,7 +20,17 @@ namespace CMap
 	};
 };
 
+//Map loading
 void HandleCustomTile(CMap@ map, int offset, SColor pixel)
 {
 	//change this in your mod
+}
+
+//Harvesting
+void MaterialFromCustomTile(CBlob@ this, uint16 &in type, float &in damage)
+{
+  if (type == CMap::tile_whatever)
+  {
+    //Material::createFor(this, 'mat_whatever', damage);
+  }
 }


ANybakk:
While working on some custom tiles, I might also have stumbled upon a bug. Custom tiles seem to fail the isTileSolid in MaterialsCommon, even though the flags have been set...

fzzle:
I see 3 issues:
- Nested includes are broken (esp. in mods) and this takes use of those.
- It's not extensible w/ modules: If one mod takes use of it, another can't.
- The script-side shouldn't contain content not used in the base game.

It's something that should be implemented engine-side. `Material::fromTile` is the sole function in the `Material` namespace that still contains a "hack" and I've planned to eliminate it.

1bardesign:
Leaving this out for now as a result of the criticisms above. We do welcome improvements to this area though, as it's definitely a stumbling block for a lot of modders!

About the solidity - feel free to open an issue to discuss this further. It's a bug in that the overload based entirely on tile type is used, which only works with vanilla tiles because the tiletype itself contains no flag information. The function would need a map offset or position to get the tile flag (or there'd need to be some engine-side provision for setting the flags expected of a given tile type, which sounds like it'd probably be useful as well).

ANybakk:
Fair enough, I guess. This "mechanism" of adding custom tiles was already there though. It just didn't seem all that useful as it was, since it would basically just be decorative. It was added to make modding easier, only it really isn't. You only get so far.

The second argument is no less true for overriding the whole MaterialCommon.as file. The difference is missing out on future updates and fixes when you do things like that, and possibly introduce bugs because you end up with a mix of new and old code. A lot of KAG modding suffers from this. It is slightly better to override files with "isolated" custom code, as is the case here. I think it's acceptable, at least for simple mods. Mods could strictly allow mods be built upon itself, if the mod writer makes similar overrides possible themselves.

Modularity is a concern. But it was never really considered, it seems. Mods can't really co-exist. That might very well change with all the improvements that are being worked on, but "we" don't know when that will be available, and it's going to break a lot of older mods. This can also be deterring when it comes to modding stuff now. I worry that the time I spend in my project, which is slow moving, will just end up being incompatible when I'm getting done. I have to say, I question the decision of doing major changes like this (from my impression), at this time, without basically releasing a new game.

This is all a little frustrating, to be honest. I've earlier complained a little about the lack of modularity, among other things, and was basically answered with "it works well enough, so just accept it". So I did, and now you seem to have changed your mind. At a time when the game is another few years older and the player base even smaller. I fear the train has left the station, and the efforts are too late, and might very well end up just causing more harm than good.

This particular small "fix" might not be doable. That's okay, but if it's not relevant enough now, it's certainly not going to be later, when all this has changed. Might as well close it.

1bardesign:
Please don't be so quick to jump to conclusions about what we have or haven't "decided".

Mods can definitely co-exist; a lot of smaller mods just add a single entity which can be spawned in however is most appropriate, the PlayerHeads mod just modifies how the runner head script works, and so on - but I understand your frustrations about the modularity comment in particular.

My primary concern with accepting this is really just @fzzle's first point, the nested includes issue.

`CustomBlocks.as` would need a lot more support to really be anything more than just decorative. Adding materials support is welcome in my eyes; more work handled for the modder is nice - but it would be good to talk to fuzzle about what in particular he has planned for the materials stuff in the near term so that this doesn't add to the maintenance workload or potentially seriously break mods when something changes.

It's particularly important for something meant to be a "framework" for modders just getting off the ground that their mods don't mysteriously stop working; I hope you can understand us behaving conservatively around this kind of thing.

PR #28, opened 2017-10-16
Added a new russian translation

Diprog:
Unfinished as all translations


1bardesign:
Thank you Diprog! Are you planning on trying to maintain this as translation support gets extended? If so, we can try to keep you in the loop as to what has changed each version.

Diprog:
Yep, that's what I wanna do.

PR #29, opened 2017-10-17
Loader utilities fix

ANybakk:
LoaderUtilities.as contains a couple of handlers that gets in the way if you write your own map script, use PNGLoader (or a derivate of it) and want to write your own handlers for those specific hooks. You will get this error message: "Script: A function with the same name and parameters already exists".

This simple fix basically just places the include directive in the individual map loading scripts instead of the file containing the class declaration.

### DIFF
--- a/Entities/Materials/MaterialCommon.as
+++ b/Entities/Materials/MaterialCommon.as
@@ -7,6 +7,8 @@
 // the server-side. An example can
 // be found in Log.as
 
+#include "CustomBlocks.as";
+
 namespace Material
 {
   const float MERGE_RADIUS = 20.0f;
@@ -224,6 +226,10 @@ namespace Material
     else if (map.isTileGold(type))
     {
       createFor(this, 'mat_gold', 4 * damage);
+    }
+    else
+    {
+      MaterialFromCustomTile(this, type, damage);
     }
   }
 }
--- a/Entities/Materials/MaterialCommon.as
+++ b/Entities/Materials/MaterialCommon.as
@@ -7,8 +7,6 @@
 // the server-side. An example can
 // be found in Log.as
 
-#include "CustomBlocks.as";
-
 namespace Material
 {
   const float MERGE_RADIUS = 20.0f;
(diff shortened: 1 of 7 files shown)


ANybakk:
In hindsight, what appeared to be a very simple fix, isn't. I just commented out the include in my mod, but I realized I'll miss out on that DummyTile stuff (whatever it is).

I would suggest coming up with some other way to get rid of this possible conflict, either by including another script mid-code that can be overridden (though form last discussion, you didn't want that kind of solutions), an inheritance approach or maybe an extra RegisterFileExtensionScript call (or similar) in DefaultLoaders.as.

1bardesign:
I'm not sure why a modder writing their own map script couldn't just provide their own modification removing the includes as necessary? You aren't forced to use the vanilla files, and taking a copy and then modifying it seems like a pretty trivial "fix" that doesn't require any changes in vanilla?

What am I missing?

Sidenote: The dummy tile stuff is a part-way finished system for putting some of the blob block's collisions into the tilemap, but I think Skinney more or less abandoned the implementation part way through and afaik they don't actually do anything currently.

1bardesign:
> Good modability comes when implementation doesn't matter to the modder.

I agree with the sentiment, but we simply don't have this. The way it is done inside the engine is exactly the way it is exposed to the mods - we have an enormous hit handler, tile swapping code, sounds played on tile change and so on. You seem to repeatedly expect a miracle of modder-friendly engineering; while it would be nice, we have just exposed what we ourselves work with engine-side.

I understand the complaint at it's heart comes from wanting to accomplish what seems like something simple, but it's not how the game has turned out, and in the past whenever we _have_ put in extra work to make something modder-friendly or easier (TDM config, shops cost configs, tags to re-enable legacy behaviour, modular animal scripts) it has been hugely under-utilised or not used at all.

We're at the point where it's easier to help people create the modding conveniences that they want rather than build anything like this pre-emptively. If you want help building out a system like what you've described for use by you and other modders, I'd be happy to help.

Issue #16, opened 2017-08-13
Translations


AsuMagic:
I have discussed translation support a bit tonight with Fuzzle and decided to do some work to expand the translation support in KAG.

I believe it would bring some good advantages. It is easier for non-native people to understand (= potentially less retention). It also would allow making some official translations appear on Steam.
It could be possible to translate the steam shop page into other languages as well, I can translate it to french if you want.

I've done some work tonight to extend the translation support in KAG, see the changes on my [repo](https://github.com/transhumandesign/kag-base/compare/master...AsuMagic:master) (forgot to take it on a branch, I don't have time fixing it up tonight).  
Some changes were less trivial than others. As for help support, e.g., I've taken the easy way and made a somewhat conservative way to perform the translation automatically (some helps translates manually). I didn't extensively try it, though, but it looks decent so far.

Though, I have a few questions:
- Irrlicht GUI: Is `Locales/zh_**.irrst` supposed to work in Release? I didn't ever manage to make it spit out translated text (with a `zh_FR.irrst` file with the locale set as `fr` in autoconfig or with `zh_EN.irrst` with the locale set as `en` in autoconfig). I was writing it this way:
```json
<entry id="PLAY">JOUER</entry>
<entry id="Create Account">Créer un compte</entry>
```
It also contains some outdated strings, it would be nice to make sure everything is up to be translatable not to miss out some stuff.
- Is having partial translations OK? The polish translation seems to translate just a small batch of strings. I would rather progressively push up the newer translation support rather than all at once (to avoid some git merging nightmare considering how much scripts are affected). I suggest an option in the settings to select the locale and mark non-english as "experimental" support while fallbacking all current clients to EN.
- How possible could it be to have blob inventory names/etc translated by the engine? This would ease out the translation support fr sure, and I doubt there are scripts that relies on the inventory name (hopefully...)
- Could it be possible to avoid ambiguous translations in the json file e.g. by allowing packing specific things in objects? For example in a case such as:
```json
"scoreboard":
{
    "kills": "meurtres", // noun
    "deaths": "morts",
    // ...
}

"killfeed":
{
    "kills": "a tué" // verb
}```
It would be extra neat to have some way to way to format the text in the translations (e.g. `{0} has killed {1}`) since it might be complicated to translate some stuff with the same ordering as in english. I could probably implement it in angelscript directly though.

1bardesign:
The XML stuff is deprecated and from classic.

[...]

PR #34, opened 2017-10-18
[modified] wrapped all strings into getTranslatedString functions

Diprog:
more coming soon


1bardesign:
This should not necessarily be done _everywhere_.

The one that sticks out is that requirements (from `Requirements.as`) should probably be translated for rendering, not for syncing over the net (as that might even use the server locale, not the player locale?). 

Same goes for the `rules.SetGlobalMessage("The princess died!");` - this will be done serverside, and in the server locale. We can make the rules global message be translated if you need, please just raise an issue for it after confirming that it isn't already, or maybe put it on the existing translations issue https://github.com/transhumandesign/kag-base/issues/16

Diprog:
Oh yeah, I'll be more attentive next time (about requirements).
And I don't see any problem in translating global messages for single-player game modes

1bardesign:
It would be better for the engine to provide the translation for all players, and `Challenge_Save.as` is the multiplayer princess-saving challenge mode, not the single-player Save The Princess mode.

Diprog:
Ow, thx. I'll fix everything tomorrow.

1bardesign:
I have a few concerns about this; it's coming along well but some things are translated at points that are maybe not the best;

- `BuildBlock`s could all be translated in one place (when the menu is created) rather than as every block is defined, which would make mods automatically translation-compatible if they have matching strings.
- `CreateSign` could just translate the string passed in rather than every create sign call having a separate translate call, or (better) the sign could translate as needed (when rendering, if no cached string is found) so that the sign is not in server locale. I'm not sure if signs are used much on any custom servers but it would still probably be nicer?
- As I said, next build `rules.SetGlobalMessage` will display translated version per-client rather than translating serverside. The global rules message is synchronised from server to client.

Otherwise it seems pretty good! If you can fix these I'll merge this soon.

Diprog:
BuildBlocks are already translated in build menu code, global messages are already fixed too. I'll change CreateSign soon

1bardesign:
I can see a few places where this has missed out one or two places but I'll get it merged for now and you can pull from master again for my fixes. Thanks for your work on improving translation support!

Diprog:
Yeah, some fixes got reverted somehow, so I didn't notice that. Thank you too.

Diprog:
Oh yeah, replaced ыemicolons with commas -_-

1bardesign:
I can see a few places where this has missed out one or two places but I'll get it merged for now and you can pull from master again for my fixes. Thanks for your work on improving translation support!

1bardesign:
(merged in - please base any changes on what's currently in master)

Just a sidenote, I notice a lot of "accidentally" in your commits - if you're editing things separately rather than inside a "working" kag folder it might cause trouble like this. I'd recommend working from inside a working install so the files you commit are the same files you are testing.

Using a GUI client like sourcetree, gitkraken, or tower (or gitk or..) can help catch these errors if you're not already.

Either way, try to check your diffs before hitting commit!

PR #35, opened 2017-10-18
[dev][modified] fixed scripts for angelscript update

AsuMagic:
The planned Angelscript update breaks a few minor things that can be easily fixed.

One of the fixes is a useless copy constructor usage that causes a compile error.  
The other fixes are related to callback handles. Add the assign operator when assigning a function pointer.

These changes should be compatible with the older Angelscript version.


Diprog:
`SColor(team.color)`
Convert SColor to SColor? lol, why was it here?

1bardesign:
>Convert SColor to SColor? lol, why was it here?

It's a useful thing to be able to take a copy, but in this case it wasn't actually accomplishing anything. Potentially vestigial (maybe it used to modify it) or maybe just a mistake :)

PR #38, opened 2017-10-23
[modified] pickup priority tweaks

tru0067:
-corpses made super low priority (used to still have decent priority when extremely close)
-added builder materials priority (gold > stone > wood)
-added saw to combat items (will ensure saw > seed for wood farms)


tru0067:
New to kag-coding so just easing myself in with some small tweaks.
Future plans for pick-up priorities:
- mat stacks prioritised by stack size (so you pick up the 250 stone instead of the 11).

1bardesign:
Looks ok 👍 careful with branch naming if you can (`patch-1` is perhaps less helpful than `pickup-priority-tweaks`), but this was a nice easy pull request! We'll see how the saw and corpses go.

PR #39, opened 2017-10-23
[added] material stack pick-up priority

tru0067:
- material stack pickup priority based on how full the stack is (stacks which are more full have higher priority)

(hope I did the branch naming properly this time)


1bardesign:
Yep that branch naming is fine, nice 😄 Thanks for contributing!

PR #42, opened 2017-10-24
Russian: fully translated tutorials (basics, CTF & TTH) and fixes + changes.

Diprog:
Thanks to @BarsukEughen for translating Tutorial03.as and for some fixes and changes.
I'm so tired, I think there are some misstypes etc.
I go to bed.


1bardesign:
No rush, I will be working on getting this build out next week so there's plenty of time to check 😉

Diprog:
@1bardesign btw, do you want Trench Run to be translated too?

1bardesign:
@Diprog no, we aren't maintaining Trench Run beyond any huge breaking changes (I need to update the EU Server IP address for example) as there's no income from it or community, so it's pointless work. It also isn't up to date with most recent engine changes so it doesn't have the updated translation code. I would focus your efforts on KAG.

PR #44, opened 2017-10-26
Organized costs and descriptions

Diprog:
Descriptions later
Edit: they are updated too now


Diprog:
I don't know where to move all these configs. And I think we need to rename "CTFShopCosts.cfg" to "CTFCosts.cfg". 
Also filled bucket is missing, you can easily add it back.

And btw, removed useless WARCosts.as includes from some files that never used that include.

1bardesign:
There are conflicts as part of this, you will need to pull master and carefully sort them out before this will be considered again.

Please don't rename files off-hand - because a new name might be "better" semantically or whatever, but if a mod has included or modified that file it will be broken by the change.

@fzzle please be more constructive about review; just leaving a thumbs down is demotivational and noncommunicative.

Diprog:
> if a mod has included or modified that file it will be broken by the change

A modder will just have to edit some files. Same thing happened with materials update earlier.

1bardesign:
It'd be possible to reorder descriptions if we had an enum index into it and used that everywhere, but then we may as well just use string literals for sure, haha.

Materials update also caused problems with mods 🙂 we'd like to have compatibility notes and ideally a way for people to upgrade their mods easily, or pull in old versions of the files they need without having to learn about git (as many modders do not)

I'll look at merging this but we need to be clear about compatibility issues raised and communicate that to modders, if possible well ahead of it being released. That last part isn't really possible if I'm going to do an update today. We'll see if it makes it in.

PR #47, opened 2017-10-30
[modified] improved mechanisms performance at idle

AsuMagic:
This seems to improve the `MechanismServer.as` performance, especially at idle.

Tested in solo Sandbox, at idle:  
`0.00372s` average -> `0.00069s`


1bardesign:
Hrm, interesting that this is so much faster but I guess it means the copy is done native, and there's a lot of chunks to update. Looks good.

PR #48, opened 2017-10-30
Improve map loading performance

AsuMagic:


1bardesign:
This is fairly extensive.

I'm not super happy on the move to raw `u32`s instead of `SColor`s for the colours, particularly as the variables are still called `color_[description]`. Is this just because AS switches won't allow cases that access a member? AS docs seem to suggest that const values that can be determined at compile time can be used, but I'm not sure if this applies to members of constant structs.

If that is so, what was the motivation for moving to switch statements in the first place? The vast majority of if/else hits would be super early (`color_tile_ground` and friends) with that sky early-out added, and the addition of fall-through and the `//TODO verify` makes me a little nervous to merge this before today's update.

It has ended up with misleading examples at the top of the resulting `LoaderColors.as` as well; a u32 cannot be constructed from 4 ints.

fzzle:
`switch` performs far better. Doesn't matter much in this situation. I dislike this entire hack.

1bardesign:
@fzzle what is the hack; this PR, or the PNGLoader, or the modding related stuff?

What would you change, and why? Would like to get the outstanding PRs closed one way or the other.

AsuMagic:
Added a commit that changes a bunch of things:

- Colors are put in an enum instead of `const u32`. I did ask on AS forums about the const member thing and it indeed won't work with `somecolor.color` as `switch` `case`s, because they can't be determined at compile-time.
- Mods will have to copy out their `BasePNGLoader.as` and `LoaderColors.as` dependencies if they have custom scripts depending on them, including shiprekt. Though, the vanilla gamemode-specific loaders were updated correctly. This is because for one the standard map colors are put into an enum now and renamed and because the `handlePixel` signature has changed (every method that is overriden in AS should be marked `override` btw, it looks like this version supports it).
- I don't know if you'll like it better or not, but most color cases are oneliners now, and aligned with other lines as much as possible. The goal being to make it easier to scroll through the file and change things quickly.

And, yeah, it makes it much faster. So much that the `handlePixel` part takes near no measurable time when loading large maps.

fzzle:
Horrendous code.

PR #50, opened 2017-10-31
Translatable class grid menus, generic buttons and shop description

Diprog:
Some more problems with icons
screen-17-10-31-17-49-43
screen-17-10-31-17-55-52


1bardesign:
You need to use the test build if you're not already to see these properly :)

Diprog:
Can I get it without any applications? I don't want to be a tester.

1bardesign:
You need to apply but you don't need to be constantly-available tester. Furai is basically accepting everyone now. I would prefer it to be fully-open but for the moment, we have to do applications.

PR #53, opened 2017-11-01
Get resupplies in CTF without an inventory check

Diprog:
Really, why do we need this? I can just put out my materials, get new ones from a resupply building/tent, and then put in my previous materials.


Diprog:
Asu is working on an overhaul.

PR #59, opened 2017-11-16
Material refactor

fzzle:
More / improved comments.
Semi-colons removed from preprocessor directives.
Underscores removed between type and name in conf's.


1bardesign:
Looks good. I would still strongly prefer you use 4 spaces or 1 tab for indentation for consistency with the rest of the codebase, but I'm not going to hold the PR for it.

PR #60, opened 2017-11-18
[updated] french translation

AsuMagic:
I didn't proof read this a lot but this:

- Has the new engine side translations for menus, etc
- Has some strings put in "in advance", in particular ones that are not working engine side yet (tooltips in settings for example)
- Makes the french translation synced with diprog's translation changes

Since the whole file was remade there might be some things gone and changed

### DIFF
--- a/Locales/French_fr.json
+++ b/Locales/French_fr.json
@@ -2,100 +2,282 @@
 // This file contains translation for various text in the game.
 
 // INSTRUCTIONS:
-// The file format is JSON.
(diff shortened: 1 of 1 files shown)


AsuMagic:
It is not in sync with the russian translation PRs, only to the test build russian translation.  
[This is the reference file](http://wyvup.com/?c=A2XJD1p) (so I can diff it out later).

1bardesign:
ok - there's some more stuff in the PRs probably worth catching but it's an improvement in terms of coverage anyway for sure 👍

PR #64, opened 2017-12-07
Accepted TDM maps december 2017

Ni-ice:
All the maps, and the new mapcycle.


terracraft321:
Seems to be in order.

1bardesign:
Cheers for handling this! In future if you're handling the PRs for more TDM maps, it'd be great to have a link to the forum discussion in the PR text as well. You'll get a handle on committing and keeping a more clean history with practice 🙂 

Thanks again for your contribution!

PR #65, opened 2017-12-07
Optimized Images

terracraft321:
Smaller file size, for smaller download and space saving in general.
I wouldn't recommend using the maps unless they work and act normally.
Total -- 3,155.22kb -> 1,792.59kb (43.19%)


1bardesign:
Interesting idea. Honestly the images are a pretty trivial amount of the size but it's a pretty obvious win as far as I can see.

Will imgbot do this periodically? And will it be against our repo or your fork?

1bardesign:
@terracraft321 Would you be able to run this against the recently added TDM maps so there's not 2 separate PRs for the same thing?

terracraft321:
@1bardesign Yes, imgbot does it always when new images are added to the repo.

terracraft321:
And i'll re-run it.

PR #77, opened 2018-02-21
[modified] translated servers browser's stuff

Diprog:


1bardesign:
Thank you @Diprog will get this merged asap along with the new PL stuff :) were there any issues with the translation facilities or it all worked?

And the other thing we're missing is login window, right?

Diprog:
screen-18-02-22-06-44-56
Not able to translate this list.
And yeah, we're missing thelogin window.

1bardesign:
Hrm, you're right. I've just checked though and that list will be a pain to add translation to - the list elements are used for filtering the gamemode directly, so if we translate them the filter will break - very annoying.

I will make a note of it but prioritise the login window when I am next working on translation support.

PR #81, opened 2018-03-06
Improve pickup priorities

AsuMagic:


AsuMagic:
Pickup logic has been improved for "smarter" priorities: Healing items have an higher priority when the player has non-full health. Some military items such as bombs have lowered priority when thrown by a teammate. Kegs have priority over bombs, archer arrows are less prioritized than healing items, etc.

Distance is barely taken into account anymore. The direction the player faces no more influences pickup. I removed recent pickups lists, so spamming `C` doesnt influence what item is picked up. IMO those last three were bringing more trouble than advantages.

Those changes allows for (hopefully) much more reliable pickup when tapping `C`, but that seriously affected actually wanting to pick up lower priority items, so I made it so hovering items allows for precisely picking up items you want.

Should require a bit of testing and there are probably ways to improve it further. Thoughts?

1bardesign:
Folks seem in favour so let's try it 👍

tru0067:
I believe the recent pick-ups list was intended to prevent you picking up an item that you had recently picked up and didn't actually want, likely to prevent the annoyance of being physically incapable of picking anything else other than xxx up when xxx is in your pick-up area. Perhaps a recent throws list is more precise, so that if you throw away an item (presumably because you don't want it) you won't pick it up again (perhaps don't apply this to explosives or other "military items").

1bardesign:
I think that's in scope for a separate PR. If you have access to the test build, these changes are already there and ready for testing.

If you're perhaps interested in working on such a "thrown" list, I'd talk it through with folks in `#development` on discord first as there was a few reservations about such a list simply increasing the complexity of such a system - it may be better to take another avenue, like a long-hold "what did you really mean to pick up" menu.

Also, a fairly simple workaround in practice for the "recently thrown" stuff is to actually physically toss unwanted items further away so they are no longer treated as a priority.

PR #82, opened 2018-03-07
Fixed climbing pass-through structures (such as enemy trapblocks)

AsuMagic:
The issue was that climbing only depended on whether the blob tile was static or not.
When climbing enemy trapblocks, you could walljump with them because they were going static at some point.

However, there was no check to make sure the climber blob can collide with the tile blob in question. This adds an optional `CBlob@` parameter to `RunnerMovement.as` allowing to make sure that the blob can collide with the tile blob.


1bardesign:
Hopefully this isn't too drastically expensive 🙂 runner movement is already pretty costly. might make sense to do it just for "myplayer" or on server for bots and just live with the slight desync? Still a fine fix for now.

PR #83, opened 2018-03-07
Fix endless TTH games

AsuMagic:
See this [bug report](https://forum.thd.vg/threads/2311-never-ending-tth-game.27231/).

The game now checks when warmup finishes for the following cases:
- When one team caps a hall but not the other, the former wins.
- When no team caps halls, it's a tie.


1bardesign:
👍 thanks for engaging with the various minor reports

PR #91, opened 2018-04-09
[modified] ignore item aim when tapping pickup

AsuMagic:
This should make it hopefully more reliable to tap [C] during a battle so it doesn't pickup aimed items in an undesired way by introducing a 15 ticks (0.5 second) delay before aim is taken into account for pickup.
It might be worth merging all the tap logic for all keys in one place if the logic happens to be copied across scripts maybe?


1bardesign:
Looks fine - thanks for handling this quickly. You might want to tell that angry guy on steam that it's been handled 👍

PR #111, opened 2018-05-14
[fixed] hopefully more conservative tth wins, fixes broken tth single…

AsuMagic:
…player tutorial

Should be tested a bit with regular games.
Ties are removed as they were not really helpful anyway. When at least a team is empty the game does not finish.


1bardesign:
Will be merged, but the behaviour this gave was suboptimal as well. What I settled on is in master now. Took the opportunity for some refactoring.

PR #112, opened 2018-05-15
Add property that allows changing buy offset

Mazey:
In your shop blob add `this.set_bool("shop offset is buy offset", true);` to change the "shop location" to wherever the shop offset is, this way you can create a shop outside of the blob's radius. For example, setting this property to true in war base you can actually buy things in the workbench. (with support for setFacingLeft)


1bardesign:
Looks like a straightforward enough benefit. 👍 

Bit of a mouthful of a property name, but doesn't matter I suppose.

1bardesign:
Only request next time would be please use a meaningful name for the branch - `patch-3` is a bit useless in the history.

PR #115, opened 2018-05-27
Fix archer grapple beyond the map top 

AsuMagic:
- The grapple hook is not physically affected during its movement when it's beyond the map top, but will visually look capped to the map top. It looked off when the grapple would hit a topmost block enough for me to think it's a better thing to deal with it like that.
- When the grapple hook hits a block, its position is clamped to the map top (this is the actual fix).

The other option would have been to just let the grapple go when it's over the map top.


1bardesign:
This way is less likely to lead to archer outcry I guess. Will need to test it a bit but looks good on first glance.

PR #118, opened 2018-05-31
Fixed arrow spawn location for mounted bows.

FuzzCat927:
Arrows (or other ammo, if that manages to happen) now start in front of the bow instead of behind.
Also, there are no inconsistencies regarding arrow spawn location and blob's facing direction anymore.


1bardesign:
(Actually, for the sake of expedient integration I'll just handle those changes myself - please review the state of this file after the merge though!)

PR #120, opened 2018-06-01
Adds Stalemates to CTF

Pirate-Rob:
Upon both teams capturing eachother's flags at the same time, a timer will start which will return all flags after a configurable amount of time. The timer can also be disabled in cfgs.

### DIFF
--- a/Entities/Special/CTF/FlagBase.as
+++ b/Entities/Special/CTF/FlagBase.as
@@ -80,6 +80,16 @@ void onTick(CBlob@ this)
 					this.server_AttachTo(b, "FLAG");
 					b.SetFacingLeft(this.isFacingLeft());
 				}
+				if (b.hasTag("stalemate_return"))
+				{
+					//sync tag, flag can play sounds
+					b.server_DetachAll();
+					this.SendCommand(this.getCommandID(flag_return));
+					b.Untag("stalemate_return"); //local
+
+					this.server_AttachTo(b, "FLAG");
+					b.SetFacingLeft(this.isFacingLeft());
+				}
 			}
 			else
 			{
--- a/Entities/Special/CTF/FlagBase.as
+++ b/Entities/Special/CTF/FlagBase.as
@@ -82,7 +82,6 @@ void onTick(CBlob@ this)
 				}
 				if (b.hasTag("stalemate_return"))
 				{
-					//sync tag, flag can play sounds
 					b.server_DetachAll();
 					this.SendCommand(this.getCommandID(flag_return));
 					b.Untag("stalemate_return"); //local
--- a/Rules/CTF/Scripts/CTF.as
+++ b/Rules/CTF/Scripts/CTF.as
@@ -23,8 +23,11 @@ void Config(CTFCore@ this)
 
 	//how long to wait for everyone to spawn in?
 	s32 warmUpTimeSeconds = cfg.read_s32("warmup_time", 30);
-	this.warmUpTime = (getTicksASecond() * warmUpTimeSeconds);
+	this.warmUpTime = 1;//(getTicksASecond() * warmUpTimeSeconds);
 
+	s32 stalemateTimeSeconds = cfg.read_s32("stalemate_time", 30);
+	this.stalemateTime = (getTicksASecond() * stalemateTimeSeconds);
+	
 	//how long for the game to play out?
 	s32 gameDurationMinutes = cfg.read_s32("game_time", -1);
 	if (gameDurationMinutes <= 0)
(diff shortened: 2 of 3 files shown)


Pirate-Rob:
Only took 40 minutes, 20$ in less than an hour, ez money.

Rules/CTF/Scripts/CTF.as
```diff
@@ -23,8 +23,11 @@ void Config(CTFCore@ this)
 
 	//how long to wait for everyone to spawn in?
 	s32 warmUpTimeSeconds = cfg.read_s32("warmup_time", 30);
-	this.warmUpTime = (getTicksASecond() * warmUpTimeSeconds);
+	this.warmUpTime = 1;//(getTicksASecond() * warmUpTimeSeconds);
```
1bardesign: Please don't leave debug modifications in official pull requests 😅 will test this as-is though and wait for changes.

Rules/CTF/Scripts/CTF.as
```diff
@@ -577,6 +582,65 @@ shared class CTFCore : RulesCore
 		}
 	}
 
+	void CheckStalemate()
```
1bardesign: afaict this could all be just on server + sync required stuff to players? seems to do that anyway

Rules/CTF/ctf_vars.cfg
```diff
@@ -13,5 +13,9 @@
 	#anything less than 0 means the game never ends
 	game_time = -1; #minutes
 	
+	#stalemate time in seconds. stalemate occurs on both teams having eachothr's flags with none of thier own. this is the time until flags are reset
+	#less than 0 turns this feature off
+	stalemate_time = 60
```
1bardesign: Thanks for adding this to the cfg - very useful.

Entities/Special/CTF/FlagBase.as
```diff
@@ -77,6 +77,16 @@ void onTick(CBlob@ this)
...
 					b.Untag("return"); //local
 
+					this.server_AttachTo(b, "FLAG");
+					b.SetFacingLeft(this.isFacingLeft());
+				}
+				if (b.hasTag("stalemate_return"))
+				{
+					//sync tag, flag can play sounds
+					b.server_DetachAll();
+					this.SendCommand(this.getCommandID(flag_return));
```
1bardesign: this command will be sent by all players if this code is not only run on the server, just double-check pls
- if it _is_ just for server, there's no need to sync the tag as the sound can be/(is?) done in the cmd
- if it's not then the cmd should only be sent by the server at least (and again, sound would probably be best in the cmd hook)

Rules/CTF/Scripts/CTF.as
Pirate-Rob: It would complicate it horribly to just do it server-side and send it, plus the check has minimal performance impact and saves bandwidth~

Entities/Special/CTF/FlagBase.as
Pirate-Rob: Yeah, it is server side, and no, it doesn't actually sync anything? That block is literally just a copy of the one above, I forgot to remove the comment xP

Pirate-Rob:
Yeah, I wasn't trying to rush honestly, I thought I was done, just skipped over some things ^^

Rules/CTF/Scripts/CTF.as
```diff
@@ -603,42 +603,56 @@ shared class CTFCore : RulesCore
...
-			if(stalemate_timer > -10 && stalemate_timer != stalemateTime){
+			if(stalemate_timer > -10 && stalemate_timer != stalemateTime)
+			{
 				rules.SetGlobalMessage("");
-				rules.set_s16("stalemate_breaker",stalemateTime);
-			} else
-			if(stalemate_timer <= -10){
+				rules.set_s16("stalemate_breaker", stalemateTime);
+			} else ///??? how the hell are elses stylised if not like this?
+			if(stalemate_timer <= -10)
```
1bardesign: (for the record)
```cpp
} else if (condition) {
```

Pirate-Rob: If this is a solution to it, I'm quite sad

1bardesign: It's the kind of thing [people have a lot of opinions about](https://en.wikipedia.org/wiki/Indentation_style). KAG source "canonically" (if you can really call it that) uses Allman style, likely inherited from MM's pascal roots.

1bardesign:
Will be integrating this today.

1bardesign:
@Pirate-Rob thank you for your work - this PR fixes an issue with a bounty, please email support with a paypal-capable email to claim the payment, citing this PR and the recent bounty announcement so we can match it all properly!

Please also make a post from your account on this issue with some kind of "proof" to mention in your email, as anyone could email claiming to be you to try to steal your bounty. We'll have to formalise this system better in future!

Pirate-Rob:
My discord id is: Pirate-Rob#5113 and I've used that account to PM furai my paypal email.

That should be proof enough :kag_yes:

PR #124, opened 2018-06-04
Translatable votes related stuff and complete Russian translation for votes.

Diprog:


terracraft321:
Seems good.

1bardesign:
@Diprog check the recent commit https://github.com/transhumandesign/kag-base/commit/81b4bef60c34e85539a3aee368025776818fae25 - it introduces some new strings that might be good to sneak into this PR, and if you need better translation features for it let me know.

For now the only concern I think is it just has the time with no unit, and relies on count-down for acknowledgement that it's a timer.

Diprog:
@1bardesign Done

1bardesign:
Thanks for the further updates @Diprog - merging it in now

PR #130, opened 2018-06-06
Update Polish_pl.json

Suprejso:
Even more actual and correct version of polish translation, thanks to BraveDragonRL for helping me with it


1bardesign:
Hey, can you close the old PR(s) in that case?

Also, you can update in-progress PRs without creating new ones - would cut down on the noise on our end if you could manage that 👍 

Glad to see the PL translation get some love!

PR #132, opened 2018-06-08
[modified] Gold material now stacks to 100 (two lumps of 50)

MrPotatobird:
It's better this way because:
- Easily grabbable chunks of 50 which is how gold is always spent, so you don't have to run out of base with your team's entire gold supply if you want to set up a tunnel
- Harder (but still possible) for one team to collect all the gold and hoard it at their base
- Easier to see when a functional amount (50) is lying on the ground instead of the current small yellow pebble
- Mining 5 blocks of ore gives you an even 100 gold anyway

### DIFF
--- a/Entities/Materials/Construction/MaterialGold.as
+++ b/Entities/Materials/Construction/MaterialGold.as
@@ -1,7 +1,7 @@
 
 void onInit(CBlob@ this)
 {
-  this.maxQuantity = 250;
+  this.maxQuantity = 50;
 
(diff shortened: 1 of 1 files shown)


Suprejso:
This would definitely help in a situation when a team has 200+ gold in one stack. I had some experiences of this kind and it's really stressful to take all the gold in your backpack if you just want to build a tunnel or whatever.

bunniewormy:
as a builder, +200

PR #133, opened 2018-06-08
Quarry updates

Mazey:
Quarry button is now translatable ("Add fuel")


Mazey:
Several changes:

- Quarry button is now translatable ("Add fuel")
- Balance pass:
-25 wood : 20 stone every 2 seconds -> 100 wood : 75 stone every 10 seconds (per minute: 750 : 600 -> 600 : 450)
-10% to drop gold instead of 20%
-Drops 20 gold from 100 wood instead of 20 gold from 25 wood
- Quarry burns the remainder of what's left in it
- Fix for quarry not taking wood when carrying wood but not having any in your inventory (thanks v6steeda)

PR #135, opened 2018-06-09
[modified] Quarry costs 50 gold and drops 50 gold on destruction

MrPotatobird:
Everything else costs 50 and drops 50 gold-wise so this should be no exception


Mazey:
No. Two blocks of gold ore give 40 gold, and two "gold drops" from a quarry is also 40 gold, the price of a new quarry. Not dropping gold on destruction is intentional, with the reason being that it occasionally generates gold.

MrPotatobird:
True, and that is something that has always annoyed me about the other gold workshops. So why don't we just #136 ?

1bardesign:
Fwiw I agree with this change, and would prefer that the quarry not create/destroy gold.

Previously the only way to destroy it was throwing (or glitching) it into the abyss; the current behaviour is a "subtle griefing" dream, as you can hamstring a team long-term without them noticing, by spending and destroying all their gold.

1bardesign:
Merging this after discussion on discord 👍

PR #136, opened 2018-06-09
[modified] Gold ore tiles yield 5 (instead of 4) gold per hit

MrPotatobird:
So each block of ore is worth 25 gold instead of 20.
- 50 gold every two blocks instead of the current 52 every two and 3/5 blocks
- Much neater stacking / spending
- This will add more gold to every map, but just one functional unit (50) per 10 ore tiles on the map so it doesn't seem that drastic to me. Especially if there's a new gold workshop being added anyway

### DIFF
--- a/Entities/Materials/MaterialCommon.as
+++ b/Entities/Materials/MaterialCommon.as
@@ -224,7 +224,7 @@ namespace Material
     }
     else if (map.isTileGold(type))
     {
-      createFor(this, 'mat_gold', 4.f * damage);
+      createFor(this, 'mat_gold', 5.f * damage);
     }
   }
 }


Diprog:
Bunnie is gay

Andresian:
Think about balance, not your needs

Diprog:
@Andresian and how is it going to disbalance game? What about some constructive criticism? Hmmm?

1bardesign:
I like the idea of more round numbers and considering the functional spending unit of gold (50 in most cases) but I don't like that it'd change the gold amount on every map.

I'd prefer to play with the gold balance after the quarry changes are in, especially since this seems comparatively controversial from the votes.

AsuMagic:
Why the downvotes? This would make it align with the other shop costs.

Vikhenzo:
This pull request was discussed a bit on the discord but we still haven't reached a decisive conclusion. The general feeling is that maps in most maps need to have their gold nerfed, this would be a good excuse/reason to start working on that. Of course as this change would add more gold maps would need to be nerfed further.

1bardesign:
This will definitely need map review alongside it.

@MrPotatobird might be best placed to do that anyway. I'm sure other community members will also have opinions on the amount of gold.

Please report back in here or discord with any results. Required map changes could be a separate PR or just appended to this one 👍

Issue #137, opened 2018-06-12
[Feature Pitch] Tap inventory key as builder to switch to the previous selected block


MrPotatobird:
When knight and archer tap F, it quickly cycles through their ammo options. It doesn't do anything for builder at the moment.

The ability to quickly cycle to the last block you were placing would make building some complex structures quite a bit more painless, as you wouldn't have to constantly keep holding f and trying to quickly seek very specific button locations with the cursor.

For experienced builders, selecting the right block is already kind of a reflex, so this might be even more welcomed by less experienced builders.

Mazey:
I feel like this would be pretty big and really handy, experienced builders can switch quickly already but this would be even more consistent and faster.

Random thought further expanding this: Add an option to link tapping F to a specific block instead of switching to the previous block.

TheOszzer:
https://github.com/transhumandesign/kag-base/issues/294

PR #138, opened 2018-06-12
[added] Tapping inventory key as builder swaps to last block

MrPotatobird:
For #137


1bardesign:
Thanks for this!

Really awesome to see people holding out and not "hoarding" for bounties, really good feeling. 👍

PR #146, opened 2018-06-23
Quarry balance/updates

Mazey:
**Work in progress!**

- rate is now 75 wood : 25 stone (was 100 wood : 75 stone)
- coverts every 8 seconds (was 10 seconds)
- increase gold cost to 100 (was 50)
- when burning wood remainder, stone rounds to the nearest multiple of 5
- max wood quarry can take is 1500 (was 1000)

24/06
- added sound to quarry


bunniewormy:
Should add a limit of 2 quarries per team.

Mazey:
Please explain why downvotes. Do you dislike the sound? Did I nerf it too much? Does it need to be nerfed even more? Come on, don't downvote if you dislike the quarry (in its current state), downvote because you disagree with the PR, you're basically saying "don't implement these changes, keep it how it is" right now because you're not giving feedback why the changes are bad.

@bunniewormy a limit of 2 quarries is the next thing to do, depending on how the changes above will play out. I didn't have time to do it and I think making them cost 100 gold will fix most of it so I didn't make it a priority (and I will go through every map making sure there isn't too much gold later).

MrPotatobird:
It's because I am against making it even more wood hungry. I also think it feels worse to have lower dispensal intervals. And honestly, I'm not even sure this is enough of a nerf. I think it comes out to a little less than half the stone output of the first? Even one quarry on live has such tremendous impact, that when I think, oh ok so now I would just have to gather 200 gold and get another two or three trees for the farm, and I'd have something comparable to my single quarry from the last build... That's worth it. I'll have to go do that first, what a pain, but now that it's harder to set up, maybe only one team will bother to or even be able to set it up.

In my suggestion #147 I said "If we're gonna just nerf the numbers, I'd say go for a hard nerf to max fuel limit + time taken, and maybe add a cap to the number of buildable quarries if possible. I wouldn't nerf conversion ratio or gold cost, it's just going to make quarries more annoying, wood hungry, and high maintenance, when the main problem is mind-boggling stone output."

I would suggest 50 gold cost, 250 wood max storage, conversion of 250 wood (or whatever's in there at the end) to 150 stone (or whatever proportionally) every 120 seconds, in intervals of either 60 or 120 seconds, but then you'd still have problem of scalability leading to OP potential so you need the quarry cap, which is what I would call "overcomplicated" because most players don't read patch notes, think about how WTF of a realization that would be in game.

I don't like the gold price hike because I feel like as long as a quarry setup is super strong and desirable, it's still gonna happen it's just also gonna leave you less gold for the other stuff. But if you think that's an  effective replacement for the hard cap on quarry limit, then I guess it's fine

1bardesign:
Gonna cherry pick the sound changes since the others dont seem popular as-is, will go for different nerf approach:

- conversion efficiency the same (to avoid making them _more_ wood hungry)
- cap at two stacks of wood (500) inserted at a time (more maintenance burden/cost than currently, less "fire and forget" bulk conversion)
- increased time taken per conversion, probably consume 100 each 30 seconds, so 150s from full to empty (also fighting wood drought and output speed)
- increased gold cost (100) for "soft cap"

**rationale on the soft cap:** if you want 2 quarries and ballista with bomb bolts you're already looking at 300 gold, and are unlikely to be able to fund tunnels... So, at least a team has to "lean in" to playing defensive, and loses more if a quarry is destroyed and the gold taken.

We can go in a different direction like #147 if it doesn't work out.

1bardesign:
Yep I definitely think coin sinks are better than sources as far as builders are concerned, haha. I don't want to swing too hard against this one though for the moment, especially with new steam players learning the ropes. It's definitely something to look into though 🙂

1bardesign:
Merged some of this manually 👍

PR #152, opened 2018-06-25
Make Boulders Fit in Inventory

Blubahub:
Make the boulders 2x2 so they can fit in a normal class's (and storage shop's) inventory. Also changed the icon's specific sprite frame so it doesn't look as bad in the hot bar when in the player's inventory.


Franrek:
I think it was like that sometime ago. It makes the slower movement when holding boulder pointless. I agree that they are little underpowered right now but i don't believe that's the way to buff them.

PR #153, opened 2018-06-25
Builder quickswap fixes/tweaks

MrPotatobird:
It was supposed to make a sound effect so it's more obvious what just happened and less confusing if you accidentally use the quickswap. It's the same sound effect that plays when you tap F as archer/knight.

Also, it was kind of weird that you could "quickswap" to the building, because that has the effect of immediately using 150 wood, so it should ignore building selection.


1bardesign:
Looks fine. I'll miss the building frame haha.

PR #155, opened 2018-06-27
[modified] Knights can no longer damage gold ore with sword

MrPotatobird:
This is one of the few ways to remove gold material from the game. Alternative: Make slashing gold ore drop some of the material


Mazey:
This would be stupid, however if you make it drop gold then a fat thumbs up.

1bardesign:
Yeah it should probably at least drop the gold on the ground.

Vikhenzo:
Yes if it drops gold ore, thumbs up

PR #160, opened 2018-06-28
[added] spanish translation almost done, only tutorial and "story' remains

Vikhenzo:
I've translated most of the text you see in normal gameplay, if everything _must_ be translated I will get it done but this should be fine for now.


1bardesign:
Thanks for this viky!

PR #166, opened 2018-07-02
Re-Adds Crates to CTF and Reworks them

Pirate-Rob:
A patch for July 2018 # 2 Bounty.

Other than allowing crates to be used as transport again, these changes also make it more useful as an inventory.

- Made crates 2 high invetory wise, to match storage.
- Players start suffocating (really slowly) if they stay in a crate far too long.
- When crates contain a player and they suffer a hard impact, they break. So you can throw teammates off of a cliff if they aren't be useful.
- Crates added to buildershop for 30 wood.
- Crates no longer get destroyed upon the last item being removed, however they still break when a player leaves them.
- Crates now automatically pick up materials they're overlapping, exactly the same as the storage building.
- Additionally, crates also now pick up any materials they collide with them.
- Crates are now immune to saws.
- Disabled emotes when inside inventories.

### DIFF
--- a/Entities/Industry/CTFShops/BuilderShop/BuilderShop.as
+++ b/Entities/Industry/CTFShops/BuilderShop/BuilderShop.as
@@ -53,13 +53,6 @@ void onInit(CBlob@ this)
 		AddRequirement(s.requirements, "blob", "mat_stone", "Stone", CTFCosts::boulder_stone);
 	}
 	{
-		ShopItem@ s = addShopItem(this, "Boulder", "$boulder$", "boulder", Descriptions::boulder, false);
-		s.customButton = true;
-		s.buttonwidth = 2;
-		s.buttonheight = 1;
-		AddRequirement(s.requirements, "blob", "mat_stone", "Stone", CTFCosts::boulder_stone);
-	}
-	{
 		ShopItem@ s = addShopItem(this, "Trampoline", "$trampoline$", "trampoline", Descriptions::trampoline, false);
 		AddRequirement(s.requirements, "blob", "mat_wood", "Wood", CTFCosts::trampoline_wood);
 	}
--- a/Entities/Industry/CTFShops/BuilderShop/BuilderShop.as
+++ b/Entities/Industry/CTFShops/BuilderShop/BuilderShop.as
@@ -53,6 +53,13 @@ void onInit(CBlob@ this)
 		AddRequirement(s.requirements, "blob", "mat_stone", "Stone", CTFCosts::boulder_stone);
 	}
 	{
+		ShopItem@ s = addShopItem(this, "Boulder", "$boulder$", "boulder", Descriptions::boulder, false);
+		s.customButton = true;
+		s.buttonwidth = 2;
+		s.buttonheight = 1;
+		AddRequirement(s.requirements, "blob", "mat_stone", "Stone", CTFCosts::boulder_stone);
+	}
+	{
 		ShopItem@ s = addShopItem(this, "Trampoline", "$trampoline$", "trampoline", Descriptions::trampoline, false);
 		AddRequirement(s.requirements, "blob", "mat_wood", "Wood", CTFCosts::trampoline_wood);
 	}
@@ -64,10 +71,17 @@ void onInit(CBlob@ this)
 	{
 		ShopItem@ s = addShopItem(this, "Saw", "$saw$", "saw", Descriptions::saw, false);
 		s.customButton = true;
-		s.buttonwidth = 2;
+		s.buttonwidth = 1;
 		s.buttonheight = 1;
 		AddRequirement(s.requirements, "blob", "mat_wood", "Wood", CTFCosts::saw_wood);
 		AddRequirement(s.requirements, "blob", "mat_stone", "Stone", CTFCosts::saw_stone);
+	}
+	{
+		ShopItem@ s = addShopItem(this, "Crate", "$crate$", "crate", Descriptions::crate, false);
+		s.customButton = true;
+		s.buttonwidth = 1;
+		s.buttonheight = 1;
+		AddRequirement(s.requirements, "blob", "mat_wood", "Wood", CTFCosts::crate_wood);
 	}
 }
 
--- a/Entities/Industry/CTFShops/CTFCosts.cfg
+++ b/Entities/Industry/CTFShops/CTFCosts.cfg
@@ -14,13 +14,15 @@
 	cost_tunnel_stone               = 100
 	cost_tunnel_wood                = 50
 	cost_tunnel_gold                = 50
+	cost_quarry_stone               = 150
+	cost_quarry_gold                = 100
 
 	#knight shop
 	cost_bomb                       = 25
 	cost_waterbomb                  = 30
 	cost_mine                       = 60
 	cost_keg                        = 120
-	
+
 	#archer shop
 	cost_arrows                     = 15
 	cost_waterarrows                = 20
@@ -38,6 +40,7 @@
 	cost_saw_stone                  = 100
 	cost_drill_stone                = 100
 	cost_drill                      = 25
+	cost_crate_wood                 = 30
 
 	#boat shop
 	cost_dinghy                     = 25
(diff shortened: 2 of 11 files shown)


Entities/Industry/CTFShops/BuilderShop/BuilderShop.as
```diff
@@ -52,6 +52,13 @@ void onInit(CBlob@ this)
 		s.buttonheight = 1;
 		AddRequirement(s.requirements, "blob", "mat_stone", "Stone", CTFCosts::boulder_stone);
 	}
+	{
+		ShopItem@ s = addShopItem(this, "Boulder", "$boulder$", "boulder", Descriptions::boulder, false);
+		s.customButton = true;
+		s.buttonwidth = 2;
+		s.buttonheight = 1;
+		AddRequirement(s.requirements, "blob", "mat_stone", "Stone", CTFCosts::boulder_stone);
+	}
```
AsuMagic: There shouldn't be this duplicate boulder entry.

Entities/Industry/CTFShops/CTFCosts.cfg
```diff
@@ -14,13 +14,15 @@
 	cost_tunnel_stone               = 100
 	cost_tunnel_wood                = 50
 	cost_tunnel_gold                = 50
+	cost_quarry_stone               = 150
+	cost_quarry_gold                = 100
```
AsuMagic: This commit shouldn't affect quarries, why is it there?

Entities/Items/Crate/Crate.as
```diff
@@ -88,6 +89,8 @@ void onInit(CBlob@ this)
 	}
 
 	this.getSprite().SetZ(-10.0f);
+	
+	this.Tag("sawed"); //HACK: prevents saws from sawing crates :(
```
AsuMagic: This kind of logic is/should be handled in Saw.as:156

Entities/Items/Crate/Crate.as
```diff
@@ -115,7 +118,8 @@ void onTick(CBlob@ this)
 			this.getCurrentScript().tickFrequency = 15;
 		else
 		{
-			this.getCurrentScript().tickFrequency = 0;
+			this.getCurrentScript().tickFrequency = 60;
```
AsuMagic: This should be 30, not 60 (not that the behavior will be different).

AsuMagic: Besides it should be lowered to a more sane amount (i.e. every few ticks) as to avoid (likely) expensive `getBlobsInBox` logic for each crate each tick.

Entities/Items/Crate/Crate.as
```diff
@@ -139,6 +143,25 @@ void onTick(CBlob@ this)
...
+		for (uint i = 0; i < blobs.length; i++)
+		{
+			CBlob@ blob = blobs[i];
+			if (!blob.isAttached() && blob.isOnGround() && blob.hasTag("material") && blob.getName() != "mat_arrows")
+			{
+				this.server_PutInInventory(blob);
+			}
+		}
+	}
+}
```
AsuMagic: Duplicate from `Storage.as`, this function should be merged into a common script for `Crate.as` and `Storage.as`.

Entities/Items/Crate/Crate.as
```diff
@@ -162,6 +185,36 @@ bool doesCollideWithBlob(CBlob@ this, CBlob@ blob)
 	return !blob.hasTag("parachute");
 }
 
+void onCollision( CBlob@ this, CBlob@ blob, bool solid ){
+	
+	if(!isServer())return;
+	
+	if(blob !is null){
+		if (!blob.isAttached() && blob.isOnGround() && blob.hasTag("material") && blob.getName() != "mat_arrows")
```
AsuMagic: Short-circuiting makes it possible to merge the two conditions: `blob !is null && !blob.isAttached() && ...`

Entities/Items/Crate/Crate.as
```diff
@@ -162,6 +185,36 @@ bool doesCollideWithBlob(CBlob@ this, CBlob@ blob)
 	return !blob.hasTag("parachute");
 }
 
+void onCollision( CBlob@ this, CBlob@ blob, bool solid ){
+	
+	if(!isServer())return;
```
AsuMagic: Add spaces between `)` and other tokens, also avoid the `f( a, b, c );` "convention" for Crimson, rather do `f(a, b, c);`

Entities/Items/Crate/Crate.as
```diff
@@ -162,6 +185,36 @@ bool doesCollideWithBlob(CBlob@ this, CBlob@ blob)
...
+	
+	CInventory@ inv = this.getInventory();
+	
+	if(inv !is null){
+		for(int i = 0;i < inv.getItemsCount();i++){
+			CBlob @item = inv.getItem(i);
+			if(item !is null){
+				if(item.hasTag("player")){
+					//Enable this line if you want crates to damage players on impact
+					//this.server_Hit(item,item.getPosition(),Vec2f(0,0),1.0f,Hitters::fall,true);
```
AsuMagic: Sounds like you could pass the velocity in there, so the fall damage applies accordingly.

Entities/Items/Crate/Crate.as
```diff
@@ -162,6 +185,36 @@ bool doesCollideWithBlob(CBlob@ this, CBlob@ blob)
...
+	
+	if(!solid || this.getOldVelocity().Length() < 6.0f)return; 
+	
+	CInventory@ inv = this.getInventory();
+	
+	if(inv !is null){
+		for(int i = 0;i < inv.getItemsCount();i++){
+			CBlob @item = inv.getItem(i);
+			if(item !is null){
+				if(item.hasTag("player")){
```
AsuMagic: Same as above, combine with `&&` short-circuiting

Entities/Industry/CTFShops/BuilderShop/BuilderShop.as
Pirate-Rob: I... don't even know how that got in, it doesn't even appear in the shop xD

Entities/Industry/CTFShops/CTFCosts.cfg
Pirate-Rob: Not intentional, my file must've been out of date or newer?

Entities/Items/Crate/Crate.as
Pirate-Rob: Should I make an exception for crate, or add a "do not saw" tag?

AsuMagic: A tag like that sounds more sane indeed, maybe for another PR though.

Entities/Items/Crate/Crate.as
Pirate-Rob: Tick frequency being a high number lowers the amount of ticking (so 60 makes it tick only every 2 seconds)

AsuMagic: Yep, my bad :man_facepalming:

Entities/Items/Crate/Crate.as
Pirate-Rob: Alrighty, wasn't sure if I should do that or not. Maybe I should make it a separate script instead? (not a common, something like AutoPickupMats.as)

Entities/Items/Crate/Crate.as
Pirate-Rob: Got it!

Entities/Items/Crate/Crate.as
Pirate-Rob: Sorry, came from the manual, or somewhere

Entities/Items/Crate/Crate.as
Pirate-Rob: Ah yeah, I was thinking about it at first, but since knight can literally just fly with crates, applying damage/forcing a fall just makes a solo crates from catapults much worse/useless. (and the idea behind only dealing 1 damage is so that any class can survive the fall)

eps0003:
30 wood could lead to a lot of spam. The size of each crate would make it horrible.

FuzzCat927:
Unless the DecayIfSpammed.as script was removed along with these changes, spamming them shouldn't be an issue. For example, lanterns cost 10 wood, but you can't cause lag with lantern spam, as they decay if there are too many of them.

1bardesign:
Closing this; would have loved to see the suffocating merged in with potatobird's crate stuff but the vote was against.

PR #179, opened 2018-07-09
Crate Hiding + Fulll Crate Rework

MrPotatobird:
If a feature that was deliberately removed is gonna come back to the game I figure we should make sure that it's gonna be different this time. But also, I don't want to kill the fun and ruin the parts about crates that people loved and wanted back in the first place.

So here it is, the full-scale crate makeover that absolutely nobody asked for, but I'm sure we were all secretly hoping for whether we knew it or not!

crates1

Featuring:
- Hide in a crate by holding it and pressing E!
    - Press E again to break the crate! Surprise!
    - Some of the items inside will pop out and make room for you, and you can't get out without breaking the crate, so make sure hiding in it is a good idea first!
    - If your teammate is sitting around in a crate and being useless, deliver them to the enemy as a present or walk up to their crate and press E to bust them out!
    - Careful of suspicious crates - if you try to open one with an enemy, you'll be stunned! Break it or find some other way to make sure it's safe!
- Carry crates around to scoop up items from the battlefield with automatic pickup!
- Booby-trap crates by putting mines into them!
- Teammates can access crates more easily than enemies. To grab + take an enemy crate, you need to be touching it!
- Throw crates into saws to process them and the logs (and/or enemies) inside them!
- Buy them at the builder shop for either 150 wood or 20 coins

Full changelog, somewhat organized, for those interested in nitty-gritty: https://pastebin.com/RG1PHqKY


Franrek:
I like all the changes except for the stun that seems unintuitive and annoying and the price which in my opinion should be 150 wood AND not OR 20 coins.

MrPotatobird:
Yeah, the stun (I think it's 1 second) was one of the things I was less sure about. But I feel like if you're suspicious of a crate you shouldn't be able to just walk up to it, charge a slash, and press E to confirm your suspicions with no risk or loss, you should either attack and break the crate, put a lit bomb in it, grab the crate and throw it in a saw / shark pond etc. etc.

As for the cost, maybe it should be joined, but this is an item that knights and archers are able to use too, and making them rely on builders or switch to builder themselves and gather 50 more wood just so they can buy a crate seems weird to me. I guess they could also just grab a lump of wood lying on the ground though.

Inferdy:
Why crates won't be heavier, if there's a player inside?

ParaLogia:
There's a currently an issue with how crates check whether you're touching them. Your "touching" status is lost after you pick up the crate, so if you pick up a crate and drop it right at your feet, you can't pick it up again or re-open it until you move away from it and come back. Example video: https://youtu.be/sKpk88-tDcM

Issue #180, opened 2018-07-10
Votekicks display wrong username info in chat


FuzzCat927:
Votekicks display the username of the person who started the kick, rather than the one who the vote is on.
The information about who started the vote is correct.

![Image 1](https://steamuserimages-a.akamaihd.net/ugc/946203282875579581/764E332973EAB0519B9886E3D5BBBDDB5DC8AFC8/)
![Image 2](https://steamuserimages-a.akamaihd.net/ugc/946203282875579433/9248073C10D49F43B124C350185635EFA3CE132C/)

ParaLogia:
Does it show the wrong name in the voting box too, or is just in chat?

1bardesign:
This should be fixed by #181 and the resulting changes afterwards; will push them to github soon but they are merged with a few more sensitive changes currently.

PR #181, opened 2018-07-11
[fixed] the message for vote kicking

Vikhenzo:
As shown in #180. Very simple fix. I'm fairly certain this was the only vote message that had this kind of error. The vote object creates a title "Kick {USER}" but that user is the one that created it and then that title was passed to the VotesCommon script. I simply circumvented it by manually getting the user_to_kick variable.


AsuMagic:
Isn't this also used for other votes, i.e. next map votes? In that case it would fail.

Vikhenzo:
No, because the other vote's titles are correct, this was the only one that even has a "custom" title, the other ones are constant and just strings.

1bardesign:
this wasn't a complete fix but i've made some changes after merging it, effectively amounting to this: 

screenshot of code fix

in cases like the nextmap vote or surrender vote we need to use the correct title. in the votekick case we can use this new string.

PR #196, opened 2018-07-29
New hover messages

AsuMagic:
```
[modified] New hover messages: Colors, better killfeed, subtle anims, new code
[modified] Material hover messages now display non-material entities entering/leaving inventory
[modified] Material hover messages takes into account materials lost when swapping classes and when dying
```

sample pic 1

sample pic 2

The whole hovermessage logic has been overhauled, now message specific data is handled in inherited classes and those inherited classes generate the tokens (string + color + offset combo) used for rendering.

Movement is subtly smoothed out, so when new messages or when message are removed it moves back smoothly, which makes it less distracting (at least for me).

Messages fade out and have a slight movement to the right effect when disappearing.

Requires more extensive testing on net although everything there is client only. Will require some feedback in testing as whether it's too intruisive (because more noticeable colors and different movement).


Vikhenzo:
Why hasn't this gained more traction? I think it's really good. Hasn't even hit testing yet.

PR #197, opened 2018-07-29
[added] prototype admin commands

Vikhenzo:
Can report people with !r Username and server admins can !m to enter moderation mode. Still need to make better username matching.


Vikhenzo:
ok i think it's fine for now, we should add it to the next test build

Issue #200, opened 2018-07-30
[Feature Pitch] add Taunts, like in Classic


bunniewormy:
In Classic, you could click V and choose a Taunt. It was a pre-made message, which could either be normal chat or team chat, that would be sent to the chat. 

There were a variety of taunts, such as:

"For the King!"
"Run for your life!"
"Chicken!"
"Your mother was a hamster!"
"Your father smelt of elderberries!"
"I fart in your general direction!"

Some class/team ones:
"The red builder always triumphs!"
"The blue archer always triumphs!"
"My sword is bigger than your sword!"
 
Some useful team chat orders:

"Cover me!"
"Hold position!"
"Fall back!"
"MOVE!"
"Follow me"
"Rush!!!"
"Repair the castle!"
"Load the catapult!"

Some team chat alarm ones:

"Incoming!"
"Under catapult fire!"
"Trap!"
"Rush in tunnels!"

Some ordering to build stuff:

"Build a wall here!"
"Build a ladder here!"
"Build a catapult!"

And some asking for stuff:
"Need builders!"
"Need bombs!"
"Need help!

My suggestion is to bring this system back. The taunting taunts could be fun, and the team chat ones were pretty nice for teamwork. Perhaps this would get more people to teamwork in pubs. You could also add custom taunts, so people can change it themselves. It would also help with teamwork in competitive, especially if you don't use voice chat.

Of course, one problem could arise because of that; spam. But I believe it would be very easy to counter that; you could put a 10-15 second cooldown on public chat taunts, and like 2-3 seconds cooldown on team chat ones. BTW, the ignore function exists for a reason ;)

I also suggest that Taunts also show chat bubbles, and have a different option for enabling/disabling them:
![image](https://user-images.githubusercontent.com/39147836/43410629-d681161e-9427-11e8-9874-02116c3ab77b.png)
This way you could disable normal chat bubbles [to prevent spam cluttering your screen], but you could enable taunt bubbles [so you can see your teammates asking for help/giving information]

Radinyn:
Bloody peasants!

AsuMagic:
This could be implemented fairly easily when my ongoing work on #154 is merged.

The settings option will require an engine side change though.

MrPotatobird:
I think it should just be sent as a chat bubble, rather than go into the actual chat box. Then you could "spam" it as much as you want! They would just have to be kept rather short to prevent massive text bubbles covering big parts of the screen, like when you write an entire sentence.

[...]

Issue #154, opened 2018-06-27
[Feature Pitch] In-game Emote Binding Editor


koihub:
A GUI for easily changing the emote bindings (1-9 keys) in-game, instead of manually finding emote IDs, editing EmoteBindings.cfg, etc. Maybe a dialog in settings, or selecting them in the emote wheel or something.

Also, persist emote bindings after updating (e.g. Move EmoteBindings.cfg out of `/Base` into install folder like mods.cfg and autoconfig.cfg).

AsuMagic:
Pretty sure rewriting emotes to be handled in-scripts could help with changing this kind of stuff.  
It could make emote wheel modification possible and blend nicely with this idea.

AsuMagic:
Currently implementing a new scripted emote menu. I made it generic enough to be used by other mods and to be reused in the future.

IDK if I will implement the emote bindings and menu editor that soon though.

Blubahub:
To note you'd probably want this to be a child script of autoconfig.cfg, so you can easily find, edit, and store it.

1bardesign:
Sorry for the late note: This pitch has been selected for a bounty.

Definitely talk with @AsuMagic if you are interested in participating to avoid conflicts.

PR #201, opened 2018-07-30
 Wheel menu and new emote code

AsuMagic:
```
[added] new wheel menu code, allowing modders to add new menus based on a wheel with customizable items
[added] new emote code based on the new wheel menu
[added] emote menu can be manually customized in EmoteEntries.cfg
```

This can be used as a basis for #154 and #200.

Customized emote list from `EmoteEntries.cfg`:
Emote pic 1

Regular emote list:
Emote pic 2

Custom wheel menu entries can be implemented through inheritance and by overriding `render`.  
Two entry types are shipped in, regular text and icon (used for emotes) entries.

This should support translations entirely.

This requires testing on different screen resolutions, on net, and will require an engine change to disable hardcoded emote code. As a retrocompat measure that could be done only when the emote texture is not set in engine.


AsuMagic:
(had a conflict because I accidentally based the code on `origin/master` rather than `upstream/master`)

terracraft321:
I find this useful as some players don't use some emotes at all, thus being able to change them for yourself makes it faster/easier to find the emote you're looking for. Alternatively emote bindings but that still requires fiddling in the game files anyways 🤔

PR #206, opened 2018-08-07
[modified] Autopickup to player before crate if possible

MrPotatobird:
So that builders get the mats they walk over and they don't all get stolen by the crate etc. etc.


1bardesign:
Thanks PotatoBird 🎉

PR #208, opened 2018-08-09
[added] list of spectators now shows usernames

Vikhenzo:
Shows username if it's different to charactername, otherwise only shows charactername. Will be useful in the future for the reporting system.
image


1bardesign:
Thanks Viky 👍

PR #211, opened 2018-08-15
change dinghy cost

FuzzCat927:
Dinghies now cost either 100 wood or 25 coins, but not both.


Blubahub:
25 coins is ... way to cheap. Talk about dinghy spam. Maybe raise the price to 80 - the same price as kegs?

terracraft321:
50/75

FuzzCat927:
25 coins is reasonable for dinghies since they are hardly used, if at all, and for good reason-- they suck.

Issue #159, opened 2018-06-28
Held items should have momentum when dropped as you die


AsuMagic:
That is, for example, when getting killed by an archer while you're thrown from a catapult and holding a keg.

When you die, the keg should keep going forward, rather than falling straight vertically.

1bardesign:
@AsuMagic is this maybe related to the weird forces on bounced items after being grabbed that @MrPotatobird found while working on the crate stuff?..

1bardesign:
This pitch has a bounty posted for $20! Please signal here if you are interested in contributing.

makmoud98:
working on it now

1bardesign:
Fixed by #218 - @makmoud98 please get in touch with us securely with your paypal email for bounty payout.

PR #218, opened 2018-09-07
held items now have momentum when dropped as you die

makmoud98:
fixes issue #159 
tested with current public build and with current test build

### DIFF
--- a/Entities/Characters/Scripts/RunnerDeath.as
+++ b/Entities/Characters/Scripts/RunnerDeath.as
@@ -1,6 +1,10 @@
 #include "Hitters.as"
 
 const u32 VANISH_BODY_SECS = 45;
+const f32 CARRIED_BLOB_VEL_SCALE = 1.0;
+const f32 MEDIUM_CARRIED_BLOB_VEL_SCALE = 0.8;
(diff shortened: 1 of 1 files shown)


Entities/Characters/Scripts/RunnerDeath.as
```diff
@@ -90,7 +94,20 @@ f32 onHit(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@ hitt
 		this.getShape().getVars().onladder = false;
 		this.getShape().checkCollisionsAgain = true;
 		this.getShape().SetGravityScale(1.0f);
-		// fall out of attachments/seats // drop all held things
+		//set velocity to blob in hand
+        CBlob@ carried = this.getCarriedBlob();
+        if(carried !is null){
+            Vec2f current_vel = this.getVelocity() * CARRIED_BLOB_VEL_SCALE;
+            if(carried.hasTag("medium weight"))
+                current_vel = current_vel * MEDIUM_CARRIED_BLOB_VEL_SCALE;
```
1bardesign: (afaik this should work with `*=` - not going to hold the PR just for that, but noting it for your benefit)

1bardesign:
Thanks @makmoud98, looks like a nice straightforward implementation. ❤️

Issue #191, opened 2018-07-27
[Feature Pitch] Fix holiday scripts


Mazey:
Holidays were a very fun addition to KAG, fixing the currently disabled holiday script would allow for the old holidays (like halloween with the gregs) to be revived and new holidays to be added, which would add some diversity to the game and allow for fun stuff that normally wouldn’t fit KAG.

This feature pitch requires extensive testing of Holiday.as (mainly when a holiday is over), I believe more information is in that scripts.

bunniewormy:
Another fun thing [kinda related] would be having random events on servers, that could occur for, lets say, 1 in 30-40 games. It could sometimes spice the game up; for example, you could make kegs 2x cheaper for one game, or have double coin gain, and so on. Of course, you'd need to also add a notification that an event is happening somewhere, so everyone knows about it.

It could of course be a server-side variable [so you can disable it for your own server]

1bardesign:
This pitch has a bounty posted for $30! Please signal your interest here if you are interested in contributing.

makmoud98:
working on it

Blubahub:
I'd think if you are gonna do what bunnie says, @makmoud98 , then you would probably want to both print it to the chat, and have yellow text pop up on screen (same as the "Spawning in ..." text) - well, that's my opinion..!

Mazey:
Bunnie's idea is not possible with holidays (and would be best done by making a separate system). The pitch is just for fixing the existing holidays system.

1bardesign:
Fixed by #219 - we'll pay out once it's passed testing but I'll close this to avoid anyone else starting on it.

Further "event" system implementation should indeed be done separately, but does sound fun 👍

PR #219, opened 2018-09-08
fixed holiday.as not removing the scripts properly

makmoud98:
seems to be working, but needs more testing
fixes issue #191


1bardesign:
Thanks - will merge this through and we'll test a little more; I'll close the bounty for now so that we don't have anyone else start on it.

PR #230, opened 2018-09-28
accolades

rayneomancer:
temporary bad code for tournament heads + shyr's scoreboard badges


1bardesign:
@rayneomancer @bunniewormy @FuzzCat927 i modified this system to kind of "unify" the accolades and tourney heads system

see `Rules/CommonScripts/accolades_data.cfg` for details.

heads will need to be integrated into the TournamentHeads.png sheet; we can do another separate one for patreon users and easily plug it into the script-side system when that code/platform exists.

PR #232, opened 2018-10-04
[modified] update accolades config

Mazey:


1bardesign:
Looks fine, I've got preliminary changes from bunnie as well but I'll merge in these corrections now

PR #237, opened 2018-10-11
Chickens In Inventory

Blubahub:
Allows chickens to be stored in a player's inventory. They are non-stackable, and each chicken takes up a single slot (1x1). The main benefit I can see if being able to safely store and/or run away with your chickens without them getting killed (i.e. if your chicken farm gets attacked).

gthwmro

Thus, (as a vanilla class) you can hold 4 chickens in your inventory.


1bardesign:
We'll see how it goes in testing. Seems like a straightforward change, I wonder if they'll misbehave on net somehow.

PR #246, opened 2018-10-18
[added] code of conduct

Vikhenzo:
added boilerplate code of conduct


1bardesign:
Same goes for this, could use some more kag-appropriate information.

A specific feature pitch template would also be great.

PR #247, opened 2018-10-18
[added] issues templates

Vikhenzo:
added templates for bug reports and feature pitches, hopefully people will use them


1bardesign:
Same goes for this as for #248 , could use more kag-appropriate info - also a specific feature pitch template for people to use.

Could maybe use one for maps as well, if the map mods are game to move map submission/review to github - but it seems that'll mostly stay through forums/discord for now.

(apologies for doubling up - had too many tabs open 😅)

PR #248, opened 2018-10-18
[added] Pull request template

Vikhenzo:
A simple pull request template hopefully people will use it


1bardesign:
This needs some modification to be kag-appropriate 🙂

We're not likely to see tests _or_ documentation, and very few contributors are using linux or even a git client 😅 Testing instructions are probably good. Overall it could be a fair bit shorter so as to be less burden on a new contributor.

PR #255, opened 2018-11-03
Removal of trampoline bounce cooldown

Seanharrs:
Status: IN DEVELOPMENT

The changes made in this version simply comment out the code previously used to set a cooldown on a trampoline after something bounces on it. This allows for multiple objects to land on a trampoline at the same time and all be bounced correctly. As far as I could determine, trampoline-cannons are still not possible in this version, nor could I create infinitely looping spikes.

To test the changes, simply place a trampoline in sandbox and drop two or more items onto it so they hit at the same time, all items should bounce.

Two alternate solutions exist in the per-obj-bounce-cooldown and unique-cooldown branches, but the latter appears to have an identical result to this and the former can lead to unexpected situations where a player/object moves through a trampoline since its cooldown is active.


1bardesign:
Thanks for your take on this, @Seanharrs!

As discussed on discord, we'll go with @makmoud98 's implementation this time around, but your interest and effort are appreciated! Hope to see other contributions from you in the future, if you have time 👍

PR #266, opened 2018-11-16
Fix to ensure emote menu disappears when you die

eps0003:
## Status

**READY**

## Description

This fix ensures the emote wheel disappears when you die

## Steps to Test or Reproduce

Die while holding Q


Vikhenzo:
It looks alright, should be merged soon

PR #270, opened 2018-11-17
Certain items no longer sit on top of saws

eps0003:
## Status

**READY**

## Description

This fix makes certain items pass through saws rather than just sitting on top
I'm pretty sure I haven't missed any items

It's a fix for my [bug report](https://forum.thd.vg/threads/26355/)

## Steps to Test or Reproduce

Throw a burger, grain, heart, scroll or steak on top of a saw


1bardesign:
This causes scroll that has fallen through platforms the affected items to ignore platforms completely. Might be better to put this into the saw collision filter directly.

eps0003:
I could tell there would be unintended side effects. It seemed too easy to be the right solution.

1bardesign:
Often the way with KAG fixes I'm afraid. The saw collision filter should be fairly straightforward to add to though! 👍

PR #273, opened 2018-11-18
More consistent crouch dodge

### LINKED ISSUE #249: Make crouch work only when the crouch sprite is visible.
Right now crouch doesn't work whenever player is moving in any way and at any speed. I would like to change that and disable crouch when it's the player who is inputing movement on his character. That would mean you can crouch while you're slowly sliding down of an edge, which right now is really annoying and seems random since movement like that can be so slight its pretty much invisible. 
ezgif com-crop 1 
ezgif com-resize 9

Change like that would also mean you could do some cool stuff like gaining a lot of speed as archer and sliding through other players. However I would count grapple as a movement action that stops crouching (not sure bout that yet) to make it require more skill.
ezgif com-video-to-gif 2

Obviously I wouldn't want to allow crouching in midair. So there should be a requirement that character is in contact with the ground below him to crouch.

eps0003:
## Status

**READY**

## Description

This changes the condition of being 'still' from having a low velocity to not pressing left and right. It makes it much more consistent to crouch dodge.

This resolves #249 

## Steps to Test or Reproduce

Test 1: Throw a boulder in the air and crouch to dodge it

### DIFF
--- a/Entities/Characters/Scripts/RunnerCollision.as
+++ b/Entities/Characters/Scripts/RunnerCollision.as
@@ -56,10 +56,8 @@ bool doesCollideWithBlob(CBlob@ this, CBlob@ blob)
 			return false;
 		}
 
-		const bool still = (this.getShape().vellen < 0.01f);
-
-		if (this.isKeyPressed(key_down) &&
-		        this.isOnGround() && still)
+		CSprite@ sprite = this.getSprite();
+		if (sprite !is null && sprite.isAnimation("crouch"))
 		{
 			CShape@ s = blob.getShape();
 			if (s !is null && !s.isStatic() &&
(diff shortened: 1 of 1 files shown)


1bardesign:
Will need to give this more of a close look, but thanks for tackling it before the bounty was even posted! ✨

benjamingorman:
Sorry for not commenting earlier.

Haven't tested this but couldn't an archer move through doors/platforms with this change just by grappling and sliding along the ground whilst holding S?

Also I think crouch sometimes fails because the character is very very slightly in air for some reason. It can also happen when standing on ground that looks like this:
```
 x
□ □
□□□
```
Perhaps a small tolerance should be added to `isOnGround` to account for the player being 0.01 pixels off ground for whatever reason.

1bardesign:
It will not allow passing through doors etc as crouch does not affect collisions with static blobs (see [here](https://github.com/eps0003/kag-base/blob/8deec8aa330ec20d7a4e34f0858e7375da9db6ae/Entities/Characters/Scripts/RunnerCollision.as#L65))

Sidenote: The ground like that is because it's the bottom quarter of the collision shape that "counts" towards being on ground. Adding extra sensors to detect "kind of" on ground is not likely to happen I'm afraid.

PR #278, opened 2018-11-21
Use highest absorbed sponge first when stunned by water

eps0003:
## Status

**READY**

## Description

When you have multiple sponges in your inventory, the one which has absorbed the highest amount of water is used first. This reduces inventory clutter by using each sponge one at a time rather than spreading uses over all of them.

## Steps to Test or Reproduce

Put sponges in your inventory and stun yourself with water

### DIFF
--- a/Entities/Characters/Scripts/RunnerKnock.as
+++ b/Entities/Characters/Scripts/RunnerKnock.as
@@ -67,19 +67,19 @@ f32 onHit(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@ hitt
 	{
 		//get sponge
 		CBlob@ sponge = null;
-		//find the sponge with lowest absorbed
+		//find the sponge with highest absorbed
 		CInventory@ inv = this.getInventory();
 		if (inv !is null)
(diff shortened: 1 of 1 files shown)


Entities/Characters/Scripts/RunnerKnock.as
```diff
@@ -67,19 +67,19 @@ f32 onHit(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@ hitt
 	{
 		//get sponge
 		CBlob@ sponge = null;
-		//find the sponge with lowest absorbed
+		//find the sponge with highest absorbed
 		CInventory@ inv = this.getInventory();
 		if (inv !is null)
 		{
-			u8 lowest_absorbed = 100;
+			int8 highest_absorbed = -1;
```
1bardesign: (this code is vulnerable to overflows, though they shouldn't happen; will be fixing it in post - don't bother using int8 in most circumstances, just use int. u8 is just used to optimise the property size on net, so using a matching type for the search made sense; using a matching bit width but signed type does not)

PR #280, opened 2018-11-27
Better crouch dodge

eps0003:
## Status

**IN DEVELOPMENT**

## Description

As requested, I've reimplemented crouch dodging by using tags to make the code more future-proof. Currently, crouch dodging can be done by pressing down while visibly crouching or while shielding as knight. This can easily be changed by adding the "crouch dodge" tag in the animation states you want.

## Steps to Test or Reproduce

Test 1: Throw a boulder in the air and crouch to dodge it while in different animation states


Entities/Characters/Scripts/RunnerDefault.as
```diff
@@ -26,6 +26,7 @@ void onInit(CBlob@ this)
 
 void onTick(CBlob@ this)
 {
+	this.Untag("crouch dodge");
```
Vikhenzo: Are you untagging and retagging every frame? That doesn't seem great.

eps0003: What should I be doing instead? Should I untag in every animation state that doesn’t allow crouch dodging?

FuzzCat927: probably need to tag / untag only when an animation state that doesn't allow starts / ends

eps0003: That’ll require lots of repetitive lines of code...

PR #285, opened 2018-12-06
updated accolades

bunniewormy:
## Status

(pick one)

## Description

updated accolades... hopefully gud this time


bunniewormy:
fixed lava and ricecooker. custom heads was originally going with order of when someone got their head, but i changed it. different sections was originally geti's idea, the sections have a hierarchy so youre always in the one that is higher than the others you have

1bardesign:
Agree with the idea of merging all the sections; I kept them separate initially as it _seemed_ more organised, but as people seem to move between categories quite often, probably makes more sense to merge them all together in a big alphabetically sorted collection as said above.

PR #286, opened 2018-12-08
Implemented better crouch dodge code

eps0003:
## Status

**READY**

## Description

This includes the requested changes from my previous pull request. This implementation uses the `"allow crouch"` tag to specify which states allow crouch dodging and an `isCrouching()` function to check if you’re crouching.

States which enable crouch dodging:
- Visibly crouching
- Placing blocks as builder
- Shielding as knight

I went with whitelisting instead of blacklisting because only a few states allow crouch dodging.
I included the builder's block place state because it isn't an attacking frame which doesn't allow people to attack while crouch dodging.

## Steps to Test or Reproduce

Test 1: Throw a boulder in the air and crouch to dodge it while in different animation states


1bardesign:
👍 thanks folks! merging this through to test now 💖

Issue #241, opened 2018-10-15
[Feature pitch] Water rework


benjamingorman:
Water arrows and bombs are overpowered and should be reworked.

They can be frustrating to play against because the only counterplay is to buy a sponge. Buying a sponge kind of sucks though because it uses an inventory slot and most of the time if you buy one at base you won't end up using it.

However I still feel obligated to buy sponges all the time because it's the only way to have a chance against someone hitting you with water. If you don't have a sponge and an archer pulls out a water arrow on you it's just game over.

Some suggestions which might help the problem:

- Water could not stun but instead just provide a large knockback. This could still be used to disrupt the enemy team; knocking people off towers etc.
- Knights could have some way to block water arrows. Maybe shield could block them like all other arrows? Maybe they could be slashed out of the air in the same way that ballista bolts can?
- The stun effect could have diminishing returns so at least you can't be chain stunned for 10+ seconds.
- The water arrow could just be replaced with an entirely different kind of arrow.

I think that if there is a change here it's important that:
- It isn't too much of a nerf to archers.
- There's still a way to put out fires.

Simply increasing the cost of water arrows, or decreasing the amount that archers can hold would not really help the problem.

ferrezinhre:
If I may propose a few solutions:

First: Simple one, decrease stun time from 45 to 30 or 25 (would say it will be playerbase choice), decrease area of "splash"(require "direct hit") and removing "stun trought walls" (personally  I would add small damage, just for fun)

Second: We just don't like stun, it's bad evil, and devilish invention (right, like you can't just evade water eh?). Let's then water arrows and bombs slow down players instead stunning them, possible but require new .as just for water slowdown (I think?)

Third: Just try evade water arrows and water bombs, it's possible (btw: stating that "they're not fun for anyone and can be very frustrating" is irrational not everyone have same view as you dear. ) anyway there are my solutions. ps: (what about endless shield stun loop? )

benjamingorman:
Thanks for supporting the issue Ferrez.

First: All these things would definitely help. Although I think that just 'tweaking the numbers' masks the problem rather than fixes it, because the water stun mechanic itself is the underlying issue.

Second: Slowdown is a good idea yeah, this could definitely work.

[...]

PR #291, opened 2018-12-09
Halve the stun from water if it was shielded

benjamingorman:
## Status

READY

## Description

This PR adds some counterplay to water stuns by allowing them to be blocked. If successfully blocked then the stun is halved (22 rather than 45 ticks).

22 ticks is still a long time - enough for water stuns to be powerful but not oppressive.

Sponges function as normal.
This affects water bombs and water arrows. I've tested both in sandbox.

References #241

## Steps to Test or Reproduce
Test it on my server (Eluded TDM)

- Play as knight
- Throw a water bomb upwards so it lands just to the left/right of you
- Notice that if you're shielding in the right direction then the stun is halved

### DIFF
--- a/Entities/Characters/Scripts/RunnerKnock.as
+++ b/Entities/Characters/Scripts/RunnerKnock.as
@@ -108,9 +108,16 @@ f32 onHit(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@ hitt
 		bool has_sponge = sponge !is null;
 		bool wet_sponge = false;
 
-		bool undefended = (force || !this.hasTag("shielded"));
-		if ((customData == Hitters::water_stun && undefended) ||
-		        customData == Hitters::water_stun_force)
+		bool defended = this.hasTag("shielded");
+
+		// If the class has a shield (check for ShieldVars) then check that the shield is pointing in the right direction
+		if (getShieldVars(this) !is null) {
+			if (!blockAttack(this, velocity, damage))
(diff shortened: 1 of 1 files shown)


Vikhenzo:
Would have to test thoroughly in a test build enviroment to come to a decent conclusion.

benjamingorman:
If anyone would like to test the changes in game then I'm hosting the mod on 'Eluded TDM'.

terracraft321:
If it works, it works. 👍

benjamingorman:
Agreed Biurza. I've changed it so that you can't shield your own water bombs.

Blubahub:
Well, as long as you can't shield your own water bombs, then I totally agree with adding this (as is, and without playing Eluded TDM - will play soon).

1bardesign:
I like this idea; it's a community-viable change to water that'll involve less knight worry.

Agree completely about the self stuns needing to be unaffected though.

If you can get that fix in, I'm happy - but it could be a good idea to take this to `#democracy` so folks are aware of what's going on and can give any other feedback required.

benjamingorman:
It's already in Geti - in the most recent commit.

Franrek:
The fact how much water is actually used in pub makes it pretty clear that it's not op by any means. It may be annoying yes, stunning mechanics are always annoying but not op for sure. There's no reason to nerf them so why would we?

1bardesign:
`#democracy` result: 66 votes for to 10 against

PR #296, opened 2018-12-11
Removed builder quickswap

### LINKED ISSUE #294: Remove quickswap for builders
Quickswap as builder allows you to swap to your previously selected block by quickly tapping F. The same thing has been implemented for knight and archer to easily select different kind of bombs/arrows.

I would like anyone reading this to consider removing this feature for builders. Experienced players who know the "builder inventory pattern" will automatically select blocks really fast, but quickswap prevents you from doing this. It will give you the last block you selected instead of the block you actually wanted, which can be really annoying when you need a certain type of block at a certain time.

Also from what I've heard, quickswap was implemented for new players to make builder-life easier. But I'm pretty sure not many new players even know about this feature and/or use it. Simply because it's not that helpful for building since most structures require more than 2 different blocks. The way I found out that quickswap was implemented was when I kept getting the wrong blocks I selected. I've discussed this topic with some other builder mains and they seem to have a problem with this aswell.

Any ideas / thoughts ?

eps0003:
## Status

**READY**

## Description

Sorry @MrPotatobird 😞 
Resolves #294 

## Steps to Test or Reproduce

Tap F to quick switch between two blocks

### DIFF
--- a/Entities/Characters/Builder/BuilderInventory.as
+++ b/Entities/Characters/Builder/BuilderInventory.as
@@ -72,8 +72,8 @@ void onInit(CInventory@ this)
 	blob.set_u8("buildblob", 255);
 	blob.set_TileType("buildtile", 0);
 
-	blob.set_u8("current block", 255); // 255 for no block
-	blob.set_u8("prev block", 255);
+	// blob.set_u8("current block", 255); // 255 for no block
+	// blob.set_u8("prev block", 255);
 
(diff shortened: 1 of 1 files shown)


1bardesign:
If this is going to be switched back and forward it'd be great to have a simple toggle `const bool quickSwapEnabled` or similar in the script tbh rather than commented-out code.

PR #298, opened 2018-12-11
Improved keg functionality

### LINKED ISSUE #279: [Feature Pitch] Tweaks to Kegs.
- water buckets deactivate lit kegs
- fire and fire arrows activate kegs

having some small counter-play within the way kegs are played currently would be great for introducing more strategy into the way kegs are currently played with. Currently, if a keg is lit and thrown from a tower the only way to defend from it is either a lucky catch or just running and hoping you don't die. Having water buckets (an object that can be thrown for a ranged splash) means that teams will have to work together & focus on using kegs to the fullest potential; ie; playing as a team. 

I chose buckets because having builders be the ones to defuse kegs seems to me like the most balanced way of implementing this feature. 

The fire arrows on the other hand serve as ways to punish teams for leaving kegs out in the open (or dropping them) which encourages the careful use of kegs (as opposed to just spamming them / blindly running in with them) 

Hopefully these changes encourage smart thinking, teamwork, and above all better quality matches.

eps0003:
## Status

**READY**

## Description

Improvements:
- Fire/fire arrows activate (arrows only collide with unlit enemy kegs and any lit keg)
- Water buckets deactivate (only water buckets to still allow launching kegs with water bombs)
- Explosion chain reaction
- Invincible to other keg explosions (to allow for chain reaction)
- Fixed weird fuse sound when relighting keg after it was frozen by admin

10-1 in favour of these changes. People were indecisive whether the fire and/or water changes were good. This will most likely still require a #democracy vote in Discord since this changes the most powerful weapon in kag.

Closes #279

## Steps to Test or Reproduce

!keg

### DIFF
--- a/Entities/Items/Explosives/KegVoodoo.as
+++ b/Entities/Items/Explosives/KegVoodoo.as
@@ -69,6 +69,8 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
 		this.SetLight(true);
 		this.SetLightRadius(this.get_f32("explosive_radius") * 0.5f);
 		this.getSprite().SetEmitSound("/Sparkle.ogg");
+		this.getSprite().SetEmitSoundSpeed(1.0f);
+		this.getSprite().SetEmitSoundVolume(1.0f);
 		this.getSprite().SetEmitSoundPaused(false);
 	}
 	else if (cmd == this.getCommandID("deactivate"))
--- a/Entities/Items/Explosives/Keg.as
+++ b/Entities/Items/Explosives/Keg.as
@@ -57,6 +57,14 @@ void onTick(CSprite@ this)
 
 }
 
+void onTick(CBlob@ this)
+{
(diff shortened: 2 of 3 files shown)


Entities/Items/Explosives/KegVoodoo.as
```diff
@@ -69,6 +69,8 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
 		this.SetLight(true);
 		this.SetLightRadius(this.get_f32("explosive_radius") * 0.5f);
 		this.getSprite().SetEmitSound("/Sparkle.ogg");
+		this.getSprite().SetEmitSoundSpeed(1.0f);
+		this.getSprite().SetEmitSoundVolume(1.0f);
```
AsuMagic: Why is this required?

Vam-Jam: 'Fixed weird fuse sound when relighting keg after it was frozen by admin'
Probably to fix this ^

PR #303, opened 2018-12-13
Factory Named Properly

Blubahub:
In TTH at the moment, (generic aka default) factories are called "Workshops" - the same for the default building in CTF/Sandbox. All the other specific factories do have "factory" in their name, though. When I first started playing KAG (specifically Take The Halls), I got confused by this, thinking I could build CTF shops, wanting to make a builder shop. So, I wasted my un-re-gainable mats on the factory when I should've built a Workbench. So I want to fix this, even though it'll mostly just help KAG beginners.

So, I've changed the factory's name in the Builder's menu to "Factory," and I also added a short description for the !factory and !workbench structures (in the building menu).

### DIFF
--- a/Entities/Characters/Builder/CommonBuilderBlocks.as
+++ b/Entities/Characters/Builder/CommonBuilderBlocks.as
@@ -107,14 +107,14 @@ void addCommonBuilderBlocks(BuildBlock[][]@ blocks)
 	else if(TTH)
 	{
 		{
-			BuildBlock b(0, "factory", "$building$", "Workshop");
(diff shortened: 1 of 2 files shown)


Blubahub:
I just updated Factory.as, since I realized in two of the translated strings it says "Convert workshop" instead of "Convert factory." Just noting since it would need a proper i.e. Russian translation later on.

1bardesign:
@Diprog @AsuMagic @Vikhenzo @Suprejso this will involve some minor translation tweaks when you're next updating translations! (feel free to tag any other translators or let me know if i included anyone wrong)

PR #304, opened 2018-12-13
More Fireplace Produce

Blubahub:
Right now, the campfire (!fireplace) will only "cook" aka convert the fishy blob into a food blob with the texture of a cooked fish (with the tooltip name "Cooked Fish") food blob. In this PR I extend that functionality to steaks (creates "Cooked Steaks"), grain ("Bread"), and eggs ("Cake").

"Half" of this PR is just a visual difference (Cooked Steak), while the rest are actual mechanical benefits (Bread, Cake) for the player. Hopefully, this will encourage the player to use the fireplace more, even when it is located in a spot where using it to light fire arrows would be useless..

### DIFF
--- a/Entities/Industry/Fireplace/Fireplace.as
+++ b/Entities/Industry/Fireplace/Fireplace.as
@@ -48,10 +48,38 @@ void onCollision(CBlob@ this, CBlob@ blob, bool solid)
 {
 	if (blob !is null)
 	{
+		// Convert items (and animals) into food blobs (different textures and names from your regular !food)
 		if (blob.getName() == "fishy" && this.getSprite().isAnimation("fire"))
 		{
 			blob.getSprite().PlaySound("SparkleShort.ogg");
-			CBlob@ food = server_MakeFood(blob.getPosition(), "Cooked Fish", 1);
+			CBlob@ food = server_MakeFood(blob.getPosition(), "Cooked Fish", 1); // Cooked Fish
+			if (food !is null) {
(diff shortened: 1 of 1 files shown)


Entities/Industry/Fireplace/Fireplace.as
```diff
@@ -48,10 +48,38 @@ void onCollision(CBlob@ this, CBlob@ blob, bool solid)
 {
 	if (blob !is null)
 	{
+		// Convert items (and animals) into food blobs (different textures and names from your regular !food)
 		if (blob.getName() == "fishy" && this.getSprite().isAnimation("fire"))
 		{
 			blob.getSprite().PlaySound("SparkleShort.ogg");
-			CBlob@ food = server_MakeFood(blob.getPosition(), "Cooked Fish", 1);
+			CBlob@ food = server_MakeFood(blob.getPosition(), "Cooked Fish", 1); // Cooked Fish
```
1bardesign: I feel like the repetition here could be reduced quite a lot. Consider checking the right animation ahead of all the name checks, a `bool cooked` above all of this, and killing the blob, setting the food velocity, and playing the sound if it _was_ cooked. Should cut the amount of code in half and make it much more future proof.

1bardesign: (to be clear, the line highlighted is just the start of a big highly repetitive section of code, not a particularly problematic line)

Blubahub:
I'm busy with other stuff at the moment right now; Epsilon will be working on this now (will be new PR, possibly with other fireplace changes by him)...

PR #307, opened 2018-12-15
Archer quiver shows arrow types

eps0003:
## Status

**IN DEVELOPMENT**

## Description

The feathers of arrows in archers' quivers are colour coded depending on the types of arrows in your inventory. This is the probably best way to show arrow types without the quiver looking horrible or having the arrows upside down in the quiver.

Currently, if you have no normal arrows and some special arrows in your inventory, there are gaps in the arrows in the quiver. This can be changed by reading the comment in the code (ArcherAnim.as line 645).

Colours:
- Normal = white (neutral colour)
- Water = blue
- Fire = red
- Bomb = gray

I feel like my method of implementing this isn't the best way so please let me know how I can improve this.

## Steps to Test or Reproduce

Have different types of arrows in your inventory


eps0003:
Feedback from #development includes:
- Only allow teammates to see special arrows
- Only allow people to see if you have a special arrow, but not which types arrows you have (different colour / different feather size)

PR #309, opened 2018-12-15
Small mine changes

ferrezinhre:
**Status**
Ready

**Description:**
Priming time of mine lowered form 45 to 30, mines "open" mid air.

**Steps to Test or Reproduce:**
throw mine

### DIFF
--- a/Entities/Characters/Knight/Knight.cfg
+++ b/Entities/Characters/Knight/Knight.cfg
@@ -4,9 +4,9 @@
 
 $sprite_factory                                   = generic_sprite
 
-@$sprite_scripts                                  = StandardPickup.as;
+@$sprite_scripts                                  = KnightAnim.as;
+                                                  StandardPickup.as;
                                                   HighlightItems.as;
-                                                  KnightAnim.as;
                                                   FireAnim.as;
                                                   BlobPlacement.as;
                                                   RunnerHoverHUD.as;
@@ -303,8 +303,9 @@ $name                                             = knight
                                                     ClimbArrows.as;
                                                     SeatHop.as;
                                                     KnightLogic.as;
-                                                    KnightSwordCling.as;
+                                                    KnightAnim.as;
                                                     KnightAutoPickup.as;
+                                                    KnightSwordCling.as;
                                                     BlobPlacement.as;
                                                     RunnerCollision.as;
                                                     IsFlammable.as;
@@ -318,7 +319,6 @@ $name                                             = knight
                                                     FallDamage.as;
                                                     Stomp.as;
                                                     RunnerHead.as;
-                                                    KnightAnim.as;
                                                     FallSounds.as;
                                                     RedFlash.as;
                                                     HurtOnCollide.as;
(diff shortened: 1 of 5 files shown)


benjamingorman:
Mines are already strong. I think this is a significant buff to mines without any explanation. Ferrez maybe you could argue why you think mines should be stronger?

ferrezinhre:
My dear, it will be pleasure to explain it to you, however I afraid it will not change your mind. Indeed you have right by saying that mines are strong, of course if both of us think about same thing. For now let's focus on mines as tool.

Mines have very situational uses, it's long term investment with small chances of profit, on other hand it's great tool for creating a fear factor. Here is example for how fear factor works in practice "Your team mate die from mine hidden in bush, - what you do? you slow down wondering if there is another mine".

Mines in CTF.
How we use mines in this gamemode? We place them in tunnels to stop rats, we place them in flagrooms, we hide them on battlefield, there is many other possibilities. I recently talked with Epsilon and Franek about their thoughts and fears , and yes I remember players throwing mines form towers. It is still possible just require more work. Back then it wasn't working so excellent, price of mines wasn't equal to coin income from kills, which require coin farmnig with platforms, and this is solved already. How do we deal with mines in ctf? Builders can safely disarm them, archers can shoot it with arrow, knights can slash it, you can throw body or force player to jump on it, use chicken, oh right knighs can shield (unless it bug) mine explosion. And you always can pickup enemy mine. I think thats all for ctf.

Mines in TDM.
Well I can't say here much, mines are tool like everything else, and you must be really creative to get kill anyone. Anyway one mine can sow fear and paranoia in enemies forcing them to check every bush they see. But still you need time to hide them.

It's time for a summary, my dear. Yes mines are strong if you think about damage what it can cause, but overall functionality and chances of scoring kill with it are not enough. We have item that can use anyone, this changes are to make mines more versatile, I don't ask you or anybody to agreed with me.



ps. for bunnie: shieldable mines is bad idea.
pss. I think thats all.

bunniewormy:
shieldable mines is a good idea - they are much more fun to use by builders and archers without being op - especially if they can activate in sky. you can throw one in the air while running away from enemies and theres a chance they will die. it was never a problem of too many coins, even if its hard to get money, mine would just be too op if it was unshieldable with your proposed changes

Entities/Characters/Knight/Knight.cfg
```diff
@@ -318,7 +319,6 @@ $name                                             = knight
                                                     FallDamage.as;
                                                     Stomp.as;
                                                     RunnerHead.as;
-                                                    KnightAnim.as;
```
AsuMagic: There should not be changes on this file, is that a leftover from another PR?

AsuMagic:
I also disagree with the changes for the same reasons as Eluded.

daskew87:
I personally am uncertain either way about the inherent merits of this change. I think mines are slightly under-powered in their present condition.. but not by much.

Frankulushus:
I think they're underpowered given how easy they are to pick back up after placement. It takes 4 seconds until it activates? reduce it by 2 and i'd say they're pretty balanced.
Shield-able mines were fun but would underpower them by far too much.

1bardesign:
Closing as a result of `#democracy` vote failing to pass - 47 :+1: vs 84 :-1:

PR #311, opened 2018-12-16
Christmas cheer

Mazey:
## Status

** HOLD **

## Description

Add Christmas holiday cheer!

## Steps to Test or Reproduce

Join any server during Christmas

Reason of on hold: 
1) trees created in onRestart will not have proper sectors after nextmapping; I believe Vamist has a fix we could use for this Christmas (but it changes vanilla more than I would like..)
2) all the christmas sprites are gone on nextmap (except trees, but they have another issue)
3) bushes sprite is not changing

I'm out of time so I hope someone else can take a look so we can have this for Christmas.

To-Do: Force everyone's head to be santa head

Special thanks to:
Franek for sprites
Golden for code
Vamist for code related help in #modding 

<details> 
  <summary>spoiler: changes</summary>

- Winter/Christmas textures
- Snow
- In CTF and TTH: every 5 minutes a maximum of 2 gifts appear on the map; one on each half.
- 

Preview:
image

</details>


1bardesign:
It's looking lovely - the issues with nextmap will need to be resolved long term.

2 options in the short term imo:
- work for a fix ahead of the end of this week
- we just replace the sprites in vanilla for now, revert later, and rework it to work "properly" with holidays at that point.

I think the 2nd option is probably the least painful/stressful as we likely need to get this through the pipe in less than a week.

Mazey:
Snow and trees are fixed so just 2 more issues:
-Bush sprite not being replaced (is it being done on server?)
-On nextmap, the background and tilemap are reset to default

Pretty straightforward issues, maybe someone can take a look or we will have replace vanilla sprites for next build and only include the code-based stuff in the holiday script.

1bardesign:
Integrated this separately 👍 sorry for the missing git-side credits. Everyone involved is still welcome to a github accolade if they'd like it!

PR #314, opened 2018-12-18
Taunt wheel

### LINKED ISSUE #200: [Feature Pitch] add Taunts, like in Classic

eps0003:
## Status

**READY-ish**

## Description

This adds taunts just like in classic!

Summary:
- Taunt key displays taunt menu which provides quick access to common phrases/taunts
- Potentially customizable taunts
- Posts taunts in global/team chat depending on the category
- Cooldown to prevent spamming (shorter for team taunts, can be adjusted in code)
- Prevent using the same taunt in a row (currently disabled, can be changed in code)

Closes #200

## Steps to Test or Reproduce

Can't currently test since the taunt button doesn't work


Shylie:
I believe you can add text to chat script-side. Check global functions, rather than `CPlayer` or whatever methods

eps0003:
It seems like people would prefer it if taunts weren't posted in chat. It would clutter the chat and lead to spam. I don't want to add a cooldown for taunts since that would be quite restrictive.

I've managed to use `blob.Chat()` in a hacky way to allow you to see other peoples taunts, but I'm struggling to make them appear only once for each taunt.

eps0003:
This is ready to go now, other than the fact that the taunt key is still not working.
I added spam cooldown and repeat taunt vars at the top of `TauntMenu.as` if you want to modify them.
This supports custom taunts but I'm still not sure what prevents the cfg file from being overwritten when joining a server.

1bardesign:
(This has conflicts now, going to need master merged in and conflicts resolved before we can merge it)

eps0003:
Resolved conflicts

PR #324, opened 2018-12-22
Adjusted some pickup priorities

Shylie:
## Status

- **READY**

## Description

Lit kegs now are even more prioritized, and unarmed enemy mines are now a priority as well.

## Steps to Test or Reproduce

- drop a bunch of stuff on ground
- light a keg and try to pick it up with tap
- drop a mine and pick it up as a different team


1bardesign:
Might be worth looking at extending @eps0003 's ingame binding stuff to tinker with the individual priorities, if anyone is going to take this much further. Has been requested multiple times.

PR #330, opened 2018-12-26
Temporary fix for missing knight animations

ferrezinhre:
## Status

**READY**

## Description

A small change that is a temporary fix for missing knight animations, until someone find right way how solve this problem.

## Steps to Test or Reproduce
uh, umm, fighting as knight?
_was tested on eu captains._

### DIFF
--- a/Entities/Characters/Knight/Knight.cfg
+++ b/Entities/Characters/Knight/Knight.cfg
@@ -4,9 +4,9 @@
 
 $sprite_factory                                   = generic_sprite
 
-@$sprite_scripts                                  = StandardPickup.as;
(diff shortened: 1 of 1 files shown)


bunniewormy:
hot, fuzzle was right about 1 line fix

Franrek:
Does this shit really works?

1bardesign:
How come it's added twice?

1bardesign:
I see, no problem.

Seems unlikely to be a very successful fix (afaict a proper fix would require actually sending all the knight timers with each delta and syncing the animations up to those) - but if it improves things, I'm happy to merge it through to test at least.

PR #331, opened 2018-12-27
Added player heads to scoreboard

### LINKED ISSUE #321: Feature Pitch: Adding heads to the scoreboard
Hello everyone, it's Sheepo!

I suggest adding a small icon of the head between the name and the class icon (on the green area), so you can see the face of everyone. That would make the heads more interesting and the menu would look more complete.

heads

Thanks everyone for reading.

eps0003:
## Status

**READY**

## Description

- Adds player heads to the scoreboard
- Works with head packs, holiday, accolade and special heads
- Slightly increased padding between players on the scoreboard so heads don't overlap as much
- Includes change to ensure KDR always has two decimal places

Closes #321

## Steps to Test or Reproduce

Press tab

### DIFF
--- a/Rules/CommonScripts/ScoreboardRender.as
+++ b/Rules/CommonScripts/ScoreboardRender.as
@@ -1,6 +1,5 @@
 #include "ScoreboardCommon.as";
 #include "Accolades.as";
-#include "RunnerHead.as";
 
 CPlayer@ hoveredPlayer;
 Vec2f hoveredPos;
@@ -41,7 +40,7 @@ string[] age_description = {
 //returns the bottom
 float drawScoreboard(CPlayer@[] players, Vec2f topleft, CTeam@ team, Vec2f emblem)
 {
-	if (players.size() <= 0)
+	if (players.size() <= 0 || team is null)
 		return topleft.y;
 	Vec2f orig = topleft; //save for later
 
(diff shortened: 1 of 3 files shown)


Rules/CommonScripts/ScoreboardRender.as
```diff
@@ -1,5 +1,6 @@
 #include "ScoreboardCommon.as";
 #include "Accolades.as";
+#include "RunnerHead.as";
```
1bardesign: This include still needs to be removed before this PR can be merged I'm afraid!

eps0003: Oh I thought you would remove it when merging. I’ll do that ASAP.

eps0003: Just remembered that it's needed to get the heads pack index on line 154. Is the reason you want it removed since you thought it was not needed?

1bardesign: No it's that it'll pull in extra hooks and a heap of extra code for what is essentially getting one string (the heads sheet) and one integer (the head frame) 😅

A solution might be storing these in a straightforward blob property (so mods can also use the functionality easily) and displaying a skull (or nothing) when the player's blob is dead.

eps0003:
Fixed the issue outlined in @1bardesign's review

1bardesign:
from `#development`:

> am still cagey about this as it uses player properties - i know you aren't syncing them so there shouldn't be an issue but i'm still like :grimacing: because you might end up fetching empty props from there upon render
> 
> i wonder about just getting them from the blob and having no head or 💀 if you dont have a blob (maybe depending on spectator state?), and blank if the blob doesn't have head properties
> 
> if you're happy with that i can do the changes next week so you dont have to touch it any more

PR #333, opened 2018-12-27
Fixed trampoline flying glitch

eps0003:
## Status

**READY**

## Description

Finally fixes a long-standing bug which allows knights to fly using trampolines.

## Steps to Test or Reproduce

Follow instructions in [this](https://forum.thd.vg/threads/25571/) bug report

### DIFF
--- a/Entities/Structures/Trampoline/TrampolineLogic.as
+++ b/Entities/Structures/Trampoline/TrampolineLogic.as
@@ -62,6 +62,10 @@ void onCollision(CBlob@ this, CBlob@ blob, bool solid, Vec2f normal, Vec2f point
 	//cant bounce while held by something attached to something else
 	if (holder !is null && holder.isAttached()) return;
 
+	//prevent knights from flying using trampolines
+	if (blob.getPosition().y > this.getPosition().y && this.getAngleDegrees() == 0)
+		return;
+
 	TrampolineCooldown@[]@ cooldowns;
 	if(!this.get(Trampoline::TIMER, @cooldowns)) return;
 


benjamingorman:
Downvoting because I like being able to fly

Franrek:
Updoot coz I'm scared of heights

benjamingorman:
For anyone that hasn't seen it in game I made a video of this ~bug~ game feature a while ago https://youtu.be/MxbaohqszxU

Entities/Structures/Trampoline/TrampolineLogic.as
```diff
@@ -62,6 +62,10 @@ void onCollision(CBlob@ this, CBlob@ blob, bool solid, Vec2f normal, Vec2f point
 	//cant bounce while held by something attached to something else
 	if (holder !is null && holder.isAttached()) return;
 
+	//prevent knights from flying using trampolines
+	if (blob.getPosition().y > this.getPosition().y && this.getAngleDegrees() == 0)
```
AsuMagic: Unless I'm reading this incorrectly, doesn't that mean that this also applies to any horizontal trampoline?

eps0003: That’s what the angle check is for.

1bardesign: I think @AsuMagic 's concern is that this would prevent bouncing on horizontal tramps - it's when "bouncing" from below though, right? That will break some tramp "elevators" on TDM levels.

Could be better to require an upwards velocity rather than being above the tramp as a workaround that preserves the fix?

PR #340, opened 2018-12-30
More fireplace improvements

eps0003:
## Status

**READY**

## Description

- More cooked food options (improved #304 code)
  - Steak → Cooked Steak
  - Grain → Bread
  - Egg → Cake
- Fixed issue where fireplaces were destroyed upon spawning in TDM (thanks @Coroz)
  - Caused by them falling slightly and destroying on impact
  - Removed fall damage to fix this
- Becomes static when it lands on solid ground
- Increased health

## Steps to Test or Reproduce

!fireplace
/rcon /loadmap Niiiiii_Watery.png

### DIFF
--- a/Entities/Industry/Fireplace/Fireplace.as
+++ b/Entities/Industry/Fireplace/Fireplace.as
@@ -8,7 +8,8 @@
 
 void onInit(CBlob@ this)
 {
-	this.getShape().getConsts().mapCollisions = false;
+	// this.getShape().getConsts().mapCollisions = false;
+	this.getShape().getConsts().rotates = false;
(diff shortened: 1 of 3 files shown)


1bardesign:
Looks mostly good 👍 

They should probably destroy upon being under-mined though? I imagine that's what the fall damage was meant to accomplish.

eps0003:
@1bardesign as we discussed on Discord, it was inconsistent whether fireplaces fell/destroyed on no support. That's why I just made it become static once settled.

Orsovi:
This could be great if steak/grain and grain wouldn't give already all your life ; otherwise, it's not really an usefull change.

PR #349, opened 2019-01-08
Voting changes

eps0003:
## Status

**READY**

## Description

- Prevent player from voting on a vote they started (admins can still force pass/cancel the vote)
- Prevent player from voting on a vote to kick them
- Prevent player from starting a vote against themselves
- Changed pass percentage back to 50% since the vote starter can't vote
- Changed what's visible in the vote box to reflect these changes

## Steps to Test or Reproduce

Vote/be voted against using players with different roles (normal, admin, THD team)


1bardesign:
Accepted by `#democracy` - 118 :+1: to 10 :-1: - Seems community on discord more positive than on here 😉 will still need some careful testing to make sure there's no new edge cases with any votes.

PR #352, opened 2019-01-08
Catapult Rocks & Ballista Bolts Pass Through Platforms Fix

MonkeyFeats:
## Status
READY

## Description
Allows catapulted rocks to pass through platforms from the backside and still hit from the front.
Adds a short period of time before catapulted rocks will hit back_wall. (prevents destroying own castle)
Allows ballista bolts to pass through platforms from the backside and damage the front.

## Steps to Test or Reproduce
Fire a catapult or a ballista at platforms facing different angles.


MonkeyFeats:
Oh lol didn't fix the Ballista bolts. Whoops.

PR #353, opened 2019-01-08
Shop keybinds

### LINKED ISSUE #240: [Feature Pitch] If you're standing next a shop, number keys buy items instead of emoting
This would make it easier to buy items quickly. Buying items at the minute is slightly awkward because you have to be quite precise with your clicks, and it's easy to buy the wrong item. This feature pitch would also help in situations where you need to buy something extremely quickly (e.g. you're about to die if you don't buy a burger).

I think it could be implemented by mapping every slot in the shop to one of the number keys. If there were more than 10 slots then shift+number, ctrl+number etc. could be used to handle a large number of items.

eps0003:
## Status

**READY-ish**

## Description

This adds the ability to buy items from shops by holding shift and pressing a number key. The keybinds match shop buttons from left to right, top to bottom. This only works once you've clicked 'e' on a shop to prevent accidental purchases. If a shop has more than nine items (which none do at the moment), there will only be keybinds for the first nine items. This also works on the buildings which allow you to choose the shop you want.

Currently, there is an engine-side issue with `CGridMenu.AddKeyCommand()` where it always seems to use the default params. I will post the fix for it once I properly setup engine-side stuff.

Closes #240

## Steps to Test or Reproduce

Press 'e' on a shop, hold shift and press a number key for the item you want


eps0003:
@GoldenGuy I've done that

GoldenGuy:
awesome

Orsovi:
Be care for compatibility, on french keyboard and linux (yes i know that's not a concern of lot of people) you already have to make shift + key to make any emote.

eps0003:
@Orsovi I have no way of testing for compatibility between keyboards. I don’t even know if the devs intended for emotes to work using shift on non-qwerty keyboards.

Entities/Common/Factory/Shop.as
```diff
@@ -91,201 +92,214 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
 	}
 	else if (cmd == this.getCommandID("shop buy"))
 	{
-		if (this.hasTag("shop disabled"))
-			return;
+		BuyShopItem(this, params);
+	}
+	else if (cmd == this.getCommandID("shop buy hotkey"))
+	{
+		CControls@ controls = getControls();
```
1bardesign: I don't think this will work on server - controls object either wont exist or at least wont be the same as the player's (how could it be, for every player). You'll need to write the command to buy clientside and find some other way to sanity-check serverside.

1bardesign: @eps0003 any thoughts on this? gonna need a workaround before it can be merged unfortunately

daskew87:
I like this idea as I love to use shortcut keys. @JayP9 At first mate it will be slower to buy items this way, but after a while of getting used to it, it should be slightly faster.

I currently use the shift key for accessing my inventory.. @eps0003  Will we have the option to change this function to another key in the control options?

eps0003:
@daskew87 I’m unable to prevent emotes from displaying if I don’t use shift while pressing a shop keybinds. So no, I don’t plan to add an option for it. 

And what’s with foreign keyboards requiring you to use shift to do things?

1bardesign:
@eps0003 would it be possible to add an optional config file for the few who do use shift to move it to an alternative key? I imagine quite a few folks with custom binds would end up using shift for _something_. Would love to get this merged through soon.

eps0003:
I don’t know how to stop emotes from showing if I don’t use shift. Any ideas?

And I ran into issues so I took a break from this PR. I’ll check it out again soon.

1bardesign:
No worries, understandable.

Could have emotes check for that key from the same config file probably, or
set a tag blocking emotes while you're interacting with a shop or something
👍

PR #356, opened 2019-01-11
Close range arrow stun duration change

eps0003:
## Status

**READY**

## Description

@Franrek wants to increase arrow stuns duration by 5 ticks (1/6sec) if not shielded.

It would make it possible for archer to sneak in 0.5hp damage using quarter-shot after successful stun. Exactly like knight does 1hp damage with jab after slashing or shieldstunnig enemy, except harder to execute since you need to charge it and get the timing right.

The change would allow archers for more aggressive play instead of the strange "foward and back" dance with tripleshot which is completely stupid and boring. Like that archers can risk more to deal more damage making the archer-knight combat more interesting and less stale for both classes.

The 0.5hp sneak in after shieldstun was possible before but there was really no point in using it without 0ping since it would just not work most of the times.

## Steps to Test or Reproduce

Shoot an enemy knight at close range while they are shielding and not shielding

### DIFF
--- a/Entities/Items/Projectiles/Arrow.as
+++ b/Entities/Items/Projectiles/Arrow.as
@@ -732,13 +732,16 @@ void onHitBlob(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@
 
 			hitBlob.AddForce(velocity * force);
 
+			// check if shielded
+			const bool hitShield = hitBlob.hasTag("shielded") && blockAttack(hitBlob, vel, 0.0f);
+
 			// stun if shot real close
 			if (
 				this.getTickSinceCreated() <= 4 &&
 				speed > ArcherParams::shoot_max_vel * 0.845f &&
 				hitBlob.hasTag("player")
 			) {
-				SetKnocked(hitBlob, 20);
+				SetKnocked(hitBlob, hitShield ? 25 : 35);
 				Sound::Play("/Stun", hitBlob.getPosition(), 1.0f, this.getSexNum() == 0 ? 1.0f : 1.5f);
 			}
 		}
--- a/Entities/Items/Projectiles/Arrow.as
+++ b/Entities/Items/Projectiles/Arrow.as
@@ -741,7 +741,7 @@ void onHitBlob(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@
 				speed > ArcherParams::shoot_max_vel * 0.845f &&
 				hitBlob.hasTag("player")
 			) {
-				SetKnocked(hitBlob, hitShield ? 25 : 35);
+				SetKnocked(hitBlob, hitShield ? 20 : 25);
 				Sound::Play("/Stun", hitBlob.getPosition(), 1.0f, this.getSexNum() == 0 ? 1.0f : 1.5f);
 			}
 		}


Panky123:
Stop trying to make bugs features please. Beyond that it really isnt fair to allowed a ranged class more strength in close combat. The arrows already pierce/stun 5.5 tiles away. Buffing that would be a serious headache for any knight actually trying to kill an archer that can still just RUN AWAY because it is ranged.

Franrek:
I've realized how op would it be after some more testing with better archers, i still think its good idea to reward offensive archer Play more than running away all the time and camping with grapple. Coz its just not fun for both classes. Edited it to ONLY increase stun when not shielded by 5 ticks.
Archer stopped being truly long range class when the stun for full shot was added
Also Jabs do cancel archer charge

Get ur facts right >:^(

Panky123:
Increasing stuns freaking sucks. Not only will it be awful as knight but builders will have less of a fighting chance. I'd love for the game to have less campy archer combat but you arent gonna fix that by just buffing archers outright. The class being slightly stronger in close combat wont stop archers from still running and having the ranged advantage(and if they chip the knight down enough an even stronger close combat game). It is insane to me that you think knights have a better chance at just running away when being at range is more of a negative for a knight(they cant do any damage unless they have special items) . We talked about it in #development and you cleared up my confusion on archer attacks but i still feel like the result is the same. Archers shoot arrows earlier than knights and the knight slash is instant? This reads as though it is somehow unfair to the archers when in reality it shows that because archers can initiate ranged attacks before a knight can rush and reach them(to do damage) they have a pretty fair niche in close combat. The most frustrating part about knight vs archer combat is that it is solely dependent on the archer making any mistakes. If they keep their distance and chip down the knight they win, if they mess up they die. This change is pretty significant to 1 vs 1 between knight and archer where the advantage solely lies with the archer unless the knight has bombs. Archers should get destroyed up close and the only advantage the stun should have is just enough time to escape, NOT to do any more attacks. Itd be ridiculous to deal with as knight and even more frustrating as builder.

Franrek:
> only advantage the stun should have is just enough time to escape

If you really think archer should be range class only and they should always keep the distance for best results the stun from close range triple shot or even grapple mechanics don't make sense in the first place.

> even more frustrating as builder

I don't mind making it a tiny bit harder for builders, it's pretty easy for builder to kill an archer in close combat even though they aren't supposed to be the fighting class. Might make archer more useful in ctf in stead of being used as turrets for special arrows when you have 1000 coins.

> in reality it shows that because archers can initiate ranged attacks before a knight can rush and reach them

No, it shows that at the same range when both classes can reach each other knight has huge advantage in timing fights. I don't have a problem with them. But yes archer have bows so they can attack before knight can reach them, it can be shielded.


Let me get this clear, sneaking the 0.5 damage is already possible the change only makes it more consistent for people that don't have 0 ping. That's why its so subtle. 
The buff ONLY helps you to deal more damage (0.5hp) after you have already won a timing fight vs the knight. It doesn't help you win it.

I didn't think it would be that controversial since it's 5 ticks and (i think) everyone knows how often archers win vs knights and how much more useful knights are in CTF.

1bardesign:
Closing this as a result of `#democracy` vote failing to pass - 58 :+1: vs 83 :-1:

PR #378, opened 2019-02-01
Even more fireplace improvements

eps0003:
## Status

**READY**

## Description

**Changes**
- Unlit fireplaces can be picked up and put in your inventory (1x1 slot size)
- Fireplaces rotate so they don't look like they are levitating when put on the edge of a block
- Modified fireplace hitbox which makes it sit properly on the ground when upside down. It also makes it slightly harder for it to roll upside down
- Builders can hit fireplaces
- Bombs / water bombs knock around fireplaces now that fireplaces no longer become static
- Fireplaces are log-powered, except those spawned in with the map (or spawned in using commands when your team is -1)
- Added fireplace to builder shop (CTF) and workbench (TTH) which costs 100 wood and 100 stone

**Fixes**
- Fireplaces no longer become static, especially when the map loads which caused them to float
- Added missing shop items to workbench and reorganised them to match builder shop layout

**Issues**
- You can throw fireplaces into logs to fuel the fireplace which doesn't really make sense. It does make it easy to fuel the fireplace with a lot of logs though. I don't know the best way to fix this.

## Steps to Test or Reproduce

Experiment with builder shop, workbench, fireplaces and logs

### DIFF
--- a/Entities/Characters/Builder/BuilderHittable.as
+++ b/Entities/Characters/Builder/BuilderHittable.as
@@ -6,7 +6,7 @@ const string[] builder_alwayshit =
 {
 	//handmade
 	"workbench",
-	//"fireplace",
+	"fireplace",
 	"ladder",
 
 	//faketiles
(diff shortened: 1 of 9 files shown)


Blubahub:
Looks good, but fireplace needs to be 1x2 (1 tall, 2 wide) instead of 1x1.

1bardesign:
This looks pretty big so I'm definitely holding out til we've got the patreon related stuff through testing I'm afraid. 😅 

Looks interesting but probably needs testing and potentially `#democracy` about introducing them as a purchased item?

1bardesign:
This is gonna be a hard sell for some after some discussion in `#development` - can we have the floating fireplace fix in a separate PR and we might try this in a separate testing server before we think about merging/voting?

PR #387, opened 2019-02-08
Adjust grapple length

eps0003:
## Status

**READY**

## Description

This adds the ability for archers to adjust the length of their grapple. This gives them a bit better control while grappling up and around walls while barely impacting the current grapple meta. The grapple extends within the rope's current limits to prevent players from extending the rope to an 'unrealistic' length. You are able to adjust the grapple as it's being thrown and while it's attached to a block. The speed of adjustment can be changed in code if it's too slow, but I found that increasing the speed causes the player to wobble about too much.

Resolves [this](https://forum.thd.vg/threads/15981/) suggestion from 2013 lol

## Steps to Test or Reproduce

Press W and S while your grapple is out

### DIFF
--- a/Entities/Characters/Archer/ArcherCommon.as
+++ b/Entities/Characters/Archer/ArcherCommon.as
@@ -40,6 +40,8 @@ const f32 archer_grapple_force = 2.0f;
 const f32 archer_grapple_accel_limit = 1.5f;
 const f32 archer_grapple_stiffness = 0.1f;
 
+const f32 archer_grapple_min_ratio = 0.2f;
+
 namespace ArrowType
 {
 	enum type
@@ -67,6 +69,7 @@ shared class ArcherInfo
 	bool grappling;
 	u16 grapple_id;
 	f32 grapple_ratio;
+	f32 grapple_ratio_target;
 	f32 cache_angle;
 	Vec2f grapple_pos;
 	Vec2f grapple_vel;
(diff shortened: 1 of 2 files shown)


MonkeyFeats:
I'm a divided on this, while I think it's a novel idea, I don't really like the mechanic in other games that have it. Also it might break the feel of the parkour game modes, guess it needs to be test to know for sure though.

AsuMagic:
I tried it today. After the fix, other than hanging from the ceiling or a tree at an adjustable height, the mechanic didn't interfere with regular movement and seemed to work pretty well.  
More "pro" archers should try this though so we can see whether this is useful enough and whether it does not get in the way.

1bardesign:
Yep would love input from folks who play a lot of archer in TDM and CTF.

Does it still auto-reel? if you don't press anything? I guess in practice you're probably holding W almost all the time anyway?

1bardesign:
Accepted by `#democracy` - 112 👍 to 50 👎

There have been multiple concerns raised however; we should work with the community to find a solution that suits. There are folks on both sides of the class divide with certain worries, and it'd be good for a change like this to not hit the ground with discontent in the air - just "not adding it" isn't a great option in the face of such a decisive vote though.

I'll merge it through and do a test build this evening but expect more changes to be needed to follow 😅

eps0003:
The main issue people have with this is that you're able to hang at a weird angle when the grapple is fully extended. A potential solution would be to reduce the horizontal force applied while pressing A/D keys depending on the length of the grapple. I would like people to think and propose changes for the issues they have with this before completely dismissing the PR.

eps0003:
Grapple crouch sliding would be greatly affected by this PR and idk what I can do to fix it. This PR will probably need to be scrapped without a decent solution since that's an important mechanic for pro archers. 😭

PR #392, opened 2019-02-12
Fixed drill inconsistency when starting to drill.

Shylie:
## Status

- **READY**

## Description

Drills do not always start drilling instantly when you start to drill, as the time between drilling is done by game tick, rather than a proper timer. This changes drills to use a proper timer.

## Steps to Test or Reproduce

Use a drill.


eps0003:
I haven’t thoroughly looked at this so correct me if I’m wrong, but couldn’t this allow people to spam click with a drill to dig/hit faster than the timer?

1bardesign:
( @eps0003 no, it only resets the timer if the skip is not performed, and the skip is now based on the timer, so it's a safe closed loop )

PR #393, opened 2019-02-12
Apple Trees, Apple Tree Seeds, Thin Logs, and Apples

Blubahub:
## Status

## Description

This PR adds Apples Trees, Apple Tree Seeds, Thin Logs, and Apples. All sprites by me (the Apple sprites might need to be changed).

**Apple Trees** are a slightly thinner and weaker version of Bushy (Oak) Trees that spawn 2 **Apples** of a random color upon being fully grown (these Apples will not regrow, so the only way to get more Apples is to chop the tree down and regrow it). Apples heal 1 full heart, and can be found in 3 different colors (red, lime green, yellow). These Apples can be picked off the tree (or be eaten by touching them when a player is low on hearts).

Apple Trees are uncommon to rare, and will automatically spawn using the existent tree colors on .png maps. **Apple Tree Seeds** spawn like normal tree seeds. Upon death, Apple Trees drop "**Thin Wood Logs**" (a smaller log that gives a bit less mats then normal logs) and any Apples they are still holding.

### "Why add it?"

This will allow for the first time (at least for non-Classic KAG) actual farming for food harvest. I made the sprite thinner after doing a vote on my Discord server (at the time that it officially "closed," there were more votes for a thin trunk vs. the normal "thick" trunk), but it can be modified so it is "thick" as normal trees (in that case, it would also use normal logs).

Do note that typically only 1-3 Apple Trees spawn per CTF/TTH map. Also, the Sandbox (SBX) game mode currently does NOT spawn Apple Trees.


Entities/Natural/Trees/ThinLogsOnDeath.as
```diff
@@ -0,0 +1,43 @@
...
+
+	for (uint i = 0; i < (segments.length); i++)
+	{
+		TreeSegment@ segment = segments[i];
+
+		if (getNet().isServer())
+		{
+			pos = this.getPosition() + (segment.start_pos + segment.end_pos) / 2.0f;
+			pos.y -= 4.0f; // TODO: fix logs spawning in ground
+			CBlob@ log = server_CreateBlob("log_thin", this.getTeamNum(), pos);
```
eps0003: You don't need an near-identical file to `LogsOnDeath.as` just to spawn thin logs

Vam-Jam:
Change the 'Random' chance for apples to spawn with a custom map colour, it gives map makers more options with map design, no chance of stalemate (since the current way allows one side to get more then the other, or all trees being apples etc etc)

Entities/Items/Food/EatCommon.as
```diff
@@ -1,29 +1,32 @@
...
+		params.write_u16(this.getNetworkID());
+
+		u8 heal_amount = 255; //in quarter hearts, 255 means full hp
+
+		if (food.getName() == "heart")	    // HACK
+		{
+			heal_amount = 4; // full heart
+		} else if (food.getName() == "apple")
+		{
+			heal_amount = 4; // quarter heat
```
Mazey: 4 quarter hearts is a full heart, not a quarter heart (if we should go by the comment) (also it says heat instead of heart)

Mazey: 4 quarter hearts is a full heart, not a quarter heart (if we should go by the comment) (also it says heat instead of heart)

Entities/Natural/Seed/Seed.as
```diff
@@ -1,139 +1,141 @@
...
+};
+
+// names of seeds
+const string[] seed_names =
+{
+	"Seed",           		//normal seed
+	"Grain Seed",     		//grain seed
+	"Pine Seed",       	 	//pine
+	"Oak Seed",     		//bushy
+	"Apple Tree Seed",		//apple tree
```
Mazey: Not a big deal, but the seed of an apple tree is the seed of an apple. Should probably change it for consistency as well.

Entities/Natural/Trees/AppleTreeLogic.as
```diff
@@ -0,0 +1,384 @@
...
+	{
+		for (int i = 0; i < 2; i++)
+		{	
+			switch (i)
+			{
+				// Apple 1
+				case 0: InitFirstApple(this); break;
+	
+				// Apple 2
+				case 1: InitSecondApple(this); break;
```
Mazey: Why would you do a for loop + switch statement here? Just do `InitFirstApple(this);InitSecondApple(this);`

Entities/Natural/Trees/AppleTreeLogic.as
```diff
@@ -0,0 +1,384 @@
...
+	}
+}
+
+// Apple 2
+void InitSecondApple(CBlob@ this)
+{
+	// pos
+	Vec2f offset = this.getPosition();
+	int v = this.isFacingLeft() ? 0 : 1;
+	offset = (offset + Vec2f(-8 + v, -33));
```
Mazey: This line is the only difference between the init apple functions? That is redundant. Remove one and add a second parameter such as `uint8 apple_id` -> `if (apple_id == 0) { // set offset for the first apple here` `else if (apple_id == 1) { // set offset for the secoind apple here` if you do this you *could* use the for loop on line 32 (i.e. `InitApple(this, i)`) but the switch statement still serves no purpose.

Entities/Natural/Trees/AppleTreeLogic.as
```diff
@@ -0,0 +1,384 @@
...
+	
+	// create
+	CBlob@ apple2 = server_CreateBlob('apple', -1, offset);
+	if (apple2 !is null)
+	{
+		// make it referable
+		this.set_u16("apple2", apple2.getNetworkID());
```
Mazey: Adding to the review comment above, settings the network id property would be changed to something like this after merging the init functions `this.set_u16("apple" + apple_id, apple.getNetworkID());`

Entities/Natural/Trees/AppleTreeLogic.as
```diff
@@ -0,0 +1,384 @@
...
+    if (apple1 !is null)
+    {            
+        CShape@ shape1 = apple1.getShape();
+        if (shape1 !is null)
+        {
+            shape1.SetStatic(false);
+        }
+    }
+	// Apple 2
+	CBlob@ apple2 = getBlobByNetworkID(this.get_u16("apple2"));
```
Mazey: Once again you're doing the same thing twice. Move this to a for loop and for each apple do `CBlob@ apple = getBlobByNetworkID(this.get_u16("apple" + i));`.

Entities/Natural/Trees/ThinLogsOnDeath.as
Mazey: To elaborate, `LogsOnDeath.as` makes use of the `onDie(CBlob@ this)` hook, as such you can make clever use of `getName();` ;)

Entities/Items/Food/EatCommon.as
Blubahub: I meant a full heart. Editing PR...

Entities/Natural/Seed/Seed.as
Blubahub: I kept it the same thing so there is consistency with KAG trees so beginners do not get confused.

Mazey: (It should be `"Apple Seed"`)

Entities/Natural/Trees/AppleTreeLogic.as
Blubahub: An accidentally-kept remnant of old code. Fixing...

Entities/Natural/Seed/Seed.as
Blubahub: Ahh, I thought you meant that the seeds should come from the apples themselves (I see the grammatical issue now), fixing...

Entities/Natural/Trees/AppleTreeLogic.as
Blubahub: Done.

Entities/Natural/Trees/AppleTreeLogic.as
Blubahub: Done.

1bardesign:
Please sort out conflicts in this PR by pulling from master. Looks like it might just be the include (which was an old relative path include).

Entities/Natural/Trees/ThinLogsOnDeath.as
Mazey: You've marked this as resolved but this file still exists in your current PR.

Mazey:
What are your plans with this PR @Blubahub? Indentation is still not fixed and it'd be nice if you could get rid of the duplicate LogsOnDeath.as.

bunniewormy:
I think the idea is pointless ('change for the sake of change') and that different kinds of trees may be confusing to new players.

Blubahub:
Currently fixing (not done yet)... I'll be sure to make the apple trees have a seperate map color instead of integrating the tree with the normal map color.

PR #409, opened 2019-02-20
TDM quick switch clone fix

Vam-Jam:
## Status
- **READY**

## Description
Fixes the quick switch bug by clearing menu's. The bug is caused by 'E' counting as a mouse click on the menu, so the respawn cmd would be called 2 times.

## Steps to Test or Reproduce

Load up TDM
Switch class by pressing E over the class menu


Vam-Jam:
It fixes the issue from what i've tested, if you press spam press e + mouse over the menu, it makes the clones since E sends a command to change class, and E also acts as a menu press, so it end's up sending 2 commands, so the clone is made.

Clearing the menu prevents this from happening, but im unsure if there are other ways to cause the clone bug to happen

Vam-Jam:
Okay, you are right, i re-tested with c# 'E' button spammer i made, it turns out it makes less clones before, but still made some, i got a different idea on how to prevent this.

Harrison-Miller:
Here's the fix, and I'm not sure why this tag wasn't used in the first place. In StandardRespawnCommand.as
```
bool canChangeClass(CBlob@ this, CBlob@ blob)
{
	if(blob.hasTag("switch class")) return false;

	Vec2f tl, br, _tl, _br;
	this.getShape().getBoundingRect(tl, br);
	blob.getShape().getBoundingRect(_tl, _br);
	return br.x > _tl.x
	       && br.y > _tl.y
	       && _br.x > tl.x
	       && _br.y > tl.y;

}
```
Change is to check the tag that is set.

I would also recommend modifying TentLogic.as to have the same changes as TDM_Ruins.as (isClient, MenuClear etc.)

Harrison-Miller:
The fix for this was merged internally. Closing PR now.

PR #418, opened 2019-02-22
Create steam_content_zh

zestybaby:


zestybaby:
Renamed, should have checked that better.

1bardesign:
Looks ok, at least, it's Chinese where there was English now 😅 do we have anyone who can review? I can throw it into google translate but I imagine I'll get google translate quality english out.

zestybaby:
Machine translators are always bad when it comes to Chinese and English due to massive differences between grammar. 

For reviewing this I think @Aristalia also speaks Chinese and capable of reviewing my stuff. This user has already pointed out some better suggestions in my game translation pull request.

1bardesign:
That would be excellent if you have time @Aristalia ! I am working on getting font-swapping support required for CJK and other language support in game now, would be awesome to get all these changes in Soon™

Aristalia:
I'll be on it to review more!

Aristalia:
@zestybaby is there an English original version of this that I could reference from?

zestybaby:
> @zestybaby is there an English original version of this that I could reference from?

https://github.com/transhumandesign/kag-base/blob/master/Locales/Steam/steam_content_en.txt

Issue #373, opened 2019-01-25
Quick change class with E (especially for TDM)


Orsovi:
Actually when you want to change class, you always have to open the contextual menu.
For noob players like me, when you're in rush (like a knight charging you in TDM and you're archer), you don't always have time to open the menu.

I propose to swap class with fast pressing on E (archer to knight and vice-versa for tdm and going around in CTF)

PR #419, opened 2019-02-23
Remove quick swap class from CTF

agassert:
## Status

**READY**

## Description

- Removes the quick swap class feature from CTF. 

Followup to issue #373 and PR #382.

## Rationale

- Current implementation breaks the expected flow for many experienced players
  - Many players press 'e' and expect the menu to appear regardless of cursor position (i.e. cursor doesn't have to be over the icon)
- Current implementation is not intuitive for new players
  - Mainly due to its not being consistent with the rest of KAG's UI interactions
- Feature was implemented without community feedback
  - Change is somewhat controversial and did not go through the normal vetting process for new features

I propose we leave the quick swap class feature out of CTF for the time being. We should get community feedback to see if the feature would be welcome. Voting in the #democracy channel of the official KAG Discord server seems like a good way to do this. If the feature gains significant support, alternative implementations that do not break the old flow (tap 'e' to bring up the class menu) should be discussed.

Quick swap class for TDM seems less controversial, so it was not touched in this PR (there are only two classes in TDM, so the class menu is less useful). More discussion needed.

## Steps to Test or Reproduce

Tap the use key ('e') while in front of a tent. The class change menu appears and the player's class is not quick-swapped.

### DIFF
--- a/Entities/Special/Tent/TentLogic.as
+++ b/Entities/Special/Tent/TentLogic.as
@@ -1,7 +1,6 @@
 // Tent logic
 
 #include "StandardRespawnCommand.as"
-#include "StandardControlsCommon.as"
 
 void onInit(CBlob@ this)
 {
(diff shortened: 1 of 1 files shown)


eps0003:
Instead of removing the code, you should have a variable at the top of the script so it can easily be re-enabled if need be.

Mazey:
@eps0003 what might make even more sense is to make quickswap an include

Harrison-Miller:
tbh don't see how making it an include makes it easier to enable or disable it. I think cycleClass is fine where it is for the most part...

Harrison-Miller:
Was resolved in another commit, with the above method.

PR #420, opened 2019-02-24
Shop keybinds

### LINKED ISSUE #240: [Feature Pitch] If you're standing next a shop, number keys buy items instead of emoting

eps0003:
## Status

**READY**

## Description

This adds the ability to buy items from shops by pressing a number key. The keybinds match shop buttons from left to right, top to bottom. This only works once you've clicked 'e' on a shop to prevent accidental purchases. If a shop has more than nine items (which none do at the moment), there will only be keybinds for the first nine items. This also works on the buildings which allow you to choose the shop you want.

Closes #240

## Steps to Test or Reproduce

Press 'e' on a shop and press a number key for the item you want

### DIFF
--- a/Entities/Common/Emotes/EmoteHotkeys.as
+++ b/Entities/Common/Emotes/EmoteHotkeys.as
@@ -18,6 +18,8 @@ void onInit(CBlob@ this)
 	this.getCurrentScript().runFlags |= Script::tick_myplayer;
 	this.getCurrentScript().removeIfTag = "dead";
 
+	this.addCommandID("prevent emotes");
+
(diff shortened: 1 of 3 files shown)


Harrison-Miller:
We should try and get this into the next testing session.

PR #427, opened 2019-02-25
Added admin red name toggle.

Shylie:
Red names default to off.

## Status

- **READY**

## Description

Adds the ability for an admin to toggle their red name, with the red defaulting to off. The goal of this PR is to help admins punish bad players-- a lot of said players quit being bad when an admin joins.

## Steps to Test or Reproduce

- as admin, type `!toggle redname` in chat. no message should send to other clients, and your name should toggle red / not red

### DIFF
--- a/Rules/CTF/gamemode.cfg
+++ b/Rules/CTF/gamemode.cfg
@@ -47,6 +47,7 @@ scripts                                           = KAG.as;
 													EmoteMenu.as;
 													TauntMenu.as;
 													BindingsMenu.as;
+													AdminRedToggle.as;
 
 # 1 day = X minutes; / 0 no cycle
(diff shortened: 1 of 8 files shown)


eps0003:
This should disable any special name colour so devs and guards can be hidden. 

I could see it being annoying to always enable the name colour of you always want it on. Maybe this should be a property in the autoconfig (default on, but toggling persists).

Harrison-Miller:
It doesn't have to be autoconfig. If you wanted to do this purely through scripts just add a new cfg and add it the autoupdate_ignore.cfg

Shylie:
I can add dev and guard color toggle if this proves to be effective.
However, joining with the color off is one of the main points of this change.

PR #435, opened 2019-02-28
Fixed camera jumping when zooming out while holding 'e' or 'c'

eps0003:
## Status

**READY**

## Description

If you zoom out from the closest zoom while holding 'e' or 'c', the camera instantly zooms out. I'm not sure why this is a thing... please let me know if this is important for anything.

## Steps to Test or Reproduce

1. Zoom in fully
2. Hold 'e' or 'c' (or press it while zooming out)
3. Zoom out


1bardesign:
No idea, it probably needs to be tested with buttons if that's the case. I seem to remember they didn't necessarily transition perfectly when zooming in-out so maybe you could end up clicking something you didn't mean to. Could have been fixed since this workaround was added.

Harrison-Miller:
I'm going to merge it we can test it as suggested by geti.

Issue #443, opened 2019-03-04
Singleplayer TTH tutorial is broken


AsuMagic:
**Description**

When capturing a hall in the TTH tutorial "blue team" instantly wins.

**Reproduction**

1. Singleplayer -> Tutorials -> Take the halls
2. Capture a hall
3. The game will be end shortly

**Expected Behavior**

The game should not end.

Vam-Jam:
Can be closed now, since its been fixed

PR #444, opened 2019-03-04
Tutorial TTH fix

Vam-Jam:
## Status
- **READY**: Working/Ready

## Description

Prevents the game from ending if you use the tutorial TTH mode. 

## Steps to Test or Reproduce

Asu explained here

https://github.com/transhumandesign/kag-base/issues/443

### DIFF
--- a/Rules/WAR/Scripts/WAR.as
+++ b/Rules/WAR/Scripts/WAR.as
@@ -723,7 +723,17 @@ shared class WarCore : RulesCore
 	void CheckTeamWon()
 	{
 		// can't lose/modify state if the match is not running
-		if (!rules.isMatchRunning() || rules.get_bool("tutorial")) { return; }
+		if (!rules.isMatchRunning()) { return; }
+		if(rules.get_bool("tutorial"))
+		{
+			CBlob@[] halls;
+			getBlobsByName("hall", @halls);
+			if(halls[1].getTeamNum()!=0&&halls[1].getTeamNum()!=1) 
+			{
+				halls[1].server_setTeamNum(1);
+				return;
+			}
+		}
 
 		int team_won = GetTeamWon();
 		if(team_won >= 0 && team_won < teams.length)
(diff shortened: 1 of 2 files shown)


1bardesign:
Using a bool for this special behaviour is good - but the tutorial should be able to be won, as well. We probably need to just force-capture the furthest hall for red team rather than preventing winning completely.

Rules/WAR/Scripts/WAR.as
```diff
@@ -723,7 +723,7 @@ shared class WarCore : RulesCore
 	void CheckTeamWon()
 	{
 		// can't lose/modify state if the match is not running
-		if (!rules.isMatchRunning()) { return; }
+		if (!rules.isMatchRunning() || rules.get_bool("tutorial")) { return; }
```
blackjoker77777:
```suggestion
		if (!rules.isMatchRunning()) { return; }
		if(rules.get_bool("tutorial"))
		{
			CBlob@[] halls;
			getBlobsByName("hall", @halls);
			if(halls[1].getTeamNum()!=0&&halls[1].getTeamNum()!=1) 
			{
				halls[1].server_setTeamNum(1);
				return;
			}
		}
```
This should make the last hall captured by the red team, making the game only end after capturing that hall as well.

Rules/WAR/Scripts/WAR.as
```diff
@@ -724,6 +724,16 @@ shared class WarCore : RulesCore
 	{
 		// can't lose/modify state if the match is not running
 		if (!rules.isMatchRunning()) { return; }
+		if(rules.get_bool("tutorial"))
+		{
+			CBlob@[] halls;
+			getBlobsByName("hall", @halls);
+			if(halls[1].getTeamNum()!=0&&halls[1].getTeamNum()!=1)
```
AsuMagic: This is misleading because the `tutorial` property does not imply that we are currently in a **TTH** tutorial, and this relies on having exactly 2 halls in the whole map.

Vam-Jam:
Indents changed, and it works just fine, i have just tested.

PR #445, opened 2019-03-05
[Fixed] Wrong position of nobuild-zone

blackjoker77777:
## Status

- **READY**: Working/Ready

## Description

- To fix the problem shown here: https://www.youtube.com/watch?v=CXXJQ_bPQZg

- Make the "nobuild extend" zone well aligned.

## Steps to Test or Reproduce

- Before the change:
screen-19-03-05-18-03-46
screen-19-03-05-18-04-10

- After:
screen-19-03-05-18-05-11
screen-19-03-05-18-05-27

- To reproduce:

1. Join a singleplayer sandbox.

2. Try building a block inside the tent/[attempt to destroy the dirt block at the left side of the tent](https://www.youtube.com/watch?v=CXXJQ_bPQZg).

### DIFF
--- a/Entities/Structures/Common/DefaultNoBuild.as
+++ b/Entities/Structures/Common/DefaultNoBuild.as
@@ -38,8 +38,8 @@ void onTick(CBlob@ this)
 		}
 
 		this.getShape().getBoundingRect(ul, lr);
-		ul.x += 1.0f;
-		ul.y += 1.0f;
+		//ul.x += 1.0f;
+		//ul.y += 1.0f;
 
 		lr += extend;
 		this.getMap().server_AddSector(ul, lr, "no build", "", this.getNetworkID());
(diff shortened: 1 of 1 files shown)


blackjoker77777:
> Good catch. I wonder why it's there, possibly something to do with it interfering with neighbouring tiles?

The one to be credited for this would be @Biurza since he's the one who reported the issue and recorded the video. :)
> As a side note, whitespace has changed (line endings by the look of it) so I'll need to check it out locally.

Yeah Asu told me about that as well, sorry for the confusion caused by me, I didn't know Sublime Text did that.. 
I'll try to use something else for my future contributions (will try using vim later and see if I get better results, or search for a 'fix' for Sublime).
> Because all the changes are in git, feel free to actually remove lines like this rather than just commenting them out!

Roger that!

1bardesign:
I actually use Sublime Text as my editor. You should be able to set it to auto-detect the line endings from the file. It could also be happening upon pulling the repo, but I believe the repo is set to use auto-crlf on checkout to prevent this from happening? @furai might have more info there.

Issue #430, opened 2019-02-28
Various issues when a gamemode doesn't have a spectator team


eps0003:
**Description**

- !m is unable to put you in spectator but still shows the visual stuff for moderation
- 5 min auto AFK is unable to move an admin to spectator

**Reproduction**

!m issue
1. Join sandbox
2. !r a player
3. !m

AFK admin issue
1. Join sandbox
2. Don't move for 5 mins

**Expected Behavior**

You should be moved to spectator

**Screenshots**

![image](https://user-images.githubusercontent.com/30195802/53549583-a5841a80-3b88-11e9-8abe-71428c699588.png)

blackjoker77777:
Working on it.
Already fixed the first issue tho, will fix the AFK as well and make a PR later at night, as soon as I have time.

PR #447, opened 2019-03-06
No spectator team fix

blackjoker77777:
## Status

- **READY**
## Description

- This PR helps solve this issue. https://github.com/transhumandesign/kag-base/issues/430

- Basically the same old behavior of !m when it comes to CTF and TTH servers.

- When on any other gamemode, using !m switches you to the spectator team (even if it doesn't _officially exist_ for the gamemode itself).

- The visual moderation stuff is **supposed to be working** [don't know what it's supposed to look like, and didn't happen to test it properly].

- If the gamemode contains lots of teams, after using !m for the second time, you get swapped back to whatever the color of your initial team was.

- To switch back from spectator to your old team, you should type !m, since using the "change team" button doesn't work.

- The KickAFK script no longer attempts to switch admins/mods to spectator (unless it's a CTF/TTH server), instead it shows in the chat a reminder that you can either use !m on servers without spectator teams to avoid being AFK, or simply leave and come back later.

## Steps to Test or Reproduce

- **AFK kick:**

1. Join a server and remain AFK.

2. You should get this: 
_(PS: the screenshot shows me being in the spectator team after using !m)_
screen-19-03-06-15-23-49

- **About the !m feature:**

1. Join a server having a gamemode that doesn't include a spectator team. (eg: sandbox)

2. Type !m in chat, you should be switched to spectators team.

3. Type !m when you're done, you should be back to your old team.

### DIFF
--- a/Rules/CommonScripts/KickAFK.as
+++ b/Rules/CommonScripts/KickAFK.as
@@ -21,7 +21,7 @@ void onTick(CRules@ this)
 	CPlayer@ p = getLocalPlayer();
 	CControls@ controls = getControls();
 	if (p is null ||											//no player
-		controls is null ||										//no controls
+		controls is null ||										//no controls										
 		p.getTeamNum() == getRules().getSpectatorTeamNum() ||	//or spectator
 		getNet().isServer())								//or we're running the server
 	{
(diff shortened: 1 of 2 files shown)


blackjoker77777:
_sigh_
Sorry in advance.
Even after changing the default line endings to CLRF Github still considers that I changed the whole thing..
The only way to see the real comparison would be to check "Hide whitespace changes" in the Diff settings..

1bardesign:
This PR has conflicts that need to be resolved before we can consider it for review - please pull the master branch and investigate the conflicting lines (could be caused by your line endings trouble 😅)

blackjoker77777:
The conflicts it had was probably due to the outdated branch I was using (Reported.as was recently updated), tried to update it and fix the issues but Github was playing tricks on me; attempting to fix it while being half-asleep myself wasn't a smart move either.
Should be fixed in this PR hopefully. --> #460 
I'm closing this messed up PR for now, sorry again for any confusions. Please have a look on #460 instead, it should have line endings fixed and should be working as a charm as I tested it thrice.

PR #407, opened 2019-02-20
Create Portuguese_br.json

MatheusGodoi:
## Status
In Development

## Description

Portuguese - Brazilian translation for the game.
Finished all tutorials, taunts, accolades, sandbox (solo), save the princess, entire ESC menu and the entire options menu too.
Re-created the PR as I reorganized the entire file so it makes more sense on the translation order, and to clean the PR a bit.

PR #449, opened 2019-03-07
Create steam_content_br.txt

MatheusGodoi:
## Status

- **READY**

## Description

Brazilian Portuguese translation of steam_content_en. Fully translated BUT I think I got a different meaning on the line 181 ( [\*]Wizard/Witch (Light and Dark versions) / [\*]Mago/Bruxa (Versão Clara e Escura) ), will maybe need to fix later.
Also there is a direct translation of line 83 ( [*]Tino Rangatiratanga (Maori) ), but Maori is a "valid" word, so I left untouched.


1bardesign:
Merged with cherry pick in aaf84178d1803675e54213296c98c9685da1c8c8 - there were conflicts from deleting the locale file here and the steam content file in #407 so i figured it was probably better to just fish out the content! Let me know if anything seems awry!

PR #450, opened 2019-03-10
[modified] the way saws affect mines

blackjoker77777:
## Status

- **READY**
## Description

- This PR modifies the way saws affect mines in two ways:

1. Saws no longer destroy same team's mines (avoid griefing/destroying your teammates mines by mistake).
2. Saws only destroy enemy mines if fed directly to the saw (the same way it works for killing players); to avoid abusing saws to clear the field (see: someone running with a saw throwing it into bushes to neutralize enemy mines).

- **How it used to work:**
[Having your mines destroyed by mistake.]
[Having an enemy using the saw as an anti-mine shield.]
old_1
[Clearing mines on the field in a matter of seconds.]
old_2

- **How it should be working after this PR:**
[Saws no longer waste your hard-earned mines.]
new_1
[Enemies can't use saws as shields.]
[In order to eliminate an incoming enemy mine, you should position/throw the saw correctly.]
[Enemies falling into a saw loose the mines they have hidden in their inventories.]
new_2

## Steps to Test or Reproduce

1. Join any gamemode.
2. Spawn/buy a mine and a saw, being in the same team.
3. throw the mine into/towards the saw.
4. Try throwing a mine of a different team into/towards the saw.

### DIFF
--- a/Entities/Industry/Saw/Saw.as
+++ b/Entities/Industry/Saw/Saw.as
@@ -159,7 +159,8 @@ bool canSaw(CBlob@ this, CBlob@ blob)
 	    name == "wooden_door" ||
 	    name == "mat_wood" ||
 	    name == "tree_bushy" ||
-	    name == "tree_pine")
+	    name == "tree_pine"||
+	    name == "mine")
 	{
 		return false;
 	}
--- a/Entities/Industry/Saw/Saw.as
+++ b/Entities/Industry/Saw/Saw.as
@@ -160,13 +160,13 @@ bool canSaw(CBlob@ this, CBlob@ blob)
 	    name == "mat_wood" ||
 	    name == "tree_bushy" ||
 	    name == "tree_pine"||
-	    name == "mine")
+	    (name == "mine"&&blob.getTeamNum()==this.getTeamNum()))
 	{
 		return false;
 	}
 
 	//flesh blobs have to be fed into the saw part
-	if (blob.hasTag("flesh"))
+	if (blob.hasTag("flesh")||(name=="mine"))
 	{
 		Vec2f pos = this.getPosition();
 		Vec2f bpos = blob.getPosition();
--- a/Entities/Industry/Saw/Saw.as
+++ b/Entities/Industry/Saw/Saw.as
@@ -159,14 +159,14 @@ bool canSaw(CBlob@ this, CBlob@ blob)
 	    name == "wooden_door" ||
 	    name == "mat_wood" ||
 	    name == "tree_bushy" ||
-	    name == "tree_pine"||
-	    (name == "mine"&&blob.getTeamNum()==this.getTeamNum()))
+	    name == "tree_pine" ||
+	    (name == "mine" && blob.getTeamNum() == this.getTeamNum()))
 	{
 		return false;
 	}
 
 	//flesh blobs have to be fed into the saw part
-	if (blob.hasTag("flesh")||(name=="mine"))
+	if (blob.hasTag("flesh") || (name=="mine"))
 	{
 		Vec2f pos = this.getPosition();
 		Vec2f bpos = blob.getPosition();
--- a/Entities/Items/Explosives/Mine/Mine.as
+++ b/Entities/Items/Explosives/Mine/Mine.as
@@ -29,7 +29,7 @@ void onInit(CBlob@ this)
 	this.set_u8("custom_hitter", Hitters::mine);
 
 	this.Tag("ignore fall");
-
+	this.Tag("ignore_saw");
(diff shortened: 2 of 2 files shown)


Mazey:
Why though? I'd consider it a feature; throw things in the saw and either something happens (logs into wood, bombs explode, etc) or the item gets destroyed.

blackjoker77777:
I find it unfair to have someone wasting their money on a mine, just for it to be discarded after falling into a saw; besides builders/archers/knights could just bring a saw and keep throwing them into bushes destroying any residing mines.
I mean mines are supposed to be used tactically in a smart way, and they should be disabled by doing some effort (builder destroying it/ archer sniping it/ knight.. jabbing it?) not just throwing something over it and voila?
That is my opinion at least. 
@AsuMagic  mentioned as well that mines weren't intended to be destroyed that way by saws; and recommended this PR.

Mazey:
Throwing a saw on a mine shouldn't destroy it, only throwing a mine into a saw, in my opinion (similar to how characters are affected by saws). Perhaps it could be like that?

eps0003:
I agree with @Mazey. Saws collide with and explode all explosives, so making mines an exception would be unintuitive. If anything, prevent mines from colliding with the same team's saw.

blackjoker77777:
@Mazey @eps0003 Just updated it, what do you think about having it working this way?

Harrison-Miller:
It feels like maybe the mine should just explode on the saw and destroy it. So it can still be used smartly to destroy mines, but costs you.

1bardesign:
Accepted by `#democracy` 128 :+1: to 12 :-1: - we'll go with it as-is I think and tweak as needed, consensus is that it's a good change though.

eps0003:
This feature was poorly implemented.

Firstly, enemy mines that fall into the saw make the saw bloody.

Secondly, why are mines the only item in the game that has this team-specific behaviour with saws? What about other explosives like kegs and bombs? It's extremely irritating when your keg rolls into a team saw and get destroyed. Then what about items like drills? Should they also act the same way? With all this inconsistency, it becomes hard to know which items are immediately sawed and which need to be thrown into the top.

So these questions need to be answered:
- What determines which types of items need to fall into the saw or not?
- What determines which types of items don't get sawed by team saws?
- What determines which types of items completely ignore saws?

blackjoker77777:
> This feature was poorly implemented.

... Thanks:question: 
> Firstly, enemy mines that fall into the saw make the saw bloody.

I probably missed that, it isn't hard to have that fixed, and I'll do it as soon as I have time. Thanks for pointing that out!
> Secondly, why are mines the only item in the game that has this team-specific behaviour with saws? What about other explosives like kegs and bombs? It's extremely irritating when your keg rolls into a team saw and get destroyed. Then what about items like drills? Should they also act the same way? With all this inconsistency, it becomes hard to know which items are immediately sawed and which need to be thrown into the top.

Ehh, simply put: because mines are not any regular item? and because they're different from other explosives? Here's what I mean:
* bombs are easy to acquire, and easier to waste, you don't hide your bombs somewhere laying on the ground; saws make them explode instantly (they're basically activated while being mid-air) and players seem to like that..
* Kegs are quite hard to buy, you're supposed to take care of them correctly, you don't hide them behind bushes, and anyone in their right state of mind wouldn't simply roll them towards team saws? if you're running with a saw in your hand and find some enemy keg laying around, your first intention would be to steal it and use it for your own advantage, instead of destroying it; right?
* mines on the other hand **should** be rolled everywhere including behind a saw, they **must** be hidden because they're a trap; and obviously they're no bomb to be neutralized mid-air by a saw, because simply they're not yet activated at that time..

To be brutally honest, I agree with you about it not being that consistent.
I was thinking about making **every item** [logs excluded, they should keep their old behavior!] behave the same way mines do in this PR, so you can get rid of *enemy* items if they fall **into** *your* saw (more realistic than what it is right now, and would make it more skill-based), and avoid having your team items discarded by mistake thanks to your team saws (saws were never intended to be the trash can, in which you discard whatever you have; I still can't accept them destroying your team items), if you bought a drill or a lantern or anything and you no longer need them.. just build a storage? or throw them somewhere!

People wouldn't like that change, because it would fix/remove some of the so-called "features".. (*stares silently to sawjumps*)

PS: I obviously wouldn't want the 'players falling into saws' mechanism to be changed.
> So these questions need to be answered:
> 
>     * What determines which types of items need to fall into the saw or not?
> 
>     * What determines which types of items don't get sawed by team saws?
> 
>     * What determines which types of items completely ignore saws?

That, my friend, needs to be asked directly to the community.. we need more feedback regarding this matter and that is a fact.

PR #453, opened 2019-03-12
Extra taunt settings

eps0003:
## Status

**READY**

## Description

These settings might be needed in case taunts are spammed in f2p or if people change their opinions of taunts.
- Click to select category bool
- Show taunts in chat bool (depends on `cl_chatbubbles` setting)


Harrison-Miller:
Could you describe more specifically what these changes do?

eps0003:
Click to select category bool makes it so you need to click on a category to select it instead of just dragging your mouse to it. 

Show taunts in chat bool is for whether taunts show in chat or not. The taunt bubbles will still display only if `cl_chatbubbles` is enabled. 

This PR doesn’t change the current behaviour of taunts. It only provides extra options in case they are needed in the future.

PR #454, opened 2019-03-12
Fireplaces in CTF/TTH

eps0003:
## Status

**READY**

## Description

This PR is the stuff from #378 that need a #democracy vote. The other stuff were already merged in #425

- Fireplaces are log-powered, except those spawned in with the map (or spawned in using commands when your team is -1)
  - The burn rate may need to be tweaked. I forsee it being a bit too quick considering how long it takes  to get logs, unless you have a teammate helping you out while you shoot
- Added fireplace to builder shop (CTF) and workbench (TTH)
  - Currently costs 100 wood and 100 stone but I didn't put much thought into it. This needs to be voted on in #democracy
- Unlit fireplaces can be picked up and put in your inventory (1x1 slot size)
  - I'm a bit unsure whether being able to put fireplaces in your inventory is really needed. Maybe let people vote on it in #democracy
- Added missing shop items to workbench and reorganised them to match builder shop layout

### DIFF
--- a/Entities/Common/Includes/Costs.as
+++ b/Entities/Common/Includes/Costs.as
@@ -24,7 +24,7 @@ namespace CTFCosts
 	//BuilderShop.as
 	s32 lantern_wood, bucket_wood, filled_bucket, sponge, boulder_stone,
 		trampoline_wood, saw_wood, saw_stone, drill_stone, drill,
-		crate_wood, crate;
+		crate_wood, crate, fireplace_wood, fireplace_stone;
 
 	//BoatShop.as
 	s32 dinghy, dinghy_wood, longboat, longboat_wood, warboat;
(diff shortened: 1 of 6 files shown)


furai:
@1bardesign I'd love to get your input on this.

1bardesign:
At this point I'm unsure.

I personally think it wouldn't change much in terms of balance and would be interesting new content. Reception hasn't been positive when talking with the community about it. It could be worth splitting CTF into a version for "purists" and version with more stuff like this and #393 - because as it is I can see half the community voting YES YES YES just cause it's new content, and half voting NO NO NO just cause it's new content 😅

AsuMagic:
I am a bit late here but that has gone stale. What do we do with it? I'm clearly not a fan of splitting CTF into two gamemodes.
The differences would either be too small, in which case it would be an unnecessary community split, or either would be significant enough for it to simply be a mod that does not need to be official.

I don't think it adds that much gameplay value, and poses balance issues. This means 0 coin fire arrows, which would be far too effective against enemy structures (particularly in the battlefield) and enemies in general. Drop one in an archer nest and you're ready to pwn.

Harrison-Miller:
Closing pull request because I don't think the community could agree on adding this. Also splitting CTF gamemodes isn't a good idea, the player base is already small enough.

PR #455, opened 2019-03-12
Dead bodies disable saws of the same team

eps0003:
## Status

**READY**

## Description

- Dead bodies disable saws of the same team
  - Gives knights and archers a way of 'disarming' saw traps without needing a builder to block it off / destroy it
  - Mines have similar behaviour which works really well
  - This gives bodies more uses other than for taunting with ;)
  - If alive teammates disable saws, shouldn't their dead bodies also do the same?
- Saws no longer destroy other saws
  - Why was this even a thing? Most of the time you accidentally destroy your own saws and get called out for griefing

## Steps to Test or Reproduce

- Throw a dead body into top of a saw of the same team
- Throw a saw into another saw

### DIFF
--- a/Entities/Industry/Saw/Saw.as
+++ b/Entities/Industry/Saw/Saw.as
@@ -120,8 +120,11 @@ void Blend(CBlob@ this, CBlob@ tobeblended)
 
 	tobeblended.Tag("sawed");
 
-	// on saw player - disable the saw
-	if (tobeblended.getPlayer() !is null && tobeblended.getTeamNum() == this.getTeamNum())
+	// on saw player or dead body - disable the saw
+	if (
+		(tobeblended.getPlayer() !is null || //player
+		(tobeblended.hasTag("flesh") && tobeblended.hasTag("flesh"))) && //dead body
+		tobeblended.getTeamNum() == this.getTeamNum()) //same team as saw
 	{
 		CBitStream params;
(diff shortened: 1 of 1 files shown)


ParaLogia:
The first half of this PR is actually pretty significant balance change. After some testing, it's trivially easy to drop a corpse into a saw trap (i.e. a saw covered by trap blocks). Currently, saw traps are pretty much the only thing that make trap blocks useful. If you allow knights to trivially disable them, then trap blocks will become even more useless, and purely detrimental to build. Even tunnel traps will only be good for one enemy, if they bring a corpse through.

The second half is fine, but this PR should not be merged without more input/democracy.

Blubahub:
This PR would not make it so [living] foes ground up in the saw won't clog the saw up with their body, right? Because if so, then saw traps would need to be constantly tended (restored after each kill).

eps0003:
No @Blubahub, if you fall into your team’s saw, the saw will deactivate. This PR adds the same behaviour to your dead body. Dead enemy bodies won’t disable the saw.

ewetcr:
The function of tks and the button disabling sawmills was to avoid massive teamkilling, not to improve the gameplay if I remember correctly.

1bardesign:
Accepted by `#democracy` 90 :+1: to 46 :-1: - still likely to need some testing but seems folks are largely in favour.

PR #460, opened 2019-03-13
[fixed] issues related to not having a spectator team in a gamemode.

blackjoker77777:
## Status

- **READY**
## Description

- This PR helps solve this issue. https://github.com/transhumandesign/kag-base/issues/430

- ~~Basically the same old behavior of !m when it comes to CTF and TTH servers.~~ (Same behavior for all gamemodes).

- When on any other gamemode, using !m switches you to the spectator team (even if it doesn't _officially exist_ for the gamemode itself).

- The visual moderation stuff works.

- Typing _!m_ again gets you back to your previous non-spectator team.

- To switch back from spectator to your old team, you should type !m, ~~since using the "change team" button doesn't work.~~ (now it does), or just use the normal "change team" button.

- ~~The KickAFK script no longer attempts to switch admins/mods to spectator (unless it's a CTF/TTH server), instead it shows in the chat a reminder that you can either use !m on servers without spectator teams to avoid being AFK, or simply leave and come back later.~~

- Now the kickAFK script does force-swap admins to spectator team.

## Steps to Test or Reproduce

- **AFK kick:**

1. Join a server, in-which you're an admin (or just /rcon /login <pass>; /rcon /reloadsecurity).

2. Remain AFK for enough time.

3. Get swapped to spectator with a message in chat stating that you can get back to your previous team using _!m_. 
[_The screenshot shows it working on a CTF server, but it would do the same with any other gamemode._]
screen-19-03-13-15-18-09

- **About the !m feature:**

1. Join a server having a gamemode that doesn't include a spectator team. (eg: sandbox)

2. Type !m in chat, you should be switched to spectators team.

3. Type !m when you're done, you should be back to your old team.


Rules/CommonScripts/KickAFK.as
```diff
@@ -75,7 +75,17 @@ void onTick(CRules@ this)
...
-			p.client_ChangeTeam(this.getSpectatorTeamNum());
+			if(getRules().gamemode_name=="CTF"||getRules().gamemode_name=="TTH")
+			{
+				p.client_ChangeTeam(this.getSpectatorTeamNum());
+			}
+			else
+			{
+				client_AddToChat("You have been AFK way too long, this server may not have a spectator team to swap you to!",SColor(255,0,0,0));
+				client_AddToChat("[!] Please either:",SColor(255,0,0,0));
+				client_AddToChat("-Type !m in chat while being AFK, re-type it when you're back.",SColor(255,0,0,0));
```
eps0003: I wouldn't call this a 'fix'. You're just telling the player to use a totally unrelated feature to fix this issue. The message it posts is useless since the player won't even see the message until they're back from being AFK. The message also takes up five chat lines for something that only needs to take up two at most.

Rules/CommonScripts/Report.as
```diff
@@ -3,21 +3,35 @@
...
+	this.addCommandID("mod_team");
 }
 
 bool onClientProcessChat(CRules@ this, const string& in text_in, string& out text_out, CPlayer@ player)
 {
 	if((text_in == "!moderate" || text_in == "!m") && player.isMod())
 	{
-		moderate(this, player);
-
+		if(this.gamemode_name=="CTF"||this.gamemode_name=="TTH")
```
eps0003: I don't think you should be hardcoding specific game modes. Maybe there's a way to check if the game mode has a spectator team.

blackjoker77777:
- Fixed code formatting. 
- Made the _!m_ new behavior the same for all gamemodes, so now no gamemodes are hard-coded.
- The KickAFK old issue is now **fixed**, as in, it now force-swaps admin to the spectator team instead of blaming them for not doing so.
- Added a new script 'AdminLogic.as', which would be responsible for swapping admins to spec team (had to move some parts related to reporting, to it as well).

Thank you @eps0003 for pointing out the above flaws.

Rules/CommonScripts/AdminLogic.as
```diff
@@ -0,0 +1,119 @@
+//AdminLogic.as
+/*
+	-Reporting players.
+	-Forcing spectator team for admins.
+*/
+
+u8 nonSpecTeam = 0; //Sandbox' default team.
+const SColor reportMessageColor(255, 255, 0, 0);
+
+void onCommand(CRules@ this, u8 cmd, CBitStream @params)
```
Harrison-Miller: moving this hook out of report.as and into AdminLogic.as is bad. Now any file that includes AdminLogic.as will have this hook so the logic will get executed multiple times.

It would be better if you left this code in Report.as and fixed it there instead.

Rules/CommonScripts/AdminLogic.as
```diff
@@ -0,0 +1,119 @@
...
+	    CPlayer@ player = getPlayerByUsername(p_name);
+	    if(on_spec) //player is alread in spec team.
+	    {
+		    player.server_setTeamNum(previousTeam);
+		    this.server_PlayerDie(player); //force the player to join his old team.
+	    }
+	    else
+	    {
+		    CBlob@ corpse = player.getBlob();
+		    player.server_setTeamNum(200); //get to new spec team.
```
Harrison-Miller: this.getSpectatorTeamNum()

same for other instance of using 200

1bardesign:
If you could address verra's concerns in the recent review that would be great - particularly the hook changing files! If you need help reconciling where the hook could go, would be better to bring discussion to #development on discord rather than let this go stale

Rules/CommonScripts/AdminLogic.as
blackjoker77777: Just changed that.

Harrison-Miller:
Thanks for fixing most of the stuff with the hooks. Could you go through and convert everything to tabs. It looks like there's a pretty big mix of spaces and tabs.

blackjoker77777:
Sorry for the spaces issue, apparently when committing changes directly from the website.. things get messed up. :disappointed: 
Just converted them to tabs, sorry for any inconvenience.
Let's hope everything works as expected this time. :thinking:

PR #469, opened 2019-03-17
Added epsilon Patreon head

eps0003:
## Status

**READY**

## Description

Added my Patreon head


Rules/CommonScripts/accolade_data.cfg
```diff
@@ -563,6 +563,7 @@ Elowan =
 	# participation 2 - EU 1v1 knight tournament no. 1
 
 epsilon =
```
furai: Can you move that section up to the Patrons head section?

furai:
Ok, it seems that we haven't avoided the merge conflict.

PR #470, opened 2019-03-17
Cinematic spectator camera

eps0003:
## Status

**IN DEVELOPMENT**

## Description

This PR greatly improves the viewing experience of spectators by adding a cinematic camera mode. You can enable the cinematic camera by right-clicking while in spectator or while dead/respawning. Moving the camera using mouse/WASD keys or clicking a player to follow them disables the cinematic camera. The cinematic camera is enabled by default.

The cinematic camera works by positioning the camera at the mean position of all players. The zoom level is the max distance between the two furthest players on the horizontal and vertical axes (clamped between min and max default zoom). When a round starts, round ends, during build time or when there are not enough players for a match, the camera zooms out to fit the entire map (or as much as it can zoom out to fit).

The cinematic camera also focuses on important events. The events are listed below, ordered by importance:

1. Missing flags
2. Capturing halls
3. Builders near enemy flag
4. Players near enemy tent
5. Lit kegs
6. Capturing vehicles with spawns (ballistae, war boats)

If there are multiple important events, the events with higher importance are prioritised. Kegs closer to exploding and vehicles closer to being captured are more important than other kegs/vehicles.

This PR also improves camera panning while the 'snap camera to pixel' setting is enabled.

## Steps to Test or Reproduce

1. Join any game mode with spectator team (event focus works best on CTF/TTH)
2. Join spectator
3. Right click if the camera isn't already cinematic

This has been thoroughly tested on Gather and in TDM tournaments.


Harrison-Miller:
Do you have any cool gifs or videos displaying this working?

eps0003:
I've only got outdated clips from TDM tournaments which only show the camera pan/zoom. I might make a few gifs soon. https://www.twitch.tv/eps0003/clips?filter=clips&range=all

Rules/CommonScripts/Spectator.as
```diff
@@ -15,101 +27,188 @@ CPlayer@ targetPlayer()
...
+			//calculate mean position of all players
+			posTarget = Vec2f(0, 0);
+			for (uint i = 0; i < players.length; i++)
+			{
+				CBlob@ blob = players[i];
+
+				posTarget += blob.getInterpolatedPosition();
+
+				//look for largest distance between two players
+				for (uint j = 0; j < players.length; j++)
```
Harrison-Miller: start at i+1

Rules/CommonScripts/Spectator.as
```diff
@@ -15,101 +27,188 @@ CPlayer@ targetPlayer()
...
+				{
+					CBlob@ blob2 = players[j];
+					u32 distX = Maths::Abs(blob.getPosition().x - blob2.getPosition().x);
+					u32 distY = Maths::Abs(blob.getPosition().y - blob2.getPosition().y);
+					maxDistX = Maths::Max(distX, maxDistX);
+					maxDistY = Maths::Max(distY, maxDistY);
```
Harrison-Miller: maybe consider an early out if max distance reached, I'm interested to see the performance of this running every tick on a full server.

Rules/CommonScripts/Spectator.as
```diff
@@ -15,101 +27,188 @@ CPlayer@ targetPlayer()
...
+				f32 zoomW = calculateZoomLevelW(maxDistX * 1.8f);
+				f32 zoomH = calculateZoomLevelH(maxDistY * 1.8f);
+				zoomTarget = Maths::Min(zoomW, zoomH);
+			}
+
+			zoomTarget = Maths::Clamp(zoomTarget, 0.75f, 1.5f);
+		}
+
+		//adjust camera pos and zoom
+		camera.targetDistance += (zoomTarget - camera.targetDistance) / CINEMATIC_ZOOM_EASE * getRenderApproximateCorrectionFactor();
```
Harrison-Miller: if it turns out that this is really slow on large servers, you could simply calculate the focus/targetPos every few ticks, but update the easing every tick.

Rules/CommonScripts/PlayerCamera.as
```diff
@@ -205,3 +211,32 @@ void onRender(CRules@ this)
...
+
+	CControls@ controls = getControls();
+	if (controls !is null)
+	{
+		//enable cinematic camera
+		const int diffTime = deathTime - getGameTime();
+		if (controls.isKeyJustPressed(KEY_RBUTTON) && (spectatorTeam || diffTime <= 0))
+		{
+			SetTargetPlayer(null);
+			timeToCinematic = 0.0f;
```
Harrison-Miller: why are you automatically sending the player into cinematic mode everywhere in this file. At the very least there is a bug with the death lock which zooms in. When you set timeToCinematic = 0.0f; deathlock won't zoom in on the locked position because of the condition on line 128 of Spectator.as.

Rules/CommonScripts/Spectator.as
Harrison-Miller: and the end condition above that should be player.lenght-1

Rules/CommonScripts/PlayerCamera.as
Harrison-Miller: ignore everything in this comment, rereading I was wrong about the bug. I do think however what you talked about was it not auto going to cinematic on some of those conditions but instead remembering.

eps0003: I don't understand what you mean. Death lock should still work like normal. Look at `PlayerCamera.as` from around line 135.

Rules/CommonScripts/Spectator.as
```diff
@@ -1,7 +1,19 @@
+#include "CinematicCommon.as";
+
 #define CLIENT_ONLY
 
-f32 zoomTarget = 1.0f;
-float timeToScroll = 0.0f;
+const u8 CINEMATIC_ZOOM_EASE = 14.0f;					//amount of ease when zooming while cinematic
+const u8 CINEMATIC_PAN_X_EASE = 12.0f;					//amount of ease along the y-axis while cinematic
+const u8 CINEMATIC_PAN_Y_EASE = 12.0f;					//amount of ease along the x-axis while cinematic
+const u32 CINEMATIC_TIME = 10.0f * getTicksASecond();	//time until camera automatically becomes cinematic in seconds
```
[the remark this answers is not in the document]
eps0003: That's good to know

Rules/CommonScripts/Spectator.as
eps0003: I can't do `player.length - 1` because the same loop is used for calculating the average position of all players. This would skip over the last player and make the camera sit off centre.

Harrison-Miller: It's not an important enough improvement to worry too much about it just leave` i < player.length`. You can still do j = i+1. (If you really wanted you can do `posTarget += players.last().getInterpolatedPosition()` before calculating the average.

eps0003:
I've made the changes @Harrison-Miller suggested and made it remember whether the cinematic camera is enabled or disabled (suggested by @furai). This should be ready after a final review.

1bardesign:
Will wait for verra to re-review once he's back but thanks for making the requested changes quickly!

The performance concerns mean that it'd be great to have a big testing session (or at least a lot of bots added), but I guess it's at least only clientside performance issues.

eps0003:
This is really laggy on a CTF server with only 12 players. I'll implement the performance improvements soon.

eps0003:
Even with the performance improvements suggested by @Harrison-Miller, this is still extremely laggy. Increasing the update interval will make the camera stop and start as it pans/zooms. I don't really know what else could be done to improve performance. 😢

1bardesign:
I feel like updating/selecting points of interest iteratively (ie over several frames, process 1 blob at a time and then sort generated points), and then just panning there as a separate process would probably be the way forward. Likely to require an overhaul of the code though.

eps0003:
I'm glad I took a break from this because I now see so many issues with how I implemented it. Here's what I have changed to improve the performance of this:
- Moved most logic out of onRender
- Filtered out unimportant blobs so sorting is much quicker
- Implemented quicksort to replace the shitty bubble sort algorithm
- Replaced the code to position and scale the camera with more efficient code
  - I used to compare every player blob with every other player blob
    `n * n/2` steps
  - Now I quicksort twice to get the furthest distance between two player blobs on each axis and then loop through all player blobs to calculate the mean position
    `2*log(n) + n` steps

This should now work with good framerate on populated servers
<details>
<summary>Graph</summary>

image
</details>

eps0003:
Optimised it even further. Now it takes `n` steps to calculate the position and scale of the camera.

Rules/CommonScripts/PlayerCamera.as
```diff
@@ -197,11 +203,94 @@ void onRender(CRules@ this)
...
+void onTick(CRules@ this)
+{
+	if (isCinematic())
+	{
+		Vec2f mapDim = getMap().getMapDimensions();
+
+		if (getGameTime() == 1)
+		{
+			//initially position camera to view entire map
+			CCamera@ camera = getCamera();
```
Harrison-Miller: Replace with the ViewEntireMap function in CinematicCommon.as

Rules/CommonScripts/PlayerCamera.as
```diff
@@ -197,11 +203,94 @@ void onRender(CRules@ this)
...
+			pos = mapDim / 2.0f;
+			camera.setPosition(pos);
+			float zoomW = calculateZoomLevelW(mapDim.x);
+			float zoomH = calculateZoomLevelH(mapDim.y);
+			camera.targetDistance = Maths::Min(zoomW, zoomH);
+		}
+
+		if (this.isMatchRunning() && !this.isWarmup() && !this.isGameOver())
```
Harrison-Miller: only need this.isMatchRunning

Rules/CommonScripts/PlayerCamera.as
Harrison-Miller: Replace with ViewEntireMap function.

Rules/CommonScripts/PlayerCamera.as
Harrison-Miller: Only need this.isMatchRunning()

Rules/CommonScripts/PlayerCamera.as
```diff
@@ -197,11 +203,94 @@ void onRender(CRules@ this)
...
+					Vec2f minPos = mapDim;
+					Vec2f maxPos = Vec2f_zero;
+
+					for (uint i = 0; i < playerBlobs.length; i++)
+					{
+						CBlob@ blob = playerBlobs[i];
+						Vec2f pos = blob.getPosition();
+
+						//max distance along each axis
+						if (pos.x > maxPos.x) maxPos.x = pos.x;
```
Harrison-Miller: can use min/max or clamp functions.

Rules/CommonScripts/CinematicCommon.as
```diff
@@ -0,0 +1,303 @@
...
+{
+	Vec2f mapDim = getMap().getMapDimensions();
+	posTarget = mapDim / 2.0f;
+	float zoomW = calculateZoomLevelW(mapDim.x);
+	float zoomH = calculateZoomLevelH(mapDim.y);
+	zoomTarget = Maths::Min(zoomW, zoomH);
+	zoomTarget = Maths::Clamp(zoomTarget, 0.5f, 2.0f);
+}
+
+ConfigFile@ openCinematicConfig()
```
Harrison-Miller: Once you read the value from the file save it to a blob/rules or global in the script and don't try to read from the file anymore.

Rules/CommonScripts/CinematicCommon.as
```diff
@@ -0,0 +1,303 @@
...
+	if (!cfg.loadFile("../Cache/cinematic_prefs.cfg") || !cfg.exists("cinematic_enabled"))
+	{
+		cfg.add_bool("cinematic_enabled", true);
+		cfg.saveFile("cinematic_prefs.cfg");
+	}
+
+	return cfg;
+}
+
+void setCinematicEnabled(bool enabled)
```
Harrison-Miller: When we read, prefer reading from the value set above if we've already loaded the config.

Rules/CommonScripts/PlayerCamera.as
```diff
@@ -197,11 +203,94 @@ void onRender(CRules@ this)
...
 		lr.y += wave;
 		//draw
 		GUI::DrawButtonPressed(ul - Vec2f(10, 10), lr + Vec2f(10, 10));
 		GUI::DrawText(text, ul, SColor(0xffffffff));
 	}
 }
+
+void onTick(CRules@ this)
+{
+	if (isCinematic())
```
Harrison-Miller: this is opening and reading a file every tick, we shouldn't be doing that. We should only read the file once. And only write to it on changes.

Rules/CommonScripts/PlayerCamera.as
```diff
@@ -114,14 +116,12 @@ void onPlayerDie(CRules@ this, CPlayer@ victim, CPlayer@ attacker, u8 customData
 			else
 			{
 				camera.setTarget(null);
+				setCinematicEnabled(isCinematicEnabled());
```
Harrison-Miller: Every line like this seems unnecessary if you've stored the variable correctly.

Rules/CommonScripts/CinematicCommon.as
```diff
@@ -0,0 +1,303 @@
...
+float calculateZoomLevelW(uint width)
+{
+	Driver@ driver = getDriver();
+	float ratio = float(driver.getScreenHeight()) / float(driver.getScreenWidth());
+	return calculateZoomLevelH(width * ratio);
+}
+
+void calculateZoomTarget(float distX, float distH)
+{
+	float zoomW = calculateZoomLevelW(distX * 1.8f);
```
Harrison-Miller: calculateZoomLevelW calls calculateZoomLevelH. Can we just combine this into a single function that returns a Vec2 instead of double calling calculateZoomLevelH?

eps0003:
It should be ready now @Harrison-Miller

PR #477, opened 2019-03-21
Admins always see other admins' name colour and moderation accolade

### LINKED ISSUE #475: Admins should be able to see other admins name colors regardless of name toggle

Currently when the name color is turned off nobody including other admins can see that a player is an admin.

**Proposed Change**

Instead admin, moderators, and devs should be able to see everyone's name color regardless of toggle setting. If you are not familiar with the names of other admins this would allow you to identify them.

Doing this would help increase accountability, and allow you to pass  chat to another admin about context, give tips about potential offenders etc.

**Additional context**

However you should still be able to see your own name toggle on and off, so that you know whether you  have it set on or off.

eps0003:
## Status

**READY**

## Description

- Admins always see other admins' name colour and moderation accolade
- You are still able to see whether your name colour is hidden or not

Resolves #475

## Steps to Test or Reproduce

1. Be admin on a server with other admins
2. Ask other admins to toggle their name colour
   - You should always be able to see their name colour and moderation accolade
3. Toggle your own name colour
   - Should work like usual - only see moderation accolade when name colour is visible

### DIFF
--- a/Rules/CommonScripts/ScoreboardCommon.as
+++ b/Rules/CommonScripts/ScoreboardCommon.as
@@ -8,7 +8,8 @@ f32 getKDR(CPlayer@ p)
 SColor getNameColour(CPlayer@ p)
 {
     SColor c;
-    bool showColor = coloredNameEnabled(getRules(), p);
+	CPlayer@ localplayer = getLocalPlayer();
+    bool showColor = (p !is localplayer && isSpecial(localplayer)) || coloredNameEnabled(getRules(), p);
 
     if (p.isDev() && showColor) {
         c = SColor(0xffb400ff); //dev
--- a/Rules/CommonScripts/ScoreboardCommon.as
+++ b/Rules/CommonScripts/ScoreboardCommon.as
@@ -8,7 +8,7 @@ f32 getKDR(CPlayer@ p)
 SColor getNameColour(CPlayer@ p)
 {
     SColor c;
-	CPlayer@ localplayer = getLocalPlayer();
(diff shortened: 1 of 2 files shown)


Rules/CommonScripts/ScoreboardRender.as
```diff
@@ -371,7 +371,11 @@ float drawScoreboard(CPlayer@[] players, Vec2f topleft, CTeam@ team, Vec2f emble
 					1 : 0),             5,     0,         0,
 				(acc.map_contributor ?
 					1 : 0),             6,     0,         0,
-				(acc.moderation_contributor && (!isSpecial(p) || coloredNameEnabled(getRules(), p)) ? //ensure accolade is visible for past admins
+				(acc.moderation_contributor && (
```
Harrison-Miller: won't nitpick you to change this right now. But in the future for readability it would be better to split this out into multiple booleans then combine them in the final check.

Rules/CommonScripts/ScoreboardCommon.as
```diff
@@ -8,7 +8,8 @@ f32 getKDR(CPlayer@ p)
 SColor getNameColour(CPlayer@ p)
 {
     SColor c;
-    bool showColor = coloredNameEnabled(getRules(), p);
+    CPlayer@ localplayer = getLocalPlayer();
+    bool showColor = (p !is localplayer && isSpecial(localplayer)) || coloredNameEnabled(getRules(), p);
```
Harrison-Miller: same here.

PR #482, opened 2019-03-24
[added] Snoweyz Patreon head

Mazey:
## Status

- **READY**


furai:
@Harrison-Miller was the split into separate files tested enough? Or should we add the head the old way?

PR #515, opened 2019-04-21
Hold S to pass through team trampolines

eps0003:
## Status

**READY**

## Description

Advantage:
- Better movement around team trampolines

Disadvantage:
- Harder for enemies to predict your movement if you can choose to jump or not

Leaving this up to the community to decide if it's a decent addition

## Steps to Test or Reproduce

Hold S while about to jump on a team trampoline


ferrezinhre:
"Better movement around trampolines" can I ask for solid proofs that trampolines disturbed flow of movement, but let's have one requirement, do not count trampolines that was **badly placed by players**, like near doors, in shops, inside tunnels or towers.

On side note, vertically positioned trampolines will lose their magic
verttramp
ps. it bit looks like change for sake of change, but maybe it will have some impact on game who knows, and everybody loves crouching.

eps0003:
> "Better movement around trampolines" can I ask for solid proofs that trampolines disturbed flow of movement, but let's have one requirement, do not count trampolines that was **badly placed by players**, like near doors, in shops, inside tunnels or towers.

The whole point of this is to improve movement around badly placed tramps, especially ones in base

> On side note, vertically positioned trampolines will lose their magic

Yeah I'm aware of that and have noted that its a disadvantage of this change

eps0003:
Changed it so you can only pass through team trampolines since this would prevent you from defensively throwing a trampoline at an enemy.

Another solution to this issue would've been to only allow anyone to pass through stationary trampolines. However, after some testing, it was difficult to know whether the trampoline is slow enough to pass through it.

Omen267901:
I don't have any issue with a change like this if it's for team trampolines only. It would definitely help trampolines be more useful for movement within bases without causing problems getting to other areas.

That being said, is there any planned implementation for indicating to players that you can press S to not bounce? If not I could see newer players getting hung up on that.

MrPotatobird:
I think it would maybe be better if, rather than having to hold S to bypass your team's trampolines, you only bounced on them when holding W. I think that will feel more natural and intuitive, because you're only really holding W when you want to go up, which is also when you wanna bounce on a trampoline. If you're required to hold S then you'll still sometimes get bounced around non-consensually.

eps0003:
@MrPotatobird after thinking about it a bit, I actually don't think it would be good to do what you are suggesting. Enemies would no longer be able to defensively use trampolines since players dodge them by default. Also, other things like crouch dodging and (possibly) the ability to pass through ladders require you to hold S to dodge, so your suggestion would go against that.

MrPotatobird:
Yeah, you probably shouldn't be able to choose to pass through enemy trampolines ever. I was just talking about your team's trampolines. But I don't think trampolines are converted to the other team when grabbed so maybe they should do that.

As for the "consistency" with crouching, aren't crouch dodging and the *potential* ladder thing the only other examples? Honestly, crouch dodging is a pretty obscure mechanic, and it's not intuitive at all. No new player is ever going to figure that out until someone else explains it to them. So basing new mechanics on that one (but not quite, because they take place in the air so actually have nothing to do with crouching) just in the name of consistency doesn't sit right with me

eps0003:
Instead of holding "S" to pass through team trampolines, you now need to hold "W" to jump. The reasoning behind this is explained in @MrPotatobird's comments.

blackjoker77777:
It totally feels much better this way. :+1:

ferrezinhre:
This will break trampolines completly, vertical trampolines will not work, you cant throw team mines from trampolines, nor catapult or ballistas, team bombs dosen't work too, and team kegs too, maybe start testing your ideas. Anyway here is proper code for this:
if (!blob.isKeyPressed(key_up) && this.getTeamNum() == blob.getTeamNum() && blob.hasTag("player")) return;
still vertical trampolines...

AsuMagic:
TBH having to press S to bypass bouncing seems better to me. It seems to me that is the most obvious way to do it, since if you are bouncing on a trampoline, I would expect pressing S to reduce the bouncing speed (which is not the case right now AFAIK).

90° trampolines seem problematic when bouncing on the trampoline requires pressing W, and I think I could honestly get annoyed by it.

Besides, messing around to push teammates off what they're doing using trampolines is just griefing.

eps0003:
After a quick discussion in #development, people came to a conclusion that holding S to pass through team trampolines was best. This PR is ready to be merged now.

Harrison-Miller:
makes it consistent with team bridges.

PR #524, opened 2019-04-22
Assist changes

eps0003:
## Status

**IN DEVELOPMENT**

## Description

- Assist players are coloured yellow in the killfeed
  - Matches the yellow +1 hover message
  - Makes it clearer that an assist occurred in the killfeed
- Any damage dealt can give you an assist
  - This removes the 'inconsistency' of assists
- Assists give you 5 coins
  - Might make it give coins depending on the damage given by the killer compared to the assister (e.g. assister damages 3h, killer damages 1h to kill victim, the assister should get more coins than the killer)
- Fixed accolades overlapping on the scoreboard

Spikes still don't give assists. I'm working to fix that...

### DIFF
--- a/Entities/Common/HoverMessage/HoverMessage.as
+++ b/Entities/Common/HoverMessage/HoverMessage.as
@@ -1,6 +1,7 @@
 // Hover messages
 
 #include "TeamColour.as"
+#include "AssistCommon.as"
 
 // Linear interpolation
 // TODO: expose in engine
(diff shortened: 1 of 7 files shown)


Harrison-Miller:
* Decrease the font size of the assister.
* How many coins does a kill give you, We should be careful giving out coins.

bunniewormy:
my opinion:
- shouldnt matter how much dmg
- all kind of stuns should assist [stun from shieldbash, stun from water]

AsuMagic:
> all kind of stuns should assist [stun from shieldbash, stun from water]

That but absolutely requires a water nerf (stack of 1 for 20 coins?) because the coin gain might be too OP.

butterdome:
how about making a box around every kill in the killfeed ( maybe transparent). In one box would be every person who helped to kill one guy (maybe even add every method of helping (sword-, arrow-, bomb-,.. icon)). So the killfeed would look less messy and kills would be seperated clearly.

Harrison-Miller:
Closing because abandoned

PR #526, opened 2019-04-22
Russian: scoreboard, settings, Patreon

Diprog:


Diprog:
> Indentation is not correct in many places

I know, lazy to do fix that, does it really matter?

eps0003:
> I know, lazy to do fix that, does it really matter?

No, but it's good to keep it consistent

PR #531, opened 2019-04-23
Lanterns only collide with static blobs

Pine2:
## Status

**READY**

## Description

In maps like Niiiiii_GetisArena, water bombs tend to collide with the static floating lanterns. This PR only makes lanterns collide with static blobs like doors and platforms.

## Steps to Test or Reproduce

Throw a water bomb at a lantern.


AsuMagic:
This probably needs to be well tested. I expected it to cause issues in singleplayer but I couldn't actually find any (it stays properly on boats etc, but not crates for instance).

PR #532, opened 2019-04-23
Held/inventory items give back half their value when TDM round ends

eps0003:
## Status

**READY**

## Description

When a TDM round ends, you are given back half the value of the items you have in your inventory or hand (except regular arrows). This allows people to get a bit of money back on any items they purchased but didn't end up using.

Thanks to @Pine2 for the idea

## Steps to Test or Reproduce

Finish a TDM round with items in your inventory or hand

### DIFF
--- a/Entities/Common/Factory/Shop.as
+++ b/Entities/Common/Factory/Shop.as
@@ -286,6 +286,8 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
 
 					if (newlyMade !is null)
 					{
+						newlyMade.set_u16("buyer", caller.getPlayer().getNetworkID());
+
 						CBitStream params;
 						params.write_netid(caller.getNetworkID());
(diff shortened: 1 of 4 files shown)


bunniewormy:
What if someone has some items they got from a chest? Or someone got items from the enemies? This may incentivise knights to hold archer items and vice versa, so that they could get free money next round, instead of giving it to the team

AsuMagic:
I thought of the same thing as @bunniewormy, though I think coins in TDM are a bit weird since a single chest will give you a huge lot of coins and you often reach the 200 coin cap due to shops not being on every map.

eps0003:
A solution I might try is keeping track of the purchaser of every item bought. If the purchaser has an item they purchased in their inventory at the end of the round, they get some money back.

eps0003:
It now only gives back coins for items you bought from a shop and not items you pick up from chests or other players.

PR #535, opened 2019-04-23
Vehicles give coins based on damage dealt

eps0003:
## Status

**READY**

## Description

The number of coins given when damaging vehicles is based on damage dealt instead of it being a fixed amount. This also makes kegging enemy vehicles a bit more worthwhile since it gives back a fair bit of coins.

5 coins are given per heart of damage dealt (jabs and fully charged arrows do 1 heart, and slashes do 2 hearts).

## Steps to Test or Reproduce

Damage enemy vehicles using various methods

### DIFF
--- a/Entities/Vehicles/Common/WoodVehicleDamages.as
+++ b/Entities/Vehicles/Common/WoodVehicleDamages.as
@@ -50,7 +50,7 @@ f32 onHit(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@ hitt
 		{
 			if (damageowner.getTeamNum() != this.getTeamNum())
 			{
-				SendGameplayEvent(createVehicleDamageEvent(damageowner, damage));
+				SendGameplayEvent(createVehicleDamageEvent(damageowner, dmg));
 			}
 		}
 	}
--- a/Entities/Vehicles/Common/WoodVehicleDamages.as
+++ b/Entities/Vehicles/Common/WoodVehicleDamages.as
@@ -50,7 +50,7 @@ f32 onHit(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@ hitt
 		{
 			if (damageowner.getTeamNum() != this.getTeamNum())
 			{
-				SendGameplayEvent(createVehicleDamageEvent(damageowner, false));
+				SendGameplayEvent(createVehicleDamageEvent(damageowner, damage));
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


bunniewormy:
should be 5c per heart of damage, like it always was. jab 5c, slash 10c, fully charged arrow 10c.

Omen267901:
Agreed. Current coin amounts for slapping catapults and ballistas is way too high, this should keep things more balanced without being too punishing. Also fixes issues with explosives not giving a good return when used on vehicles, as mentioned.

eps0003:
So people realise, I put the coins back up to 5 per heart of damage like it currently is. This PR now only changes how many coins you get based on the damage dealt to vehicles.

bunniewormy:
Sounds good now

MrPotatobird:
> not crazy amounts - punishes teams for letting ballistas be captured too easily. ballistas should be in safe secure spots.

BIG fat disagree there. Teams are already punished for letting ballistas be captured too easily by losing an expensive ballista, which can sometimes be used against them. The fact that you can get mad dosh by poking them, *even if you don't actually capture it*, is just ridiculous. I always hated how siege was basically begging to be utilized as defensively as possible, for something that's supposedly meant to be a stalemate-killer mechanic.

For reference, a ballista has 30 hearts (150 coins) and a catapult has 24 hearts (120 coins). I don't like that they're worth that much to attack, but at least now it will be properly based on the damage being done. Pretty sure swords actually do 1/4 damage to siege, which should be 1.25 coins per jab and 2.5 for a slash, which is kinda messy but I dunno how coins work exactly.

Entities/Vehicles/Common/WoodVehicleDamages.as
```diff
@@ -50,7 +50,7 @@ f32 onHit(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@ hitt
 		{
 			if (damageowner.getTeamNum() != this.getTeamNum())
 			{
-				SendGameplayEvent(createVehicleDamageEvent(damageowner, false));
+				SendGameplayEvent(createVehicleDamageEvent(damageowner, damage));
```
MrPotatobird: looks like this should use `dmg` instead of `damage` so you're basing the coins off the actual damage being done, and you don't get 4x as much money for hitting it with a sword

```suggestion
				SendGameplayEvent(createVehicleDamageEvent(damageowner, dmg));
```

Orsovi:
What about the bomb arrow damage ?
You would have 120 coin with one bomb arrow, that's a bit OP

eps0003:
> What about the bomb arrow damage ?
> You would have 120 coin with one bomb arrow, that's a bit OP

I agree, but people seem to want the current coins per hit amount which I think is way too much.

MrPotatobird:
With this change you get coins for the amount of damage that the attack WOULD do to the siege. If a catapult is at half a heart and you hit it with a bomb arrow you're still getting 120 coins. The coins should be based on actual damage done to the siege, so either the dmg or the remaining health of the siege, whichever is lower.

PR #540, opened 2019-04-26
[Bug fix] Fixed minor issues with builder block placement preview

eps0003:
## Status

**READY**

## Description

- Fixed block placement preview showing that you can place blocks next to grass when you actually can't
- Fixed jittery tile preview when you're about to place a block while moving

## Steps to Test or Reproduce

- Attempt to place a block next to grass
- Move around with a block in your hand while hovering your cursor over a single tile

### DIFF
--- a/Entities/Common/Building/BlockPlacement.as
+++ b/Entities/Common/Building/BlockPlacement.as
@@ -173,7 +173,7 @@ void onRender(CSprite@ this)
 			if (bc.cursorClose && bc.hasReqs && bc.buildable)
 			{
 				SColor color;
-				Vec2f aimpos = bc.tileAimPos + getCamera().getInterpolationOffset();
+				Vec2f aimpos = bc.tileAimPos;
 
 				if (bc.buildable && bc.supported)
 				{
--- a/Entities/Common/Building/PlacementCommon.as
+++ b/Entities/Common/Building/PlacementCommon.as
@@ -107,16 +107,13 @@ bool isBuildableAtPos(CBlob@ this, Vec2f p, TileType buildTile, CBlob @blob, boo
 		}
 	}
 
-//printf("c");
-	bool canPlaceOnBackground = ((blob is null) || (blob.getShape().getConsts().support > 0));   // if this is a blob it has to do support - so spikes cant be placed on back
-
 	if (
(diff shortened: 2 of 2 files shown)


Entities/Common/Building/PlacementCommon.as
```diff
@@ -113,10 +113,10 @@ bool isBuildableAtPos(CBlob@ this, Vec2f p, TileType buildTile, CBlob @blob, boo
 	if (
 		(!canPlaceOnBackground || !map.isTileBackgroundNonEmpty(backtile)) &&      // can put against background
 		!(                                              // can put sticking next to something
-			canPlaceNextTo(map, left) || (canPlaceOnBackground && map.isTileBackgroundNonEmpty(left))  ||
-			canPlaceNextTo(map, right) || (canPlaceOnBackground && map.isTileBackgroundNonEmpty(right)) ||
-			canPlaceNextTo(map, up)   || (canPlaceOnBackground && map.isTileBackgroundNonEmpty(up))    ||
-			canPlaceNextTo(map, down) || (canPlaceOnBackground && map.isTileBackgroundNonEmpty(down))
+			canPlaceNextTo(map, left) || (canPlaceOnBackground && map.isTileBackgroundNonEmpty(left) && !map.isTileGrass(left.type))  ||
```
AsuMagic: Unsure if hardcoding grass there is the best solution, maybe checking against `isTileSupport` is better?

Entities/Common/Building/BlockPlacement.as
```diff
@@ -173,7 +173,7 @@ void onRender(CSprite@ this)
 			if (bc.cursorClose && bc.hasReqs && bc.buildable)
 			{
 				SColor color;
-				Vec2f aimpos = bc.tileAimPos + getCamera().getInterpolationOffset();
+				Vec2f aimpos = bc.tileAimPos;
```
AsuMagic: Whoops that wasn't supposed to get there :P
Would be nice to get it really fixed and working with interpolation but that's still better than the bug we have...

Entities/Common/Building/PlacementCommon.as
eps0003: Please take a look at what I changed to see if it is any better. I don't think `canPlaceOnBackground` is needed and it was causing the issue so I removed it. From testing, it didn't seem to change anything with placing spikes and other blob tiles.

PR #541, opened 2019-04-26
[Bug fix] Removed ability for knights/archers to insta-slash/shoot when exiting crates

eps0003:
## Status

**READY**

## Description

If a knight/archer charges a slash/bow before entering a crate, they are able to instantly slash/shoot upon exiting the crate. This used to work by sitting in vehicles but that was fixed ages ago, so I don't see why this shouldn't be fixed too.

Please don't say this is a mechanic since most people don't know about it, and the people who do most likely never use it.

## Steps to Test or Reproduce

1. Hold a crate
2. Charge a slash/bow
3. Press 'E' to enter the crate
4. Before exiting the crate, hold left mouse button
5. Exit the crate
6. Release left mouse button to slash/shoot

### DIFF
--- a/Entities/Characters/Archer/ArcherLogic.as
+++ b/Entities/Characters/Archer/ArcherLogic.as
@@ -589,17 +589,17 @@ void onTick(CBlob@ this)
 		return;
 	}
 
-	if (getKnocked(this) > 0)
+	if (getKnocked(this) > 0 || this.isInInventory())
 	{
 		archer.grappling = false;
 		archer.charge_state = 0;
 		archer.charge_time = 0;
+		this.getSprite().SetEmitSoundPaused(true);
+		getHUD().SetCursorFrame(0);
 		return;
 	}
 
 	ManageGrapple(this, archer);
-
-	if (this.isInInventory()) return;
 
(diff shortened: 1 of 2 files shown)


Panky123:
>insert snarky bug comment here<


Good fix i just wish there was a way to fix it more privately as im sure some people are learning about something new to abuse through here. Probably an insignificant portion but i didnt know this happened.

MrPotatobird:
This is a mechanic...

Ok, just kidding. I could have sworn I deliberately put in a mini-stun on exiting the crate to prevent this, as well as archer shenanigans. Archers can still triple shot out of crate, apparently (just tested it). I guess if it didn't even do what it was supposed to do, then the mini-stun can be removed.

eps0003:
I just prevented archers from insta-shooting when exiting crates and set the cursor frame of both classes to its normal state.

> I could have sworn I deliberately put in a mini-stun on exiting the crate to prevent this, as well as archer shenanigans. Archers can still triple shot out of crate, apparently (just tested it). I guess if it didn't even do what it was supposed to do, then the mini-stun can be removed.

From the looks of it, you only added a stun if the crate breaks.

PR #542, opened 2019-04-27
[Improvement] Ability to fall through ladders + ladder footstep sounds

eps0003:
## Status

**READY**

## Description

This adds the ability for players to fall through ladders by holding S before colliding with ladders. If they let go of S while colliding with a ladder, they will climb the ladder like usual. Pressing S while climbing the ladder won’t make the player fall through it until they stop colliding with the ladder.

The only issue I see with this is that players wanting to immediately climb down ladders will most likely fall through them. They would need to not press S until they have collided with the ladder, then they can climb down without falling through.

Even with this issue, I believe the benefits this will bring to movement will outweigh the issues it may cause.

I added footstep sounds when climbing ladders to make it clearer if you are actually climbing the ladder.

I also increased the tick rate from 5 to 30 to make ladders catch you more often. I'm not sure if this will impact performance at all. I don't even know why it was even reduced in the first place...

## Steps to Test or Reproduce

- Fall through a ladder while not holding W or S (you should fall through the ladder)
- Fall through a ladder while holding W or S (you should begin climbing the ladder)
- Let go of W or S while climbing a ladder (you should stay on the ladder)

### DIFF
--- a/Entities/Characters/Scripts/RunnerFootSteps.as
+++ b/Entities/Characters/Scripts/RunnerFootSteps.as
@@ -4,7 +4,7 @@
 
 void onInit(CSprite@ this)
 {
-	this.getCurrentScript().runFlags |= Script::tick_onground;
+	// this.getCurrentScript().runFlags |= Script::tick_onground;
 	this.getCurrentScript().runFlags |= Script::tick_not_inwater;
 	this.getCurrentScript().runFlags |= Script::tick_moving;
 	this.getCurrentScript().removeIfTag = "dead";
@@ -13,8 +13,16 @@ void onInit(CSprite@ this)
 void onTick(CSprite@ this)
 {
 	CBlob@ blob = this.getBlob();
-	if (/*blob.isOnGround() && */(blob.isKeyPressed(key_left) || blob.isKeyPressed(key_right)))
-	{
+
+	const bool left		= blob.isKeyPressed(key_left);
+	const bool right	= blob.isKeyPressed(key_right);
+	const bool up		= blob.isKeyPressed(key_up);
+	const bool down		= blob.isKeyPressed(key_down);
+
+	if (
+		(blob.isOnGround() && (left || right)) ||
+		(blob.isOnLadder() && (left || right || up || down))
+	) {
 		RunnerMoveVars@ moveVars;
 		if (!blob.get("moveVars", @moveVars))
 		{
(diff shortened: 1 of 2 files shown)


GoodCreativeUsername123:
What if pressing both up/down _(W&S)_ makes you ignore ladder collision and nothing else is changed

bunniewormy:
Sounds like a good idea in general, but it needs proper testing in actual matches to make sure there aren't any problems related to it and if it feels intuitive.

AsuMagic:
Other than that it feels nice on singleplayer, probably needs testing to find any edge case or issue as bunnie said.

AsuMagic:
Worth noting that the sound only plays every few ticks from what I gathered, so when you fall from pretty high on a ladder you may not even hear any sound which is a little odd.
That might be because of ladders updating on tick, though?

eps0003:
I changed the way this works. Instead of holding "S" to fall through ladders, you now always fall through. To climb ladders, you need to hold "W" or "S". @MrPotatobird and I felt like this is a bit more intuitive than the current implementation.

AsuMagic:
> I also increased the tick rate which makes ladders catch you more often. I'm not sure if it was the tick rate was reduced? I hope increasing it doesn't impact performance.

I don't recall who implemented it (maybe pirate-rob) but it would seem more sane to use the collision begin and end hooks.

> Instead of holding "S" to fall through ladders, you now always fall through. To climb ladders, you need to hold "W" or "S".

A bit odd as say archer in archer nests can often make use of those ladders. Is it possible at least to stay still by keeping both W and S pressed?

eps0003:
> A bit odd as say archer in archer nests can often make use of those ladders. Is it possible at least to stay still by keeping both W and S pressed?

Haha no, once you're climbing the ladder you don't fall when you let go.

AsuMagic:
> Haha no, once you're climbing the ladder you don't fall when you let go.

Oh, makes sense. Will test it a bit locally.

Coroz:
just make sure waterstun (or any stun?) still makes you fall down

AsuMagic:
> just make sure waterstun (or any stun?) still makes you fall down

It does.

MrPotatobird:
I'm worried about this one. I think it should be reverted for now.

I think in practice, people don't all naturally keep holding the jump button when they're falling, which is when they most want a ladder to catch them. I like it personally because I know how it works, but I think it's just not intuitive in the end. Some people like me will enjoy moving slightly faster through some unwanted ladders, but a LOT more people will be frustrated by accidentally falling when they didn't want to. It's not worth it, and they might not even be able to figure out why ladders sometimes catch them and sometimes don't.

1bardesign:
While I think the fall through behaviour could be reverted, the footstep sounds and higher tickrate might be a good addition? Assuming server load is not super adversely affected with the higher tickrate...

@eps0003 do you have time for that this week at all? We have some patreon heads waiting that it'd be good to get into release for patrons, but we can do a release with _just_ those commits if need be I guess.

PR #543, opened 2019-04-27
New players drop hearts when killed by new players

### LINKED ISSUE #523: Make new players drop hearts when killed by new players

New players [under 2 weeks, i believe] dont drop hearts when killed. This was done to prevent pubstompers from constantly healing on new players

**Proposed Change**

I think they should drop hearts again, but only if killed by new players as well. If killed by veterans, they shouldn't

eps0003:
## Status

**READY**

## Description

Resolves #523

- New player kills new player (heart is dropped)
- New player kills old player (heart is dropped)
- Old player kills new player (heart isn't dropped)
- Old player kills old player  (heart is dropped)

## Steps to Test or Reproduce

I tested by overriding the `newVictim` and `newKiller` bools in code to replicate the scenarios above

### DIFF
--- a/Entities/Items/Heart/DropHeartOnDeath.as
+++ b/Entities/Items/Heart/DropHeartOnDeath.as
@@ -3,23 +3,23 @@
 
 #define SERVER_ONLY
 
+const u32 NEW_AGE = 14;
+
 const f32 probability = 1.0f; //between 0 and 1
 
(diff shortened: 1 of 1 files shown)


blackjoker77777:
I think new players should drop a heart when killed by natural causes (bison, shark, fall damage, drowning, etc...), don't you think so? :thinking:

PR #559, opened 2019-05-03
Changing the vote displayed in chat

Orsovi:
### Status : 
Ready

### Description :
Changing the way that kickvote display in chat adding the user voted
 
### To Reproduce
Connect in a server (wait 10 minute ...) make a vote kick, now it show : A vote was started by {USER} against {KICKUSER}
This is mainly for the admin to remind who voted who easily


Rules/CommonScripts/VoteCommon.as
```diff
@@ -79,8 +79,10 @@ void Rules_SetVote(CRules@ this, VoteObject@ vote)
 		if (CanPlayerVote(vote, getLocalPlayer()))
 		{
 			client_AddToChat(
-				getTranslatedString("--- A vote was started by {USER} ---")
-					.replace("{USER}", vote.byuser),
+				getTranslatedString("--- A vote was started by {USER} against {KICKUSER} ---")
+					.replace("{KICKUSER}", getTranslatedString(vote.user_to_kick))
```
AsuMagic: The way it is might actually have been done on purpose, seeing the comments below in the code: `//more info for server and those who cant see the vote`.

It would rather make sense to get rid of the entire `if (CanPlayerVote(vote, getLocalPlayer()))` branch and just use the detailed case as I don't really get why it decides to hide information.

Orsovi: Yeah, i noticed that but as i didn't know if it was usefull for another purpose or not, i didn't touched it

AsuMagic: The issue is that the current PR would ignore whether it's a map vote or not. If the player can't vote for a map change (e.g. surrender), I believe it will show "--- A vote was started by potato against ---"

Orsovi: Hum, i'll try that

Harrison-Miller:
@Orsovi do you have any plans to update this per Asu's requests?

AsuMagic:
Closed for now as it is inactive

PR #583, opened 2019-05-11
Resupplies add to your inventory instead of refilling to a set amount

bunniewormy:
## Status

(pick one)

## Description

Whenever you wanted to resupply from a tent or a builder shop, the resupplies only refilled your materials so that you have 100 wood & 30 stone (in game time) or 300 wood & 100 stone (in build time). This changes it - now you just gain additional materials. (for example, if you resupply in build time with 100 wood and 100 stone, you will have 400 wood and 200 stone)

## Steps to Test or Reproduce

- go into Sandbox
- build a builder shop
- wait for a resupply when you still have materials


MatheusGodoi:
Please, it ends the dropping materials struggle every time just to get a resupply

Coroz:
yes this should have been so when kag came out, only little thing is maybe the archer that also keeps resupplying with the code  you have atm?

1bardesign:
Should be modified to not give more than 60 arrows (same as picking off the ground) and not continue to give if your inventory is actually _full_, or an idle builder at the tent will be a free materials fountain. I'm ok with it in theory though.

Junko-Enoshima:
more reces give more variativity for game asda

benjamingorman:
There's some code in `KickAFK.as` that attempts to guess whether someone is afk using `CControls.lastKeyPressTime`. A similar check can be added to the resupplies code to prevent AFK builders becoming fountains of mats.

bunniewormy:
> There's some code in `KickAFK.as` that attempts to guess whether someone is afk using `CControls.lastKeyPressTime`. A similar check can be added to the resupplies code to prevent AFK builders becoming fountains of mats.

i'll look at it, sounds like a better solution than a limit on max mats

furai:
This should be looked over once again, any merge conflicts fixed. 

@1bardesign can you have a look at it as well?

PR #633, opened 2019-06-15
Drill yields all materials when mining tiles/blobs

### LINKED ISSUE #618: Drills should not give 50% of the resources

title, it's unintuitive, and people think drills are used to dig faster + it's annoying when someone uses a drill on gold, which is a limited resoruce

**Proposed Change**

drills should give 100% of the resources, at least when used on ores [gold, stone]

eps0003:
## Status

**READY**

## Description

Increased from 50% to 100%
Fixed raycast issue in preparation for engine-side fix

Resolves #618 

## Steps to Test or Reproduce

Drill tiles/blobs for their materials and compare to the amount given when mining by hand

### DIFF
--- a/Entities/Industry/Drill/Drill.as
+++ b/Entities/Industry/Drill/Drill.as
@@ -290,6 +290,10 @@ void onTick(CBlob@ this)
 								{
 									for (uint i = 0; i < 2; i++)
 									{
+										//tile destroyed last hit
+										if (map.getTile(hi.hitpos).type != tile)
+											break;
+
 										map.server_DestroyTile(hi.hitpos, 1.0f, this);
 										Material::fromTile(holder, tile, 1.0f);
 									}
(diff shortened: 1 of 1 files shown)


ferrezinhre:
Did you test it? No you don't like always! now drilling gold veins gives 24 gold wonderful isnt it?
ffs

ferrezinhre:
Ah what a wounderful update, but did you don't forget about something? Oh right testing! Before this lovely update we could drill trought 6 block length dirt, now is 3, same for blocks. Let's call it unintended feature!

eps0003:
After looking at drill code a bit more, I've realised that drilling blobs only works due to a bug in the `map.getHitInfosFromArc()` function. I've made the fix for this in an engine-side PR but that will end up breaking the code for drilling blobs. I'll modify the code to fix this, but it may require a bit of testing to ensure it doesn't have any unintended side effects.

Issue #630, opened 2019-06-14
Platforms and ladders should give a small amount of wood when mined


benjamingorman:
It's inconsistent/unintuitive that mining platforms or ladders doesn't return any wood. On the other hand mining wooden doors does give wood.

Platforms/ladders should be changed so as to give a small amount of wood when mined - perhaps roughly 50% of the cost of the block similar to doors.

Vam-Jam:
Merged in a few updates ago

PR #635, opened 2019-06-15
Platforms and ladders give a small amount of wood when mined

ferrezinhre:
## Status

(pick one)

It should be free from _unintended features_.

## Description

Platforms and ladders give a small amount of wood when mined. it's for this issue https://github.com/transhumandesign/kag-base/issues/630

## Steps to Test or Reproduce

use pickaxe on platform to recive 2 woods per hit, on ladder it's 3 per hit

### DIFF
--- a/Entities/Structures/Platform/WoodenPlatform.as
+++ b/Entities/Structures/Platform/WoodenPlatform.as
@@ -12,14 +12,23 @@ void onInit(CBlob@ this)
 	CShape@ shape = this.getShape();
 	shape.AddPlatformDirection(Vec2f(0, -1), 89, false);
 	shape.SetRotationsAllowed(false);
-
+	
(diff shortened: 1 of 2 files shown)


Entities/Structures/Platform/WoodenPlatform.as
```diff
@@ -12,14 +12,23 @@ void onInit(CBlob@ this)
...
+	
 	this.server_setTeamNum(-1); //allow anyone to break them
-
 	this.set_TileType("background tile", CMap::tile_wood_back);
-
 	this.set_s16(burn_duration , 300);
 	//transfer fire to underlying tiles
 	this.Tag(spread_fire_tag);
+	
+	if (this.getName() == "wooden_platform")
```
1bardesign: Not 100% on the purpose of the name check here and in the ladder; aren't these single-blob scripts?

Issue #600, opened 2019-05-30
Ballista/Catapult and tunnel interaction


mugg91:
When you ride a ballista or catapult near a tunnel, and there are 2 tunnels only, you can travel through the tunnel with the catapult or ballista.

But when there are more than 2 tunnels, there will be a selection which tunnel to travel to and you cannot make the selection while riding a vehicle. This seems like an inconsistency.

Either make the selection possible while riding vehicles, or make the character unmount before traveling.

mugg91:
You also cannot make the selection while carrying a dinghy (but you can when carrying a crate).

mugg91:
Here is a bug you can do if you are mounted to the charger on the catapult or ballista and use the tunnel.

Because of the "free camera movement" when mounted to that spot of the vehicle, you can move the cursor out of bounds.

https://www.youtube.com/watch?v=hbGxtPw3zRo

eps0003:
The bug above looks related to #164

PR #640, opened 2019-06-18
GenericButton improvements

### LINKED ISSUE #577: Building interaction while inside crate
When inside a crate, the only way out is to jump out and break the crate.

But if you are also near a workshop of another class, you can switch to that class which makes you leave the crate without breaking it.

The other thing you can do is you can sleep in the quarters while staying in the crate. You can cancel your sleep by pressing "up", which makes a headless sprite appear at the quarters, but you will still be in the crate.

eps0003:
## Status

**READY**

## Description

New stuff:
- Buttons only display when you are not attached to or inside the inventory of a blob except the one you're interacting with
- Prevented displaying shop buttons while attached to the shop
  - This currently only prevents you from opening the shop while sleeping in a quarters (which you couldn't even buy from while sleeping)
- Made seat arrows not display while inside a blob's inventory (crates)
- Visible buttons are cleared when you attach to a blob or a blob attaches to you
  - Prevents opening a shop menu while attached to a blob (see issue below)

Issues:
- High ping still allows you to open a shop menu while attached to a blob if you're fast 
  - This is completely useless though since you can't buy from the shop while riding a siege/boat

Resolves #577 
Resolves https://github.com/transhumandesign/kag-base/issues/600#issue-450135103

## Steps to Test or Reproduce

Tested most CTF shops
Tested sieges, boats and crates
Didn't test TDM ruins, TTH related stuff (factory, scrolls), and wiring components (lever, button, coin slot, etc.) but it uses identical code to everything else so it should be fine


furai:
@eps0003 can you resolve this conflicts?

eps0003:
I swear I left a comment right after my commit. @furai can you check this PR out now that I've resolved the merge conflicts.

PR #641, opened 2019-06-18
Ladders can be placed on top of players

### LINKED ISSUE #617: You should be able to place ladders even if another player is standing in the spot you want to place the ladder in

you can't

**Proposed Change**

you should be able to. it's very annoying, especially when trying to climb up enemy tower with your team

@edit: i meant players from your team in title

eps0003:
## Status

**READY**

## Description

Resolves #617 

## Steps to Test or Reproduce

Place ladders on top of other players

### DIFF
--- a/Entities/Common/Building/BlobPlacement.as
+++ b/Entities/Common/Building/BlobPlacement.as
@@ -259,7 +259,7 @@ void onTick(CBlob@ this)
 						{
 							CBlob@ blob = b[nearblob_step];
 
-							if (blob is carryBlob || blob is this) continue;
+							if (blob is carryBlob || blob.hasTag("player")) continue;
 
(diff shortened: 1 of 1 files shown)


Entities/Common/Building/BlobPlacement.as
```diff
@@ -259,7 +259,7 @@ void onTick(CBlob@ this)
 						{
 							CBlob@ blob = b[nearblob_step];
 
-							if (blob is carryBlob || blob is this) continue;
+							if (blob is carryBlob || blob.hasTag("player")) continue;
```
AsuMagic: But then you still cannot build over some things like bushes, chickens or bisons, which should not block ladder building.
IMO it needs a broader check than the `player` tag.
I am not certain what, though.

furai:
@eps0003 could you have another take at this?

PR #657, opened 2019-06-29
Show's map name in scoreboard

Vam-Jam:
This displays the map name on the scoreboard

onInit and onRestart is used because sometimes the map names would show up with the file location + map name (e.g. map/ctf/mapnamehere.png)

### DIFF
--- a/Rules/CommonScripts/ScoreboardRender.as
+++ b/Rules/CommonScripts/ScoreboardRender.as
@@ -697,6 +697,11 @@ void getMapName(CRules@ this)
 	{
 		string[] name = map.getMapName().split('/');	//Official server maps seem to show up as
 		string mapName = name[name.length() - 1];		//``Maps/CTF/MapNameHere.png`` while using this instead of just the .png
+
+		name.clear();
+		name = mapName.split('.'); //Now remove the .png
+		mapName = name[0];	//select the first part since thats all we care about
+
 		this.set_string("map_name",mapName);
 		this.Sync("map_name",true);
 	}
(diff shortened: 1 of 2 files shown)


Vam-Jam:
This is for #657 

I also changed the indent on ScoreboardCommon.as to tabs (4) since it wasnt already.

eps0003:
Being a bit picky here but it would look a bit better if you trimmed `.png` from the map name

Vam-Jam:
Done

Rules/CommonScripts/ScoreboardRender.as
```diff
@@ -677,18 +677,32 @@ void onTick(CRules@ this)
...
+{
+	CMap@ map = getMap();
+	if(map !is null)
+	{
+		string[] name = map.getMapName().split('/');	//Official server maps seem to show up as
+		string mapName = name[name.length() - 1];		//``Maps/CTF/MapNameHere.png`` while using this instead of just the .png
+
+		name.clear();
+		name = mapName.split('.'); //Now remove the .png
+		mapName = name[0];	//select the first part since thats all we care about
```
eps0003: In singleplayer sandbox the map name is `test.kaggen.cfg`. This would only get `test`, unless that is intentional. Use `name.substr(0, name.length() - 4)` instead.

Rules/CommonScripts/ScoreboardRender.as
```diff
@@ -677,18 +677,29 @@ void onTick(CRules@ this)
...
 }
+
+void getMapName(CRules@ this)
+{
+	CMap@ map = getMap();
+	if(map !is null)
+	{
+		string[] name = map.getMapName().split('/');	 //Official server maps seem to show up as
+		string mapName = name[name.length() - 1];		 //``Maps/CTF/MapNameHere.png`` while using this instead of just the .png
+		mapName = mapName.substr(0,mapName.length() - 4);//
```
furai:
```suggestion
		mapName = mapName.substr(0,mapName.length() - 4);
```

PR #666, opened 2019-07-12
Food null error fix

Vam-Jam:
Caused by CPlayer being invalid if the blob is holding food and dies.

I've added extra null checks just in case as well

Thanks @bunniewormy for giving logs

### DIFF
--- a/Entities/Items/Food/Eatable.as
+++ b/Entities/Items/Food/Eatable.as
@@ -78,23 +78,33 @@ void onCollision(CBlob@ this, CBlob@ blob, bool solid)
 	}
 }
 
+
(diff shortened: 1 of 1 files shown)


Entities/Items/Food/Eatable.as
```diff
@@ -78,23 +78,33 @@ void onCollision(CBlob@ this, CBlob@ blob, bool solid)
...
+
+	if (isServer())
 	{
 		Heal(attached, this);
 	}
 
+	CPlayer@ p = attached.getPlayer();
+	if(p is null){return;}
+
 	this.set_u16("healer", attached.getPlayer().getNetworkID());
```
Harrison-Miller: p.getNetworkID()

PR #677, opened 2019-08-05
[WIP] Chat command rework + sandbox changes

Vam-Jam:
## Status
**IN DEVELOPMENT**: 
- New commands (!class, !tp and so on)
- Sandbox tweaks (Limited commands currently)

## Description

Updating chat commands, currently changed it from ``if else`` to a ``switch``, making it faster (so more commands wont hurt)

client_AddToChat will now show up in the clients message box as long as you call ``sendText``

More commands for sandbox coming soon, along with other tweaks

### DIFF
--- a/Rules/CommonScripts/ChatCommands.as
+++ b/Rules/CommonScripts/ChatCommands.as
@@ -7,15 +7,22 @@
 #include "MakeCrate.as";
 #include "MakeScroll.as";
 
+
+void onInit(CRules@ this)
+{
+	//Magic cmds
+	this.addCommandID("sendText");
+}
+
 bool onServerProcessChat(CRules@ this, const string& in text_in, string& out text_out, CPlayer@ player)
 {
 	//--------MAKING CUSTOM COMMANDS-------//
 	// Making commands is easy - Here's a template:
 	//
-	// if (text_in == "!YourCommand")
-	// {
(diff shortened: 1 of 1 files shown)


Rules/CommonScripts/ChatCommands.as
```diff
@@ -34,268 +41,339 @@ bool onServerProcessChat(CRules@ this, const string& in text_in, string& out tex
...
-		getBlobs(@all);
 
-		for (u32 i = 0; i < all.length; i++)
+	if(isMod)
+	{
+		switch(hash)
 		{
-			CBlob@ blob = all[i];
-			print("[" + blob.getName() + " " + blob.getNetworkID() + "] ");
+			case -149123247://!bot
```
AsuMagic: It doesn't seem very necessary to me to replace the if-else chain by a switch statement when it implies looking for hashes individually. What would be most ideal is a dictionary to functions (using [funcdefs](https://www.angelcode.com/angelscript/sdk/docs/manual/doc_global_funcdef.html)).
This could potentially help with modularity as well, so you could add commands from other scripts. I don't know how feasible this is in practice though, as behavior across different modules is not very obvious here.

Vam-Jam:
Closed as i dont plan on updating this any time soon

PR #704, opened 2019-10-04
Drill change

Vam-Jam:
Current changes (EDITED)

- Damage is 1.0f
- Heat damage reduced to 0.5f (now deals 1.5f total)
- Mat drop to 75%, instead of 100%
- Now overheats about 20/15% faster
- Ignores useless blobs so it doesnt heat up
- Doesnt decay in water

From testing it seems to work fine, but community feedback is welcomed, if its too strong, too op since it uses fire etc, say and give what you think would be a good change


Vam-Jam:
Current change list

- damage reduced to 0.75f (from 1.0f)
- damage no longer ramps up with heat (cant deal 2.0f damage)
- drill no longer decays in water (since it can be used to cool down the drill)
- drill does more damage to stone/wooden doors (overall damaged reduced from 2f to 1.25f against stone, 1.0f for wood)
- on overheat, drill takes damage (long with drill holder like before)
- on high heat, drill will now burn enemies and wooden tiles 
- drill no longer heats up to hitting blobs like mat_wood/mat_stone (this would cause it to overheat way too fast when digging)
- reduced damage against stone blocks, at high temp's you have a chance to deal more damage (requested by IguanaMan)
- walking with a drill will cool it down faster (before it use to make hardly any difference, now it makes a fairly small difference)

iguanaman:
Main issue imo is drills being used as weapons.  So my solution would be to keep it simple and just;

```
- Use b.hasTag("player") to reduce damage to players (0.5f)
- Stun on hitting shielded knight
```

Drills will still be very useful for resource gathering, tunneling and destroying defences.  Personally I'd also reduce damage to stone, but others have said they are meant to also be stalemate breakers (role belongs siege imo).

Vam-Jam:
The reduced damage should already make it low enough to make it 'harder' to get kills with ,it'll take 3 shots instead of 1 on an archer, and even more on a knight.

Im not sure about stuns, stuns in general are annoying, and unfun to play against, I feel like it'll make the drill strong if it can stun against shields. Knocking them back quite a bit would be a good solution tho

In the future, if the drill still is a weapon of mass destruction we can always still nerf it more like you suggested.

iguanaman:
You misunderstand, my changes werent supposed to be on top of yours.  Those are the only changes I think necessary.  So block damage would remain 1.0f.

The stun on hitting shield isn't stunning the knight, I'm saying stun the drill user.  To make it a less effective weapon.  Drills don't need to be weapons they already have multiple uses, and would be widely used without also being a weapon.

Vam-Jam:
Current drill change list
- Damage is 1.0f
- Heat damage reduced to 0.5f (now deals 1.5f total)
- Mat drop to 75%, instead of 100% 
- Now overheats about 20/15% faster
- Ignores useless blobs so it doesnt heat up
- Doesnt decay in water

ferrezinhre:
If I may say something, what if problem isn't a damage that drill does? What if after these changes builders still will be able to kill with drill, you will nerf damage more and more? 

I would say nerf builder mobility when using drill instead, like 0.75 speed factor and 0.50-0.25 jump factor. What it would do you ask? Hinder possibility of escape. If you want we can go further, let's say while holding drill builder speed and jump factor is 0.75 both.
That is just my thinking, now don't apply these changes with damage one, use it as replacement if you must.
About others things:
"Doesnt decay in water" - hey don't scatter around your tools, I wish that more things decay in water..
 "Now overheats about 20/15% faster" - fine we can adapt to it and abuse it.
"reduced damage against stone blocks, at high temp's you have a chance to deal more damage" - now you are mad lad, but seriously, how it will impact on preformance during collapsing thick walls and towers hmm...
"on overheat, drill takes damage" - why hurt poor drill? 1 heart for holder problem solved.
"on high heat, drill will now burn enemies and **wooden tiles**"- now let's think for moment? Did **YOU** just gave **grefers** another tool? Maybe It's time to fix report stuff? 

That would be all, cheers and love.

Vam-Jam:
@ferrezinhre updated the list, the current list of changes is 

- Damage is 1.0f
- Heat damage reduced to 0.5f (now deals 1.5f total)
- Mat drop to 75%, instead of 100%
- Now overheats about 20/15% faster
- Ignores useless blobs so it doesnt heat up
- Doesnt decay in water
(so no more fire etc)

Reason why I removed the decay in water is because its a bit weird with the drill. When your holding it in water, it cools down just fine, but sometimes you might have a lot of drills, but now you cant cool all of them down in water without holding them since they get damaged pretty fast and break.

Speed changes is a nice idea tho, if this change still makes drills a tool of mass destruction, then thats something we can add.

ferrezinhre:
My dear I think you misunderstand something, speed change or damage change, both it is just overkill for drill. And drill isn't a tool of mass destruction, but if it's in hands of clever and skilled player well, everything can happend right.

deep-fried-faceplant:
Another thought, why not make them also cost gold to unlock, or maybe the builder shop needs to be upgraded with gold to be capable of producing them. They're very good at everything they do, and it's a common strategy to just rush drills at round start to mine through walls while they're still thin. Adding the gold cost means that the team would need to commit to drilling over an additional tunnel or siege shop.

Vam-Jam:
Added Drill cooldown in inventory (was in a different PR)
Merged ferre's PR to prevent getting more wood then it costs to build stuff 
Bug fixes (thanks to bunnie for telling me about it)

ferrezinhre:
Vamist now you need add same code in my pr to ladders that give 18 wood if not heated drill and 24 if heated and platforms 20 nH and 24 H, due pr that makes em return wood. Have fun.

Entities/Industry/Drill/Drill.as
```diff
@@ -254,34 +262,48 @@ void onTick(CBlob@ this)
...
 							HitInfo@ hi = hitInfos[i];
 							bool hit_constructed = false;
-							if (hi.blob !is null) // blob
+							CBlob@ b = hi.blob;
+							if (b !is null) // blob
 							{
+								// blob ignore list, this stops the drill from overheating f a s t
+								// or blobs to increase damage to
+								int nhash = b.getName().getHash();
+								switch(nhash)
```
AsuMagic: not a fan of hash switches

Entities/Industry/Drill/Drill.as
```diff
@@ -254,34 +262,48 @@ void onTick(CBlob@ this)
...
+								{
+									case 575725963:   // mat_stone
+									case -282477713:  // mat_wood
+									case -1370030172: // mat_gold
+									{
+										continue;     // we dont want to damage them
```
AsuMagic: does this even work as intended in AS? in C++ this would do nothing because the `continue;` would apply to the switch and do nothing

PR #705, opened 2019-10-04
Added zav25 head

Vam-Jam:
Will get updated in the future as the user changes their head.


furai:
Pretty much temporary head that we'll later replace.

PR #707, opened 2019-10-04
Disabled Quarry's

Vam-Jam:
Contains a disabled quarry
Why:
Currently quarry's are disliked by the community, and are extremely OP. (Even after all its balance changes)
It grants you access to basically unlimited stone, which is good and all, but leads to more stalemate matches, and easy coin farms.

To fix this, we should disable the quarry for a few days, and see if matches improve, see if we have shorter or longer matches, people approve of this or dislike this and so on.

If this however does not provide good results, we should look into nerfing it more, like having it require coins instead of wood for example.

### DIFF
--- a/Entities/Industry/Building/Building.as
+++ b/Entities/Industry/Building/Building.as
@@ -11,7 +11,7 @@ const bool builder_only = false;
 
 void onInit(CBlob@ this)
 {
-	AddIconToken("$stonequarry$", "../Mods/Entities/Industry/CTFShops/Quarry/Quarry.png", Vec2f(40, 24), 4);
+	//AddIconToken("$stonequarry$", "../Mods/Entities/Industry/CTFShops/Quarry/Quarry.png", Vec2f(40, 24), 4);
 	this.set_TileType("background tile", CMap::tile_wood_back);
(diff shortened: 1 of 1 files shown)


MrPotatobird:
Ehh, if we're hoping to dial quarries back with nerfs (which I think we should be) then it doesn't make sense to me to disable them just to bring them back again, that's just flip-floppy. People have been playing with quarries for like a year and half now, they can live with them for a little bit longer, or however long it takes to nerf them (hopefully not too long). As much as I thought quarries were a poorly planned disaster, at this point I think it would be a bad idea to just up and revert them because there are people who think they're the greatest things since sliced bread. I think it's better to just make them weaker and more harmless so they impact the game less drastically.

Kind of a tangent but I also take issue with the "let's just do the change to see if people like it" philosophy. First of all, you say it's just "for a few days" but who knows when the next build would be? Also, we don't actually have a very good way of seeing if we have longer or shorter matches afaik, or whether "people approve". Probably some people are going to be pleased to see them gone and some people are going to be upset.

Vam-Jam:
Will make a PR with some test changes (so people can suggest changes as needed) soon

Issue #682, opened 2019-08-17
Make it possible to properly replace items in full inventory as builder


CatWiz:
**Current in-game behaviour.**

If you are trying to put an item like mine or drill instead of double stack of mats (250+ wood/stone or 50+ gold), the item you are holding gets dropped on the floor and a single stack of mats is picked up

**Proposed Change**

Make all remaining mats in that inventory slot drop on the floor and put the item in their place. Or just increase holdable stack size to 500 wood/stone and 100 gold

PR #708, opened 2019-10-04
[WIP] Changes max stack on mats

Vam-Jam:
Fixes issue #682 
Gold max stack is 100
Stone max stack is 500
Wood max stack is 500
Each take up 1 slot and one blob instead of being 2 blobs merged together, fix's this issue without having to change any engine inventory code

### DIFF
--- a/Entities/Materials/Construction/MaterialGold.as
+++ b/Entities/Materials/Construction/MaterialGold.as
@@ -1,7 +1,7 @@
 
 void onInit(CBlob@ this)
 {
-  this.maxQuantity = 50;
+  this.maxQuantity = 100;
 
   this.getCurrentScript().runFlags |= Script::remove_after_this;
 }
--- a/Entities/Materials/Construction/MaterialGold.cfg
+++ b/Entities/Materials/Construction/MaterialGold.cfg
@@ -71,4 +71,4 @@ u8 inventory_icon_frame_width          = 16
 u8 inventory_icon_frame_height         = 16
 u8 inventory_used_width                = 1
 u8 inventory_used_height               = 1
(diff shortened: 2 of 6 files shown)


MrPotatobird:
This is a bad idea if I'm understanding it correctly and you would be unable to take gold out of storage 50 at a time, and stone/wood 250 at a time. The whole reason the stack limit of gold was changed was because it was bad to have to lug more gold than you needed around when you wanted to build a forward tunnel or something.

The issue you tagged would be nice to implement as described, if possible, but this is not a good enough solution to be worth it IMO.

furai:
I don't think it's good either.

Vam-Jam:
Possible solution: Allow users to right click mats in their invo to pick up half of it.

CatWiz:
> Allow users to right click mats in their invo to pick up half of it.
Sounds good, but would be better to make it pick up 250 (50 for gold) instead of half of stack

bunniewormy:
is the new solution ever going to happen? it's very annoying for builders

another way would be to drop the rest of the stack in inventory if you're trying to replace it with something else. eg trying to put drill into a slot with 400 wood, so it puts 250 wood into your hand and drops 150 wood on ground instead of putting 250 wood into your hand and dropping the drill

CatWiz:
Is the new solution going to be implemented? Nobody objected it and i think it would be nice to see how it will work in practice

Harrison-Miller:
@Vam-Jam It seems clear that this is not the desired solution. Let's look at detecting when a player is trying to replace an inventory slot with a held item and just drop everything from that slot wether or not it fits in their hand.

PR #731, opened 2019-11-17
CSprite null checks for blob

Vam-Jam:
Sometimes blob can be null while CSprite onTick is called. This just add's a null check where needed.

Unsure how this happens, but it does. Somebody sent me their logs with it happening.


1bardesign:
This might be better to do engine-side just for `CSprite::onRender`?

Vam-Jam:
I think thats the best idea, while doing the F1 PR, i noticed there's a lot of places where CBlob is never checked for CSprite tick or render, (along with some comment saying that on rare cases it can be null)

PR #736, opened 2019-11-20
Revamped CTF Music system

Mazey:
## Status

- **READY**: The First iteration - may need tweaking and be "upgraded" further

~~There is one bug that has to be fixed related to flags before this can go live.~~

## Description

A music system that is exclusive to CTF (not shared with TTH anymore).

- Much more dynamic.
- Underground and sky are no longer ambient music but actual tracks now (similarly to classic, I consider this to be more immersive).
- Added tracks 13 & 14 to battle music (previously defined but unused in code).
- Added proper "battle start" sound (no longer using the "temporary" research sound).
- Outro music is now used in code (instead of defined but unused).
- No noticeable performance impact (when testing this with Vamist - thanks!).

To-do in a future version:
- Incorporate any feedback given if deemed valid (this was not tested extensively in live games, but we should test this soon during captains when I have time/night off).
- Play the "cavern creatures" sound (semi) at random when active underground.
- Modify Small CTF (when merged into the main branch) to use TTH music system (read: default music system before this PR) due to a system this dynamic not fitting (super) small maps very well.

If someone decides to add this to their CTF server please do share any issues/points of improvements! It should work on any fairly regular CTF server (left team vs right team, yadda yadda)


Mazey:
Ready for test/merge now.

Entities/Meta/CTFMusic.as
```diff
@@ -0,0 +1,289 @@
...
+
+	CMap@ map = blob.getMap();
+	if (map is null)
+		return;
+
+	if (rules.isWarmup())
+	{
+		gameStarted = false;
+		
+		if (timer % getTicksASecond() * 3 == 0)
```
Harrison-Miller: Instead of doing this here and similar check on line 151. Let's just set the tickFrequency 3-5 permanently.

Entities/Meta/CTFMusic.as
```diff
@@ -0,0 +1,289 @@
...
+		{
+			Vec2f pos = blob.getPosition();
```
Harrison-Miller: blob.getPosition and checkUnderground can be pulled out to line ~116 because it's done in both major branches.

Entities/Meta/CTFMusic.as
```diff
@@ -0,0 +1,289 @@
...
+				{
+					chosen = (team_num == 1 ? world_home : world_battle);
+				}
+				else // mid
+				{
+					chosen = world_calm;
+				}
+			}
+
+			if (chosen != world_battle) // below starts more "heavy" checks -> no need to execute this if we are already on battle theme
```
Harrison-Miller: chosen will never be equal to world_battle with the current logic. Perhaps you meant to make chosen a global variable?

Entities/Meta/CTFMusic.as
```diff
@@ -0,0 +1,289 @@
...
+						CBlob @b = flagBases[i];
+						if (b.hasTag("flag missing"))
+						{
+							chosen = world_battle;
+							break;
+						}
+					}
+				}
+
+				CBlob@[] blobsInRadius;
```
Harrison-Miller: should check if chosen == world_battle here to see if we already set it because of the flag.

Entities/Meta/CTFMusic.as
```diff
@@ -0,0 +1,289 @@
...
+					for (uint i = 0; i < blobsInRadius.length; i++)
+					{
+						CBlob @b = blobsInRadius[i];
+						if (blobList.find(b.getConfig()) >= 0)
+						{
+
+							if (b.getConfig() == "keg" && !b.hasTag("exploding")) // only exploding kegs
+								continue;
+
+							if (classList.find(b.getConfig() && b.hasTag("dead")) // skip corpses
```
Harrison-Miller: This isn't correct I believe you meant classList.find(b.getConfig()) && b.hasTag("dead"). Missing a paranthesis.

Entities/Special/CTF/FlagBase.as
```diff
@@ -106,6 +106,7 @@ void onTick(CBlob@ this)
 		{
 			this.Untag("flag missing");
 		}
+		this.Sync("flag missing", true);
```
Harrison-Miller: this is not necessary the tag is set on all clients without this line.

Mazey: I added this after testing because before this flag missing was not synced (it didn't work before sync / it worked after)

Issue #603, opened 2019-06-07
Prevent characters from attacking when buying items from shops


CatWiz:
**Current in-game behaviour.**

If you buy an item from shop, you also do the action that's assigned to lmb. For knights, its jabbing. This may result in accidentally destroying the enemy shop when buying items from it.

**Proposed Change**

As title says, make it impossible to attack if the player is in shop menu

PR #742, opened 2019-12-07
Shop jab prevention

Vam-Jam:
Only for knight (as archer/builder doesnt have this issue afaik (let me know if they do))

Knights no longer jab while buying items
Knights no longer jab if the player just switched to it (happens when you change class at an enemy shop)


Harrison-Miller:
Based on this issue: https://github.com/transhumandesign/kag-base/issues/603

PR #743, opened 2019-12-08
Add more zoom levels to the camera

andreyrk:
## Status
READY

## Description
Camera now have more zoom levels to switch to, making it easier to set to a mode you are more comfortable with depending on your monitor/screen resolution.

OBS: Now using Clamp instead of the manual if/else chains. Also tweaked the zoom speed and tap time to better suit the changes. The new zoom levels should not have much effect on pixel scaling quality, as perfect 16-32 multiples were used.

## Steps to Test or Reproduce
Simply fetch the PR's changes, launch the game and use your mouse's scroll to see the new zoom levels.

### DIFF
--- a/Entities/Common/Controls/StandardControls.as
+++ b/Entities/Common/Controls/StandardControls.as
@@ -323,7 +323,7 @@ void AdjustCamera(CBlob@ this, bool is_in_render)
 	CCamera@ camera = getCamera();
 	f32 zoom = camera.targetDistance;
 
-	f32 zoomSpeed = 0.1f;
+	f32 zoomSpeed = 0.05f;
 	if (is_in_render)
 	{
 		zoomSpeed *= getRenderApproximateCorrectionFactor();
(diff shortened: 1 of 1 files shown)


Vam-Jam:
Code is clean, formatting is correct. I tested locally in single player and it works fine. I’m unsure about these changes personally (since I’m use to the 3 camera zoom), but I also like changes here.

If other players could give their opinion on this change, it would be helpful.

Entities/Common/Controls/StandardControls.as
```diff
@@ -357,39 +363,17 @@ void ManageCamera(CBlob@ this)
...
-			}
-			else if (zoomLevel == 3)
-			{
-				zoomLevel = 2;
-			}
-			else if (zoomLevel == 1)
-			{
-				zoomLevel = 2;
-			}
+			zoomLevel = Maths::Clamp(zoomLevel + 1, 0, 8);
```
Mazey: Probably change clamp into min/max, or keep clamp and get rid of the double if statement.

andreyrk: Thanks, made the change as requested.

MrPotatobird:
Uh, some players (like me) like to quickly toggle between the normal zoom (which is comfortable) and the zoomed-out view so we can check out what's happening further away occasionally. I don't think this is a good change. There isn't necessarily a single zoom level that a player will want to stay at, and it sounds like this will make it fussier to switch between the useful zoomed-out view and the level of zoom that you're more comfortable with.

I dunno, I'll test it out sometime

andreyrk:
Good point. I'm open to hearing ideas on ways this could be implemented without making it too unfamiliar to people used to it.

To explain my reasoning behind the change, I've personally had issues with the three zoom levels, the nearest one doesn't feel usable for any class due to being too zoomed in. As for the other two I constantly find myself switching to the farthest one as the 1:1 zoom is not wide enough when there are archers and siege above,but I end up losing the awareness of where my character is and where others are.

For this, my change makes me avoid the big zoom change at each scroll, which gives takes away my attention and awareness every time I need to see something farther away.

So if you have some time to try out the change and tell whether you still have the same opinion, please do.

andreyrk:
This last amend includes what @Vam-Jam suggested on Discord: holding CTRL uses the extra zoom modes while keeping the 3-zooms as default. This way everyone is happy.

andreyrk:
OBS: The amend also changes from 8 to 6 levels, helps maintain fewer steps while keeping reasonable distance changes. Both modes are tracked separately but the transitions between the two are seamlessly.

Issue #725, opened 2019-11-03
Class badges in killfeed


Turtlebutt:
**Current in-game behaviour.**

The current killfeed only displays name and how the person died. example : deynarde *sword* KF9000.

**Proposed Change**

Add a class badge with the respective team color before the name. I feel like this could be a nice feature, aswell if someone has their steam name changed to a character that doesn't show, you could still see why someone died.

The before and after.
![rga](https://user-images.githubusercontent.com/41750971/68088245-9b2bc600-fe65-11e9-88a1-e2e46e2126d3.png)

PR #744, opened 2019-12-10
Add player icons to the killfeed

andreyrk:
## Status
READY

## Description
Implements the feature discussed in #725.

killfeed

OBS: Also slightly tweaked the hitterIcon centering.

## Steps to Test or Reproduce
Fetch the changes and trigger the killfeed in-game.

### DIFF
--- a/Rules/CommonScripts/KillMessages.as
+++ b/Rules/CommonScripts/KillMessages.as
@@ -12,10 +12,19 @@ class KillMessage
 {
 	string victim;
 	string victim_tag;
+	string victim_head;
+	int victim_head_index;
+	
 	string attacker;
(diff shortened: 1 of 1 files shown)


eps0003:
I’m against adding more things to the killfeed because it makes it too cluttered.

Harrison-Miller:
Yeah I agree, this change just adds more to an already cluttered kill feed.

AsuMagic:
Plus kill messages are slow enough to render already because GUI draw calls are expensive, which is not ideal. That can be seen in the scoreboard which has got increasingly slower over the years.

andreyrk:
Thx for the feedback, that's understandable 👍

PR #751, opened 2019-12-24
[Change] OnHeal changes

Vam-Jam:
Item's that get healed require a player before we play their particle effect

This prevents drill's and what not from having the particles.


Vam-Jam:
No longer randomly plays the heal particle when you shield (due to damage being 0 and the while being >= 0)

XORRandom the position of the particle too, before it was random position in a circle, now its a random position within the circle

PR #754, opened 2020-01-14
Revert "Christmas cheer"

bunniewormy:
Reverts transhumandesign/kag-base#750
christmas over


AsuMagic:
We really need to get it as part as the holiday stuff...

PR #755, opened 2020-01-16
[fix] Tree no longer extends hitboxes of objects

Vam-Jam:
image

Ever had this happen and complained about it?

This pr fixes it!

Big thanks to Greendino for showing me how to reproduce this, and GoldenGuy for giving me an idea on how to fix the issue.

### DIFF
--- a/Entities/Natural/Trees/TreeSync.as
+++ b/Entities/Natural/Trees/TreeSync.as
@@ -34,7 +34,12 @@ void InitTree(CBlob@ this, TreeVars@ vars)
 	//prevent building overlap
 	CMap@ map = this.getMap();
 	const f32 radius = map.tilesize / 2.0f;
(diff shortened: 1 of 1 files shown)


Harrison-Miller:
Is there an issue on how to reproduce the bug.

Vam-Jam:
The bug reproduce wasnt included since people can (and have) used it to prevent people from building near flags, spawns etc

I dont see any existing issue about it, its been posted every so often on discord.

The bug is just the tree will extend a 'no build' Sector, using the function `map.getSectorAtPosition(pos, "no build");`
This will only pick the first one there out of the group (always the shop/underlying no_build, since its the first element in the array), and then extends it to the size of the tree.

The solution is just to store the Sector by using blob.set('name', data), so we can get the sector we created without extending anything else.

Issue #712, opened 2019-10-10
[Bug] Drill progress bar jittery when moving


eps0003:
**Description**

Title

**Reproduction**

Walk with drill in hand

**Expected Behavior**

It should be buttery smooth

furai:
I think also when riding vehicle name tags suffer from the same issue.

Vam-Jam:
I had a PR that was reverted that fixed this issue. getInterp seems to have issues on a dedicated server.

PR #757, opened 2020-01-21
[fix] Drill screen blur fix

Vam-Jam:
After hours of debugging, I figured this is the best way to fix this issue.

Currently, using engine side getInterp on an attached blob will not work correctly, due to the vars.oldpos being out of date, the same as getPosition or just completely wrong. This only happens on dedicated servers.

I have no clue currently why this happens, but for now this is a good solution since its a fairly small feature and I dont feel like its a big enough issue to spent more hours looking into it.

This PR makes it work script side. We use our player (holders) as the interp place, then work out the difference between that player and the blob to get the interp position of the drill.

This fixes issues #712


Vam-Jam:
Closed as it will be put into a different PR that touches the same file

PR #761, opened 2020-02-05
Drill changes part2

Vam-Jam:
**What has changed**
- Drill UI is no longer blurry when moving apart of #757 
- Drill can no longer be used until the game has lasted more then 4 mins
- Drill now freely follow's your mouse, instead of being locked at 45 degrees near your mouse
- Drills deal half damage to shielded people

**Why did you change x**
Drills can no longer be used until game reaches 4 mins to help prevent flag rush early game, which most people agree on is not fun what so ever.

Drill follows your mouse instead of being at a 45 degree angle near your mouse because it use to lead to issues where you would dig a tile, but would do no damamge. This lets you move it freely, and just feels much better.

Drill now deals half damage to shielded people. This might change if people would rather have something different happen when you shield a drill.

**I dont like x**
Please talk about it down below, at very least the UI fix will be put in, but the rest are up to you guys to talk about.

### DIFF
--- a/Entities/Industry/Drill/Drill.as
+++ b/Entities/Industry/Drill/Drill.as
@@ -58,9 +58,9 @@ void onInit(CBlob@ this)
 	}*/
 
 	this.set_u32("hittime", 0);
-	this.Tag("place45");
-	this.set_s8("place45 distance", 1);
-	this.Tag("place45 perp");
+	//this.Tag("place45");
+	//this.set_s8("place45 distance", 1);
+	//this.Tag("place45 perp");
 	this.set_u8(heat_prop, 0);
 	this.set_u16("showHeatTo", 0);
 	this.set_u16("harvestWoodDoorCap", 4);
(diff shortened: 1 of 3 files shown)


Harrison-Miller:
To  sum up some of the discussion in the discord.
It sounds like most of these changes are pretty agreeable especially the free aim cursor.

The only one that's been hotly debated is the 4 minute wait on being able to use the drill the arguments are:
* This is mostly a non-issue that happens on low population servers and very rarely.
* This is a poor design decision (it feels hacky, and makes yet another exception to a rule).

My recommendation is that if you want to make it less likely that somebody will use it at the beginning of the game then you should increase the cost. I think increasing the coin cost is the best candidate.

Harrison-Miller:
@Vam-Jam can we drop the 4 minute thing so we can merge all the other changes and then revist the discussion on what to do about rushing another time.

Vam-Jam:
Reverted 4 min timer as requested.

Entities/Industry/Drill/Drill.as
```diff
@@ -176,20 +177,20 @@ void onTick(CBlob@ this)
 	sprite.SetEmitSoundPaused(true);
 	if (this.isAttached())
 	{
-		this.getCurrentScript().runFlags &= ~(Script::tick_not_sleeping);
```
Harrison-Miller: I don't think this was supposed to be removed.
I noticed there's a lot of bugs with the drill after this PR.

Other classes can use the drill (no rotation).
The drill doesn't cool while held in water.
Probably others.

Vam-Jam:
Drill no longer locks in place (was fun to debug, blob needed a "place norotate" tag since blobplacement.as would attempt to lock its angle to 0.0f).

Drill works correctly in water. The thing that caused this was me disabling the shape_active when we picked the drill up, however it would mean that the blob position would not update. 
Instead i changed the gravity of the blob to 0, and it seems to have also prevented the blur effect with the HUD

PR #783, opened 2020-03-26
[Fix] Only sync when needed

Vam-Jam:
I found this while messing with sync engine side

CTF flags sync every tick * per flag in map "flag missing"

Archers always sync their bool hasarrows, when we only need to sync it when it swaps. This is also per player.

Each player also syncs air_count every few ticks, which once again is not needed unless aircount is changing. 

This PR means they don't sync until its required. This should save network resources, not like it’s very resource intensive anyway.


Harrison-Miller:
This probably needs some thorough testing later.

PR #787, opened 2020-04-18
[Changes] Sandbox changes

Vam-Jam:
This PR changes sandbox, mostly aiming to fix issues that ive seen while playing on sandbox.

# The changes;
 ~ All chat commands are re-enabled by default
 ~ Command blacklist now in place
 ~ Command chat delay has been added (3 seconds by default)
 ~ Admins ignore chat delay & blacklist
 ~ Server will auto remove kegs/mines when 20+ are in the server
 ~ Added coin command that accepts coin amount

# Why;
All chat commands have been re-enabled because just having access to stone and wood was pretty boring

Blacklist is in place to stop people being annoying, by spawning in 50 necromancers and so on, along with this we can disable things that can be used to crash/harm the server
Current list includes;
- Necromancer 
- Greg
- Bison
- Shark
- Hall
- CTF_flag
- flag_base

Chat commands have a delay of 3 seconds. This means that you can only do one chat command every 3 seconds. This is so people can't use macro's to spawn mines and kill the server/cause bad deltas

Server auto removes kegs/mines when over 20+ is reached because its sandbox, and the only use for them most the time is to make a pit and blow it up. 
The cap is there to stop this from slowing the server down, and making the server unjoinable

Admins ignore the chat delay & blacklist because they can be trusted not to abuse it heavily.

### DIFF
--- a/Rules/CommonScripts/ChatCommands.as
+++ b/Rules/CommonScripts/ChatCommands.as
@@ -7,6 +7,18 @@
 #include "MakeCrate.as";
 #include "MakeScroll.as";
 
+const bool ChatCommandCoolDown = false; // enable if you want cooldown on your server
+const uint ChatCommandDelay = 3 * 30; // Cooldown in seconds
+
+
+void onInit(CRules@ this)
+{
+	this.addCommandID("SendChatMessage");
+}
+
+
+
+
 bool onServerProcessChat(CRules@ this, const string& in text_in, string& out text_out, CPlayer@ player)
 {
 	//--------MAKING CUSTOM COMMANDS-------//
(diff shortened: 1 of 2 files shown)


Rules/Sandbox/Scripts/Sandbox_Rules.as
```diff
@@ -378,5 +386,41 @@ void Reset(CRules@ this)
...
+		{
+			blob.server_Die(); // wont explode because its just been made
+		}
+	}
+	else if (blob.getName() == "keg")
+	{
+		kegCount += 1;
+		if (kegCount > maxKegs)
+		{
+			blob.server_Die();
```
eps0003, line 409: Wouldn't it make more sense if the oldest keg is removed, not the most recent? Couldn't someone just hide a lot of kegs underground somewhere so it's impossible for someone else to spawn new ones?

This also applies to mines too.

Rules/CommonScripts/ChatCommands.as
```diff
@@ -313,3 +364,25 @@ bool onClientProcessChat(CRules@ this, const string& in text_in, string& out tex
...
+
+bool IsBlacklisted(string name)
+{
+	return  name=="hall" || // used to dig hole to bottom of the map, spawns lots of migrants
+			name=="shark" || // greif
+			name=="bison" ||
+			name=="necromancer" || // annoying / grief
+			name=="greg" || // annoying / grief
+			name=="ctf_flag" || // sound spam
+			name=="flag_base";// sound spam / grief
```
eps0003, line 387: You should put all these names in an array at the top of the file so it's neater.

PR #796, opened 2020-05-09
[Changes] Trampoline can now rotate 360 like drill

### LINKED ISSUE #790: Make trampolines rotate when in hand depending on cursor
Currently trampolines only can point diagonally (45 degree angle) or up. You could change this so that they rotate smartly depending on the position of your cursor.

Vam-Jam:
Code done by @CatWiz, asked me to make a PR for him

**Key features**
- Trampolines now follow the mouse in 360 degree, like drill does
- Press ctrl to revert back to 45 degree lock like before

Resolves #790

### DIFF
--- a/Entities/Structures/Trampoline/TrampolineLogic.as
+++ b/Entities/Structures/Trampoline/TrampolineLogic.as
@@ -1,4 +1,5 @@
 // TrampolineLogic.as
+#include "StandardControlsCommon.as"
 
 namespace Trampoline
 {
@@ -41,12 +42,20 @@ void onTick(CBlob@ this)
 
 	Vec2f ray = holder.getAimPos() - this.getPosition();
 	ray.Normalize();
-
+	
 	f32 angle = ray.Angle();
-	angle = angle > 135 || angle < 45? (holder.isFacingLeft()? 135 : 45) : 90;
-	angle -= 90;
+	CControls@ controls = getControls();
 
-	this.setAngleDegrees(-angle);
+	if (controls.isKeyPressed(KEY_LCONTROL))
+	{
+		angle = angle > 135 || angle < 45? (holder.isFacingLeft()? 135 : 45) : 90;
+		angle -= 90;
+		this.setAngleDegrees(-angle);
+	}
+	else
+	{
+		this.setAngleDegrees(-angle+90);
+	}
 }
 
 void onCollision(CBlob@ this, CBlob@ blob, bool solid, Vec2f normal, Vec2f point1, Vec2f point2)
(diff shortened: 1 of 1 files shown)


andreyrk:
Does this actually affect the direction of impulse?

bunniewormy:
One issue is that you can't actually drop a trampoline at the angle you want (for example to put a trampoline on top of tower between some blocks), even when holding CTRL.

AsuMagic:
I think we probably should have new engine-side keybindings for these since it is not explained anywhere or intuitive at all that the ctrl key is useful for that (and even then you'd have to look into the bindings again to figure this out).

Would a E bubble button make more sense maybe?

andreyrk:
Ideally I think it would be better to not have an additional key for that and leave it as default. But I don't know what effect that would have on gameplay mechanics as to which people are used to

Entities/Structures/Trampoline/TrampolineLogic.as
```diff
@@ -41,12 +42,20 @@ void onTick(CBlob@ this)
...
 	ray.Normalize();
-
+	
 	f32 angle = ray.Angle();
-	angle = angle > 135 || angle < 45? (holder.isFacingLeft()? 135 : 45) : 90;
-	angle -= 90;
+	CControls@ controls = getControls();
 
-	this.setAngleDegrees(-angle);
+	if (controls.isKeyPressed(KEY_LCONTROL))
```
eps0003, line 49: Any reason why you chose ctrl over shift?
I think you should use `AK_BUILD_MODIFIER` which is bound to ctrl by default and is bindable.
```suggestion
	if (controls.isKeyPressed(AK_BUILD_MODIFIER))
```

Entities/Structures/Trampoline/TrampolineLogic.as
```diff
@@ -41,12 +42,20 @@ void onTick(CBlob@ this)
...
+	{
+		angle = angle > 135 || angle < 45? (holder.isFacingLeft()? 135 : 45) : 90;
+		angle -= 90;
+		this.setAngleDegrees(-angle);
+	}
+	else
+	{
+		this.setAngleDegrees(-angle+90);
```
eps0003, line 57: The trampoline spazzes out if you move your cursor to the centre point of rotation. There should be a threshold at which it stops updating its angle if the cursor is too close.

eps0003 (changes requested):

eps0003:
The snapped rotation while holding ctrl is only visible client-side.

Entities/Structures/Trampoline/TrampolineLogic.as
```diff
@@ -40,13 +41,28 @@ void onTick(CBlob@ this)
...
+	if (controls.isKeyPressed(KEY_LCONTROL))
+	{
+		if (Maths::Abs(angle) > 45)
+		{
+			angle = holder.isFacingLeft() ? -45 : 45;
+		}
+		else
+		{
+			angle = 0;
+		}
```
eps0003, line 57: I think the snapped rotation should work in all 8 directions
```suggestion
		angle = Maths::Floor((angle + 22.5f) / 45) * 45;
```

Vam-Jam:
Closed - @CatWiz request

PR #812, opened 2020-05-20
Update ScoreboardRender.as

MonkeyFeats:
## Status
- **READY**

## Description
Fixes Scoreboard scaling to full screen width

## Steps to Test or Reproduce
Press Tab


MonkeyFeats:
How am I going to do that? I would have to make a new font.

PR #822, opened 2020-05-29
[Fix] Scoreboard right click text not showing

Vam-Jam:
Bug with CPlayer, on localhost, getLocalPlayer().set_int, strign etc would save, but on dedicated servers, this does not work, the value is always 0/default.

Moved to CRules to prevent this, along with adding "client_" to all set names.


Harrison-Miller:
I'm going to merge this but can we look into fixing the issue with local player not keeping properties?

PR #846, opened 2020-06-14
Minor fixes to pickup-wheel rendering

Nananas:
## Status

## Description

Entry rendering of pickup wheel seem to have wrong offset. 

Also reduced radius of wheel. The original radius felt too big. Please review.

### DIFF
--- a/Entities/Common/GUI/WheelMenuCommon.as
+++ b/Entities/Common/GUI/WheelMenuCommon.as
@@ -11,7 +11,7 @@ namespace WheelMenu
 	const SColor pane_title_color(0xFFCCCCCC);
 	const SColor pane_text_color(0xFFFFFFFF);
 
-	const float item_distance = 0.4f;
+	const float item_distance = 0.25f;
 
 	const float hover_distance = 0.07f;
 	const float auto_selection_distance = 0.3f;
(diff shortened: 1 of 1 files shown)


Entities/Common/GUI/WheelMenuCommon.as
```diff
@@ -11,7 +11,7 @@ namespace WheelMenu
 	const SColor pane_title_color(0xFFCCCCCC);
 	const SColor pane_text_color(0xFFFFFFFF);
 
-	const float item_distance = 0.4f;
+	const float item_distance = 0.25f;
```
Harrison-Miller, line 14: Changing this constant effects all while menus in the game. Instead make the WheelMenu base class have an item_distance member that defaults to this constant (0.4) and override it in the PickupWheelMenu

Nananas: Woops. Fixed. Might be interesting to have as option in settings. 
Not sure how to do that though.

Harrison-Miller: You can create custom menus for options (see emote binding menu) and you can save/load preferences to cfg files. Though for this right now I wouldn't recommend it.

Probably better solved with proper UI scaling.

Issue #788, opened 2020-04-24
Make an empty bucket fill with water if you use a water bomb or water arrow on it


bunniewormy:
**Describe the feature**

title

just small feature

PR #847, opened 2020-06-14
Refill bucket using water bomb

Nananas:
## Status

## Description

Allow refilling of buckets using water bombs. Related to #788. This implementation is very simple. As such, it allows an "unlimited" buckets to be filled with a single water bomb. No water quantity calculations etc are done, but I do not feel they are required. Currently the buckets are also filled fully (3 splashes). 

Thoughts?

## Steps to Test or Reproduce

- spawn empty bucket(s)
- buy water bomb, let it explode on the buckets
- buckets are now full of water

### DIFF
--- a/Entities/Items/Bucket/Bucket.as
+++ b/Entities/Items/Bucket/Bucket.as
@@ -1,7 +1,7 @@
 
 #include "Hitters.as";
 #include "SplashWater.as";
-
+#include "ArcherCommon.as";
 //config
 
 const int splash_width = 9;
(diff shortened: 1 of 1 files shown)


Harrison-Miller:
This is a duplicate issue of: https://github.com/transhumandesign/kag-base/issues/788
Please review existing issues before creating a new one.

Harrison-Miller:
can you modify this to also include water arrows.

Nananas:
Now it should work for water arrows and bombs.

Harrison-Miller:
WRT to unlimited buckets from a single water arrow/bomb. You could add a tag to the hitter blob when it fills a bucket, and check for that before filling.

Nananas:
Thanks for the help. Didn't think of tagging. I have added the 'water quantity calculations' too.
How does this affect synchronization with other players, though? Will they have the same buckets filled?

Harrison-Miller:
Yeah that's a good question, I think the onHit hooks should be called in the same order. However to be safe you could check is this is the server before doing the logic then sync the filled property of the bucket. You need to check somewhere in the bucket logic to update the animation.

Nananas:
Now it should calculate refill on the server and use `sendCommand` to sync client's animation/filling.
I tried using `Sync()`, but that did not work well in combination with the `sendCommand`.

PR #848, opened 2020-06-14
Consider quantities for pickup priority

Nananas:
## Status

## Description

A tricky PR. This way picking up items will give a slight nudge towards bigger stacks of items (depending on the maxQuantity). This should be tested, maybe only applied to resources?

## Steps to Test or Reproduce

- Make a small and a large stack of resources
- Quick-tap pickup or Pickup-wheel
- Notice the larger stack gets priority


Nananas:
Used different implementation. Should be more straightforward

PR #850, opened 2020-06-16
replaced deprecated .length() with .size()

jammer312:
## Status
## Description

Pretty much what title says, this PR replaces .length() calls for arrays with .size(). This is done to prevent warnings from being generated on staging build (where newer angelscript version is used?).
Also removes random whitespaces at the end of some lines in these files.


jammer312:
Turns out there is more to change, i.e. PR is not yet ready.

PR #853, opened 2020-06-21
Updated tips

CatWiz:
## Status

## Description

Rephrased some of the tips to make them more informative, added new ones, removed a tip that tells you to open dead bodies' inventores


Rules/CommonScripts/HelpfulDeathTips.cfg
```diff
@@ -98,5 +97,21 @@ tips = To hide these tips, and other in-game help, hit F1.;
...
+		Double slashing deals damage faster than normal slashing or jabbing.;
+		A drill will overheat very quickly when drilling doors and platforms.;
+		Placing platforms inside walls doesn't help against explosions, no matter what direction they're facing.;
+		You can use a water bucket to extinguish lit kegs.;
+		Carrying a sponge in your inventory greatly reduces the length of next 2 stuns from water bombs or water arrows.;
+		Shielding water bombs and water arrows halves the stun time.;
+		Stone doors are durable, but very expensive - dont place too many of them in one doorway.;
+		Placing a mine on your flag will make it difficult for the enemy to take it without dying.;
+		As an archer, you can shoot enemy mines to blow them up. But be careful, as mine's explosion can still kill you and your teammates!;
+		You can blow up enemy mines by throwing bodies of your teammates at them.;
```
Harrison-Miller, line 116: Could add one about disabling saws with a body.

Rules/CommonScripts/HelpfulDeathTips.cfg
```diff
@@ -98,5 +97,21 @@ tips = To hide these tips, and other in-game help, hit F1.;
...
+		You can get resupplies in class workshops, not just at the tent!;
 		You can highlight useful items for your class by holding the pickup key.;
+		As a knight, you can always jab the enemy after successfully slashing them to deal extra heart of damage.;
+		You should always have one transport tunnel near your tent and another one at your tower for quick transportation.;
+		You should always have a working saw trap on your spawn tunnel.;
+		You can throw saws at logs to quickly turn them into usable wood.;
+		Knights can cut down trees quicker than builders. If your team has a tree farm, cut down trees as knight to save time for your builders.;
+		Double slashing deals damage faster than normal slashing or jabbing.;
+		A drill will overheat very quickly when drilling doors and platforms.;
+		Placing platforms inside walls doesn't help against explosions, no matter what direction they're facing.;
```
Harrison-Miller, line 109: In the future update it will work if placed the correct direction (they will be more blast resistant than a normal block). I might honestly just remove this one in general.

Rules/CommonScripts/HelpfulDeathTips.cfg
```diff
@@ -98,5 +97,21 @@ tips = To hide these tips, and other in-game help, hit F1.;
 		Knights can use their swords to destroy wooden blocks and doors, but not stone;
 		Doors and siege weapons are vulnerable to explosives.;
 		Bison can be tamed with burgers.;
-		Resources can be resupplied by changing classes in workshops, not just at the tent!;
+		You can get resupplies in class workshops, not just at the tent!;
 		You can highlight useful items for your class by holding the pickup key.;
+		As a knight, you can always jab the enemy after successfully slashing them to deal extra heart of damage.;
+		You should always have one transport tunnel near your tent and another one at your tower for quick transportation.;
+		You should always have a working saw trap on your spawn tunnel.;
```
Harrison-Miller, line 104: This is a specific building strategy which I don't think needs to be covered in the tip.

CatWiz: Having one tunnel near your tent and another tunnel (tunnels) somewhere in front tower is the only way of using tunnels that makes sense. It could be changed to something like "if you're using tunnels, make sure they are easily accessible and trapped" so it doesnt outright tell you to build this and build that.

Rules/CommonScripts/HelpfulDeathTips.cfg
CatWiz: If you make a "block platform block" wall and shoot it with a bomb arrow, platforms will take damage after bomb arrow hits the outer layer of the wall, same as in release. In test build they dont instantly break, but considering the fix to platforms blocking explosions, there will be no reason to place platforms instead of normal wooden blocks. 
Though this tip is targeted at people who use a specific old building meta that no longer works, so yeah, its better to just remove it.

Rules/CommonScripts/HelpfulDeathTips.cfg
```diff
@@ -23,25 +23,25 @@ tips = To hide these tips, and other in-game help, hit F1.;
...
 		Enemies can use your workshops. Don't let them buy explosives in your base!;
 		Teamwork is the key to success!;
-		Try to help your team mates if the situation gets tight.;
+		Try to help your team mates.;
 		As a knight, you can put your shield up in the air to glide.;
 		Wood blocks are cheaper than stone blocks, but are also weaker and can catch fire.;
-		1 bomb can deal up to 3 hearts of damage.;
-		Drills are a fast way to dig, but only yield 50%. Don't use them on precious gold or during build time!;
+		A bomb can deal up to 3 hearts of damage.;
+		Drills only yield 75% of stone, but 100% of gold.;
```
eps0003, line 35: It might not be wise to include such specific information considering drills are constantly being tweaked.

Rules/CommonScripts/HelpfulDeathTips.cfg
```diff
@@ -23,25 +23,25 @@ tips = To hide these tips, and other in-game help, hit F1.;
...
 		Good players can play any class when needed.;
 		Make sure you're helping your team advance; don't just hide out in the back!;
 		Ride a shark on land and press up to pick him up.;
 		A red name indicates a server admin - ask them for help if you need it.;
 		A bush plus a mine equals happy fun time.;
 		Enemies can walk through your doors if open them. Don't let them in!;
 		If you see an object/body holding a door open - move it away!;
-		Visit the KAG forums for more friendly help and advice!;
-		A bomb's timer lasts four seconds before exploding;
+		Visit the KAG discord for more friendly help and advice!;
```
eps0003, line 43: Discord is a proper noun and should be capitalised.

Rules/CommonScripts/HelpfulDeathTips.cfg
```diff
@@ -98,5 +97,20 @@ tips = To hide these tips, and other in-game help, hit F1.;
 		Knights can use their swords to destroy wooden blocks and doors, but not stone;
 		Doors and siege weapons are vulnerable to explosives.;
 		Bison can be tamed with burgers.;
-		Resources can be resupplied by changing classes in workshops, not just at the tent!;
+		You can get resupplies in class workshops, not just at the tent!;
```
eps0003, line 100: "You can get resupplies ~~in~~ **from** class workshops, not just at the tent!"

Rules/CommonScripts/HelpfulDeathTips.cfg
```diff
@@ -98,5 +97,20 @@ tips = To hide these tips, and other in-game help, hit F1.;
...
 		You can highlight useful items for your class by holding the pickup key.;
+		As a knight, you can always jab the enemy after successfully slashing them to deal extra heart of damage.;
+		You should always have one transport tunnel near your tent and another one at your tower for quick transportation.;
```
eps0003, line 103: "at your tower" is a bit vague. I think it should instead suggest building a second tunnel "closer to the enemy".

eps0003: Also, it isn't essential to use tunnels for quick transport (ballistae exist), therefore, I recommend changing "You should always" to "It is advantageous to".

Rules/CommonScripts/HelpfulDeathTips.cfg
```diff
@@ -98,5 +97,20 @@ tips = To hide these tips, and other in-game help, hit F1.;
...
+		It's good to have some protection against enemies on your transport tunnels, such as traps or mines.;
```
eps0003, line 104: "It's a good idea to protect your transport tunnels from enemies by creating a trap at the entrance of your tunnel."
This way of phrasing implies that creating a trap at the exit of your tunnel isn't as effective as at the exit. Also, a mine can be considered a trap.

Rules/CommonScripts/HelpfulDeathTips.cfg
```diff
@@ -98,5 +97,20 @@ tips = To hide these tips, and other in-game help, hit F1.;
...
+		You can throw saws at logs to quickly turn them into usable wood.;
+		Knights can cut down trees quicker than builders. If your team has a tree farm, cut down trees as knight to save time for your builders.;
+		Double slashing deals damage faster than normal slashing or jabbing.;
+		A drill will overheat very quickly when drilling doors and platforms.;
+		You can use a water bucket to extinguish lit kegs.;
+		Carrying a sponge in your inventory greatly reduces the length of next 2 stuns from water bombs or water arrows.;
```
eps0003, line 110: Once again, I don't think its good to include such specific information in case sponges get buffed/nerfed. I suggest:
"Carrying a sponge in your inventory greatly reduces the duration of stuns from water bombs and water arrows".
The player can figure out that it is reusable by trying it out.

Rules/CommonScripts/HelpfulDeathTips.cfg
```diff
@@ -98,5 +97,20 @@ tips = To hide these tips, and other in-game help, hit F1.;
...
+		Shielding water bombs and water arrows halves the stun time.;
+		Stone doors are durable, but very expensive - dont place too many of them in one doorway.;
+		Placing a mine on your flag will make it difficult for the enemy to take it without dying.;
+		As an archer, you can shoot enemy mines to blow them up. But be careful, as mine's explosion can still kill you and your teammates!;
```
eps0003, line 114: "But be careful, as ~~mine's~~ **the** explosion can still kill you and your teammates!"

Rules/CommonScripts/HelpfulDeathTips.cfg
CatWiz: Before this tip said that drills give 50% stone and destroy gold. Even if mat yield for stone will be changed (which i doubt because most players seem fine with how drills currently work), we can always update the tip.

PR #855, opened 2020-06-21
[added] sponges dry over time

### LINKED ISSUE #789: Make sponges re-usable after a certain duration of time
Currently if you get stunned twice by a water arrow or bomb, the sponge dies. If you use a sponge to absorb water blocks, there's also a limit before the sponge becomes full.

The idea is that after a certain amount of time, the sponge could become re-usable.  This way you could put used sponges in storage to reuse them later.

Could be 2-3 minutes.

Nananas:
## Status

## Description

Sponges dry out over time. Closes #789. 
Current timer is full dry (100 quantities absorbed) over 200 seconds.


Entities/Items/Sponge/SpongeDryInAir.as
```diff
@@ -0,0 +1,26 @@
+#include "SpongeCommon.as";
+
+const int slow_frequency = 5; // don't run this script too often
+
+void onInit(CBlob@ this)
+{
+	this.getCurrentScript().tickFrequency = 30 * 2 * slow_frequency; // 0.5 "absorbs" per seconds, run only every 5 "absorbs"
```
Harrison-Miller, line 7: you can use the script flags tick_not_inwater and tick_ininventory (instead of doTickScripts).

Entities/Items/Sponge/SpongeDryInAir.as
```diff
@@ -0,0 +1,26 @@
...
+}
+
+void onTick(CBlob@ this)
+{
+	u8 absorbed = this.get_u8(ABSORBED_PROP);
```
Harrison-Miller, line 12: It would be nice if we still did accelerate this when it's on fire or on/near a block on fire.

Entities/Items/Sponge/SpongeCommon.as
```diff
@@ -0,0 +1,9 @@
+
+const int ABSORB_COUNT = 100;
+const string ABSORBED_PROP = "absorbed";
+
+
+void spongeUpdateSprite(CSprite@ this, u8 absorbed)
+{
+	this.animation.setFrameFromRatio(f32(absorbed) / ABSORB_COUNT);
```
Harrison-Miller, line 8: Something that would be nice is if the current frame is greater than the new frame (i.e the sponge has dryed by one frame) Is to make some water particles like the water is coming out of it.

Entities/Items/Sponge/SpongeDryInAir.as
```diff
@@ -0,0 +1,26 @@
+#include "SpongeCommon.as";
```
Harrison-Miller, line 1: I believe this script can be #define SERVER_ONLY

Entities/Items/Sponge/SpongeCommon.as
Nananas: Should be in latest commit

Entities/Items/Sponge/SpongeDryInAir.as
Nananas: Do you mean accelerate drying when sponge is in inventory too?

Entities/Items/Sponge/SpongeDryInAir.as
Nananas: tick_ininventory does not seem to work for me. Is also not used anywhere...

Issue #606, opened 2019-06-12
placed seeds won't collapse


mugg91:
You can place seeds on (or above, next to or below) background for some reason.

When the background (or whatever the seed is placed on) collapses, the seeds will not collapse but remain in the air.

Harrison-Miller:
Can you provide a video?

AsuMagic:
It's fairly easy to reproduce: 
![placing on wooden backwall](https://user-images.githubusercontent.com/8986152/74220267-6a9ec800-4caf-11ea-958e-c0f04bef66f7.png)
![collapsing wooden backwall](https://user-images.githubusercontent.com/8986152/74220278-71c5d600-4caf-11ea-8c46-b8307209ce59.png)
Looks like it should check whether there is a dirt background rather than just a regular background. I'm not sure where this is handled but make sure you don't break placing stuff on backwalls (e.g. spikes) when intended.

Panky123:
If you leave the seeds floating too long they also despawn. Really old bug that would be a pleasant surprise to see fixed.

eps0003:
Fixed in #856

PR #856, opened 2020-06-22
[modified] prevent placement of seeds on non-growing tiles. Seeds fall down when losing supporting tile

Nananas:
## Status

## Description

Placement of seeds now only allowed on tiles where seeds can actually grow.
Seeds should have improved physics, which means they will fall down when their support is lost.
This should also fix issue #606.

## Steps to Test or Reproduce

Try and fail planting seeds on stone, wood, platforms, shop etc.
Plant a seed on ground tile (only place you can put them now), then dig out that tile and the seed will drop down (with a slight delay).

### DIFF
--- a/Entities/Common/Building/PlacementCommon.as
+++ b/Entities/Common/Building/PlacementCommon.as
@@ -125,15 +125,16 @@ bool isBuildableAtPos(CBlob@ this, Vec2f p, TileType buildTile, CBlob @blob, boo
 	{
 		bool isLadder = false;
 		bool isSpikes = false;
+		bool isSeed = false;
 		if (blob !is null)
 		{
 			const string bname = blob.getName();
 			isLadder = bname == "ladder";
 			isSpikes = bname == "spikes";
+			isSeed = bname == "seed";
 		}
 
 		Vec2f middle = p;
-
 		if (!isLadder && (buildSolid || isSpikes) && map.getSectorAtPosition(middle, "no build") !is null)
 		{
 			return false;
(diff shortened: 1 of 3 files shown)


Entities/Common/Building/PlacementCommon.as
```diff
@@ -182,6 +183,14 @@ bool isBuildableAtPos(CBlob@ this, Vec2f p, TileType buildTile, CBlob @blob, boo
 				}
 			}
 		}
+
+
+		if (isSeed)
+		{
+			// from canGrow.as
+			return (map.isTileGroundStuff(map.getTile(p + Vec2f(0, 8)).type));
```
Harrison-Miller, line 191: This includes dirt back wall. I would narrow it down so you can only place it on dirt.

Entities/Natural/Seed/Seed.as
```diff
@@ -136,4 +136,25 @@ void onTick(CBlob@ this)
 			}*/
 		}
 	}
+
+	if (isServer())
```
Harrison-Miller, line 140: checkout FallOnNoSupport.as. You might want to look into using that script. Or using onBlobCollapse hook.

Entities/Natural/Seed/Seed.as
```diff
@@ -136,4 +136,25 @@ void onTick(CBlob@ this)
...
+			// drop down when not supported
+			CShape@ shape = this.getShape();
+			shape.server_SetActive(true);
+
+			shape.SetStatic(false);
+		}
+	}
+}
+
+bool doesCollideWithBlob(CBlob@ this, CBlob@ blob)
```
Harrison-Miller, line 157: Good catch.

Entities/Common/Building/PlacementCommon.as
Nananas: Should be fixed in latest commit

Harrison-Miller:
gonna call this good, probably don't need to worry about how the falling logic is done.

PR #859, opened 2020-06-25
[modified] add flag (fast) return indicators, improved progress bar

### LINKED ISSUE #782: Flag return timer freezes 1-2 seconds before flag returns
1. Join CTF
2. Pick up the enemy flag
3. Drop it on the ground and wait for it to return
4. A few seconds before flag returns, bar that shows how much time is left before flag returns will freeze

Nananas:
## Status

## Description

Added tick and fast-tick indicators to flag progress bar. Moved "!!" icon up to make space for these.
Changed way flag progress bar is drawn to allow red bar at the end.

See [discord](https://discord.com/channels/337250580303052801/337266863841214465/725752161467629651) for video

Should fix #782 too.


Entities/Special/CTF/CTF_Flag.cfg
```diff
@@ -5,7 +5,7 @@
 $sprite_factory                                   = generic_sprite
 @$sprite_scripts                                  = CTF_Flag.as;
 $sprite_texture                                   = CTF_Flag.png
-s32_sprite_frame_width                            = 8
```
Harrison-Miller, line 8: Why did you change this and the frame number below? Frame sizes should be powers of 2

Entities/Special/CTF/CTF_Flag.as
```diff
@@ -135,18 +140,37 @@ void onRender(CSprite@ this)
...
+				GUI::DrawRectangle(pos + Vec2f(6, 6), bar + Vec2f(2, 4), SColor(255, 99, 112, 95));
+				GUI::DrawRectangle(pos + Vec2f(6, 6), bar + Vec2f(2, 2), SColor(255, 125, 139, 120));
+			}
 
 			if (getGameTime() % 15 > 10)
 			{
 				if (!shouldFastReturn(blob))
 				{
-					GUI::DrawIconByName("$ALERT$", Vec2f(pos2d.x - 32.0f, pos2d.y - 30.0f + wave));
+					GUI::DrawIconByName("$ALERT$", Vec2f(pos2d.x - 32.0f, pos2d.y - 80.0f + wave));
```
Harrison-Miller, line 168: Let's either have the alert always flashing, or just removed entirely.

Entities/Special/CTF/CTF_Flag.as
```diff
@@ -135,18 +140,37 @@ void onRender(CSprite@ this)
 
 		if (returncount < return_time)
 		{
-			GUI::DrawProgressBar(Vec2f(pos2d.x - 30.0f, pos2d.y - 10.0f + wave), Vec2f(pos2d.x + 30.0f, pos2d.y + 5.0f + wave), 1.0f - float(returncount) / float(return_time));
+			Vec2f pos = pos2d + Vec2f(-30.0f, -20.0f + wave);
```
Harrison-Miller, line 143: When you throw the flag the bar is really stuttery (similar problem with the drill heat bar). I think you should be using the interpolated position.

Entities/Special/CTF/CTF_Flag.cfg
Nananas: Why is this? I modified the width of the flag sprite to align better with the box2d shape. Made it easy to draw the progress bar in the middle.

Harrison-Miller: non-power of 2 can cause rendering issues on older graphics drivers/cards. we can modify the sprite_offset_x or move the flag 1 pixel over in the spritesheet, so it's centered in it's frame. Those would be alternatives.

PR #863, opened 2020-06-27
[fixed] destroying backwall tiles at top of map

### LINKED ISSUE #593: Background at the top of the world cannot be destroyed
(This is not a bug per se so I'm submitting this as a change request)

At the top of the world, you cannot build solid blocks past a certain point, but you can place background 1 tile into that red area. You cannot remove the background once it is placed (except by collapse).

It should either not be possible to place background in that top row OR it should be possible to remove it.

Nananas:
## Status

## Description

Fixes #593.

Caused by inconsistency between placing and destroying tiles as builder

### DIFF
--- a/Entities/Characters/Builder/BuilderLogic.as
+++ b/Entities/Characters/Builder/BuilderLogic.as
@@ -141,13 +141,13 @@ bool RecdHitCommand(CBlob@ this, CBitStream@ params)
 
 	if (blobID == 0)
 	{
+		// block
 		CMap@ map = getMap();
 		if (map !is null)
 		{
-			if (map.getSectorAtPosition(tilepos, "no build") is null)
+			uint16 type = map.getTile(tilepos).type;
+			if (!inNoBuildZone(map, tilepos, type))
 			{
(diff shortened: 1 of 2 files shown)


Entities/Characters/Builder/BuilderLogic.as
```diff
@@ -326,7 +327,8 @@ void Pickaxe(CBlob@ this)
 	@bestinfo = hit_p.bestinfo;
 	bestDistance = hit_p.bestDistance;
 
-	bool noBuildZone = map.getSectorAtPosition(tilepos, "no build") !is null;
+	Tile tile = map.getTile(tilepos);
+	bool noBuildZone = inNoBuildZone(map, tilepos, tile.type);
```
eps0003, line 331: What's this doing? Looks unnecessary.

Nananas: Do you mean the `noBuildZone` variable? It is used a couple lines below, old code.

eps0003: Oh right. I didn't see that.

PR #865, opened 2020-06-28
Archer: some fixes for no-arrows cycling

### LINKED ISSUE #576: Revert auto-swap nerf for archers

Since a few months ago the archer special arrow swapping doesn't work as it used to, it's fairly awkward in comparision. If you are out of regular arrows and buy a special one (fire/water/bomb), the arrow type doesn't automatically change to the only arrow you are capable of shooting, and you have to do it manually, not by tapping F but by opening the arrow menu. 

**Proposed Change**

I'd like the behaviour to be reverted to the previous system, where whenever you only had 1 type of arrow the type selected would automatically become that one or it would fastly change by tapping F. 

**Additional context**

Nananas:
## Status

## Description

Fixes two issues: 
1. Cycling from no-arrow type to has-arrow type quickly (!) now actually cycles. Fixes #576.
2. While charging, cycling from no-arrow type to has-arrow type will cycle and start charging.

## Steps to Test or Reproduce

- as archer get normal arrows and one special arrow type
- shoot a normal arrow
- drop the normal arrows in inventory, so you don't have any more normal ones
- try shooting a normal arrow and notice the empty-bow draw animation
- then either:
  - (1) start charging the empty bow, cycle to the special arrow. Notice it starts charging without having to cancel first.
  - (2) in quick succession start charging the empty bow, stop charging, cycle, start charging again. Notice the special arrow is now charging. (this is the timing issue)

### DIFF
--- a/Entities/Characters/Archer/ArcherLogic.as
+++ b/Entities/Characters/Archer/ArcherLogic.as
@@ -963,7 +963,7 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
 		}
 		if (archer.charge_state == ArcherParams::no_arrows)
 		{
-			archer.charge_state = ArcherParams::readying;
+			archer.charge_state = ArcherParams::not_aiming;
 		}
 
 	}
--- a/Entities/Characters/Archer/ArcherLogic.as
+++ b/Entities/Characters/Archer/ArcherLogic.as
@@ -492,7 +492,8 @@ void ManageBow(CBlob@ this, ArcherInfo@ archer, RunnerMoveVars@ moveVars)
 			{
 				charge_time--;
 
-				if (charge_time <= 0)
+				// when no arrows, ignore charge timer
+				if (charge_time <= 0 || !hasarrow)
 				{
 					charge_state = ArcherParams::not_aiming;
 					charge_time = 0;
--- a/Entities/Characters/Archer/ArcherLogic.as
+++ b/Entities/Characters/Archer/ArcherLogic.as
@@ -961,6 +961,11 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
 				break;
 			}
 		}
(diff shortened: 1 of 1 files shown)


eps0003 (changes requested):
Although this fixes the issues stated, you have hardcoded the fix into the code to cycle arrows. If a new feature were to change the arrow type through another means (for example #909), the issues would re-emerge.

Another issue with your fix is that it doesn't play the bow pull sound after you switch using the steps you provided.

I suggest you add the following code after [this line](https://github.com/transhumandesign/kag-base/blob/5e1c8d9516da86dbf18edc09f22ebce238feb1f1/Entities/Characters/Archer/ArcherLogic.as#L477) instead of your code to resolve both issues stated above:
```angelscript
hasarrow = hasArrows(this);

if (hasarrow)
{
	charge_state = ArcherParams::not_aiming;
	charge_time = 0;
}
```

Nananas:
Your proposed change did not fix the sound issue on my machine :thinking:.
I tried fixing it myself with the latest commit. It should now also play the sound (good catch btw).
Could you check if this fixes the "hardcoded" issue too? I don't really understand that part, sorry.

eps0003:
I'm yet to actually test this but I can already see that you haven't fixed the "hardcoded" issue ("hardcoded" probably wasn't the correct word to describe it). What I mean by that is you've placed your fix in the "cycle" command which is only called when tapping F.

When I said that the issue would reemerge with my keybinds PR or any other feature which manually changes arrow type, I'm referring to how the keybinds don't use the "cycle" command but instead go through a [new "switch" command](https://github.com/transhumandesign/kag-base/pull/909/files#diff-27d80d8112737cb45e79e3876f03da92R960).

You need to fix the issue directly within the archer's ManageBow method so it fixes the issue regardless of how the arrow type is changed.

PR #880, opened 2020-07-26
Drill, water bucket changes

CatWiz:
## Status

## Description

- Water splashes now cool the drill down by 70% of max heat (was 100%)
- Heat up rate against natural tiles increased from 4 to 5
- Delay between hits when drilling dirt, stone or thick stone increased from 8 to 10
- Water buckets have 2 splashes instead of 3
- Drill cooldown delay increased from 5 to 6, making it cool down slower
- Fixed drill overheating on mat blobs that arent stone, wood or gold

Main goal of these changes is to make drill less efficient at quickly digging long rat tunnels.


eps0003 (changes requested):
Please update the original post with the most up to date values, specifically the following:
* "Water splashes now cool the drill down by ***70%*** of max heat (was 100%)"
* "Drill cooldown delay increased from 5 to 6, making it cool down slower" (I don't see this in your code)

In terms of whether these are good changes is for up to the community to decide. Please encourage more people provide their feedback.

Entities/Industry/Drill/Drill.as
```diff
@@ -273,7 +279,7 @@ void onTick(CBlob@ this)
 								// or blobs to increase damage to (for the future)
 								string name = b.getName();
 
-								if (name == "mat_stone" || name == "mat_wood" || name == "mat_gold")
+								if (b.hasTag("material") || b.hasTag("invincible") || b.hasScript("Eatable.as") || name == "tent")
```
eps0003, line 282: All materials, food, and the tent are given the "invincible" tag in IgnoreDamage.as (except for a few exceptions for items that are rarely/never used), so having the extra checks for tags is redundant.

PR #883, opened 2020-08-11
BuilderAutoPickup.as bugfix

Hobey:

BuilderAutoPickup.as used an incorrect expression for SetDamageOwnerPlayer on line 88: blob.getPlayer() instead this.getPlayer(); "this" is the builder, and "blob" is the mat_wood/mat_stone or whatever; that means blob.getPlayer() is always null.

Changing this allows builders to immediately autopickup mats that other people drop when they switch off builder (after building time ends, if they resupply for you at tent every time they respawn, etc.) by simply walking over the mats.

The one downside of this is that because the bug that makes people drop their 300 wood + 100 stone sometimes on nextmap still exists in vanilla (it's fixed on captains), this change might make picking up your own mats in that scenario impossible; that bug should just be fixed in vanilla as well.

### DIFF
--- a/Entities/Characters/Builder/BuilderAutoPickup.as
+++ b/Entities/Characters/Builder/BuilderAutoPickup.as
@@ -17,7 +17,7 @@ void Take(CBlob@ this, CBlob@ blob)
 		blobName == "mat_stone" ||
 		blobName == "mat_wood"
 	) {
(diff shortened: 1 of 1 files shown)


Hobey:
I had to close and reopen this pull request, because I derped and forgot to make it a branch after forking (I used Hobey:master)

PR #884, opened 2020-08-11
Changed minimap colors of blocks/backwalls; added minimap icons for trees, ballista, tunnel.

Hobey:

The initial goal for these changes was to make stone ore and bedrock distinguishable from dirt; later decided to add icons for tree/tunnel/ballista.

### Example Screenshots
screen-20-08-11-10-23-46

screen-20-08-10-22-00-03

screen-20-07-16-01-31-05
_(the third screenshot is from an early version without tunnel icon, and a fat version of tree icons)_

### Changed the block colors on the minimap of:

Dirt
Bedrock
Stone ore
Rich stone ore
Gold ore

Wood block
Stone block

Dirt backwall
Wood backwall
Stone backwall
Sky/empty tiles

(Sidenote: the minimap block colors now exactly match the colors that the map pngs use to represent tiletypes)

### Added minimap icons for:
Pine tree
Oak tree
Tunnel

Modified the ballista minimap icon to look like a ballista instead of a flag.

Note: decided to set tunnel and ballista icons to "not always visible", meaning if the tunnel/ballista is underground (or the area is darkened with ceiling and filled up background) and you are on the enemy team, you will not be able to see the tunnel/ballista, unless there is a teammate or lantern near the tunnel/ballista (the blue/red player dots actually work the same way; you can't see a rat in a dark area).
Possibly trees also should be set to "not always visible", but I like the idea of being able to find a lost sapling in a totally absurd mid clusterfuck.

### Example of "not always visible":

Blue underground tunnel+ballista invisible on the minimap to red:
screen-20-08-11-10-40-37

... but visible to blue:
screen-20-08-11-10-40-07

... and visible on the minimap as red with a lantern:
screen-20-08-11-10-40-43

... icons also appears on the minimap when standing near them as red:
screen-20-08-11-10-41-18


Scripts/MapLoaders/MinimapHook.as
```diff
@@ -1,96 +1,76 @@
...
+	// bool isTileBackground(const Tile&in tile)
+	// bool isTileBackgroundNonEmpty(const Tile&in tile)
+	// bool isTileGroundBack(uint16 tile)
+	// bool isTileGold(uint16 tile)
+	// bool isTileGrass(uint16 tile)
+	// bool isTileWood(uint16 tile)
+	// bool isTileCastle(uint16 tile)
+	// bool isTileSand(uint16 tile)
+	// bool isTileGroundStuff(uint16 tile)
+	// bool isTilePlatform(const Tile&in tile)
```
eps0003, line 48: You forgot to remove this

Scripts/MapLoaders/MinimapHook.as
```diff
@@ -1,96 +1,76 @@
...
+	// } else if(map.isTileBackground(ctile) && !map.isTileGrass(tile))
+	} else if(map.isTileBackgroundNonEmpty(ctile) && !map.isTileGrass(tile)) {
+		
+		// TODO(hobey): maybe check if there's a door/platform on this backwall and make a custom color for them?
+		if        (tile == CMap::tile_castle_back) {
+			col =       color_minimap_castle_back;
+		} else if (tile == CMap::tile_wood_back)   {
+			col =       color_minimap_wood_back;
+		} else {
+			col = color_minimap_back;
```
eps0003, line 67: This entire section needs to be formatted better:
* Put curly brackets on their own line
* Add a space after `if` on line 59
* Remove excess whitespace after `if` on line 62

eps0003 (changes requested):
* Change the tunnel icon outline colour to the correct colour (the same as the outline of the ballista icon).
* Why are the icons in separate files instead of in MinimapIcons.png? If the reason is that there isn't enough room, then just increase the height of the image so it doesn't ruin the indices of all the other minimap icons.
* Remove the extra, unneeded whitespace changes in MinimapHook.as (lines 9, 12, 14, and 73).
* Why did you create a new file to contain the ballista minimap code instead of putting it in Ballista.as? None of the other blobs visible on the minimap have this.

Harrison-Miller:
I may be willing to help fix up this PR, thoughts on gold being visible in the minimap? I think I'm in favor of it just showing as dirt.

CatWiz:
Why would it show as dirt? If we're updating the minimap to show more useful info, making it not show something important that current minimap map already shows would be counterintuitive and weird.

Hobey:
Did the requested changes.

The map vote minimaps now use the new colors

Example:
screen-20-12-23-11-54-31

Hobey:
I'm won't edit this branch; I'm exhausted.

PR #890, opened 2020-08-25
Double the frequency of resupply checks

benjamingorman:
## Status

## Description

This doubles the frequency of checks at shops and tent for whether resupplies should be given to a player. It should make resupplies feel a lot more snappy, and remove some cases where your character walks over a shop but does not receive a resupply.

## Steps to Test or Reproduce

Walk over a shop and notice that you get given your resupply mats faster than usual.

Affected:

- Tent
- Builder shop
- Knight shop (if it actually gave any mats)
- Archer shop


CatWiz:
You should also be able to get resupplies the same way at the ballista/warboat, without changing your class from builder to builder.

eps0003 (approved):

PR #891, opened 2020-08-25
[Change] Explosions now drop gold if it hits a gold tile

Vam-Jam:
Title, drops gold if it damages a gold tile instead of dropping nothing.


Hobey:
doesn't work with kegs because kegs have their tile hit logic in LinearExplosion()

also, I have noticed that explosions sometimes drop more than 20 gold per gold ore block (I've also noticed that with knight jab/slash before; dunno where that bug comes from)

Vam-Jam:
Good catch, ill look into it in a bit

Harrison-Miller:
knight gold logic: https://github.com/transhumandesign/kag-base/blob/master/Entities/Characters/Knight/KnightLogic.as#L1266

vs knight attacking logs logic (from test build):
```

					if (b.getName() == "log")
					{
						damage /= 3;
						dontHitMoreLogs = true;
						CBlob@ wood = server_CreateBlobNoInit("mat_wood");
						if (wood !is null)
						{
							int quantity = Maths::Ceil(float(damage) * 20.0f);
							int max_quantity = b.getHealth() / 0.024f; // initial log health / max mats
							quantity = Maths::Min(quantity, max_quantity);

							wood.Tag('custom quantity');
							wood.Init();
							wood.setPosition(hi.hitpos);
							wood.server_SetQuantity(quantity);
						}

					}

```

Harrison-Miller:
fix or ded

Vam-Jam:
Works with kegs now

``also, I have noticed that explosions sometimes drop more than 20 gold per gold ore block``

I noticed this, but it just turns out it's because the AOE explosion goes through tiles, so you could be hitting multiple other gold ore hidden under dirt.

Entities/Items/Explosives/Explosion.as
```diff
@@ -374,6 +380,12 @@ void LinearExplosion(CBlob@ this, Vec2f _direction, f32 length, const f32 width,
 						if (!justhurt)
 							damaged = true;
 
+						if (map.isTileGold(t))
+						{
+							Material::fromTile(this, t, 1.0f); // drops gold at bombs position
```
AsuMagic, line 385: should this also be `1.0`, if we deal `5.0` damage just afterwards?

AsuMagic:
Going to tag this as abandoned for now. If anyone wants to fix the remaining issues that'd be nice.

PR #892, opened 2020-08-25
[Change] Don't take vehicle in to the tunnel

Vam-Jam:
If we enter a tunnel in a vehicle, you teleport the whole vehicle with you. This disables that (as you normally cannot get the tunnel button while sat down anyhow).

Vehicle teleporting is still allowed outside CTF, for modes like sandbox where it's humorous to do so, but in CTF it will kick you out of your vehicle if you try it.

### DIFF
--- a/Entities/Industry/Tunnel/TunnelTravel.as
+++ b/Entities/Industry/Tunnel/TunnelTravel.as
@@ -1,4 +1,4 @@
-﻿// include file for blobs that use tunnel travel capabilities
+// include file for blobs that use tunnel travel capabilities
 // apply "travel tunnel" tag to use
 
 #include "TunnelCommon.as";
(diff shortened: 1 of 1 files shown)


Entities/Industry/Tunnel/TunnelTravel.as
```diff
@@ -139,7 +139,7 @@ void Travel(CBlob@ this, CBlob@ caller, CBlob@ tunnel)
 		//if (isKnockable(caller) && caller.get_u8("knocked") > 0)
 		//	return;
 
-		if (caller.isAttached())   // attached - like sitting in cata? move whole cata
+		if (caller.isAttached() && getRules().gamemode_name != "CTF")   // attached - like sitting in cata? move whole cata, only in none CTF
```
eps0003, line 142: I dislike hardcoding the gamemode. What about other gamemodes like TTH and SmallCTF?

eps0003:
The same issue is fixed in my pull request (#922) but it removes it for all gamemodes.

PR #893, opened 2020-08-26
Bombs dont collide with shielding enemy knights if they're shielding away from the bomb

CatWiz:
## Status

## Description

Title. Bombs only collide with enemy shields now


eps0003 (approved):
This is good, but I wonder if it would work if you allowed collision with shielding teammates.

CatWiz:
I'd rather not. People would accidentally block bomb throws or shield slide into teammates' bombs. Trolls would also use it to prevent bomb jumps and play it off as an accident.

PR #894, opened 2020-08-27
Mine ownership text on mouse hover

bunniewormy:
## Status

(pick one)

## Description

Whenever you hover your mouse on any mine belonging to your team, you will see "Owner: (mine's damage owner here)".

### DIFF
--- a/Entities/Items/Explosives/Mine/Mine.as
+++ b/Entities/Items/Explosives/Mine/Mine.as
@@ -108,7 +108,7 @@ void onRender(CSprite@ this)
 	bool mouseOnBlob = (mouseWorld - center).getLength() < renderRadius;
 	string mineOwner = "Owner: " + blob.get_string("owner");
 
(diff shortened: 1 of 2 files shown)


Vam-Jam:
image
Could you add an offset to the text, so its higher then the mine, as your cursor just blocks off the text.
Same thing happens when you hold C to pick up the mine, the text is under the blob and mouse, making it unreadable.

Could you also move ``void onRender(CSprite@ this)`` Under onTick

The reason for this is most files go (just so its more consistent) 

- onInit
- onTick
- onRender

The onTick sync isnt stuff needed. getDamageOwnerPlayer is already synced to everybody who's currently in the game.

However it may be needed because of this bug (still happens with your sync stuff however);

image
Any user who joins the game after the mine has an owner will show up like so (no owner), but for everybody else, it appears as the correct owner.

Otherwise good work

Vam-Jam (changes requested):
(last comment)

bunniewormy (commented):
i dont understand github (and now i dont even know how to remove this comment)

PR #895, opened 2020-08-27
Boulder

CatWiz:
## Status

## Description

- Boulders now break into 10 rocks ~on heavy impact or~ if launched from cata
- Collision with teammates disabled

Shattering on heavy impact is currently disabled because it gets really OP with archers


Entities/Items/Boulder/Boulder.as
```diff
@@ -239,6 +268,13 @@ f32 onHit(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@ hitt
 	return damage;
 }
 
+bool doesCollideWithBlob(CBlob@ this, CBlob@ blob)
+{
+	string name = blob.getName();
+	if ((name == "archer" || name == "knight" || name == "builder") && this.get_u8("launch team") == blob.getTeamNum()) return false;
```
benjamingorman, line 274: How about just make it not collide with any player blobs? This would work better with mods that have custom classes
```suggestion
	if (blob.getPlayer() !is null && this.get_u8("launch team") == blob.getTeamNum()) return false;
```

eps0003:
What's the benefit now of firing a boulder instead of just regular stones? Its usefulness is now reduced imo.

CatWiz:
> Its usefulness is now reduced imo.

That's kinda the point, although to me it feels like rocks from a boulder do much more damage, especially to doors/platforms. The slight benefit is that damage from boulder is spread around the point of impact instead of being concentrated in one spot. If you think it needs to be stronger, we could just increase amount of rocks/tweak their spread.

eps0003:
> The slight benefit is that damage from boulder is spread around the point of impact instead of being concentrated in one spot.

That's what regular rocks are for. Unlike regular rocks which are designed for killing players, boulders were designed to cut through structures (at least that's what it seems like). I don't think their primary use is to kill players. Your changes remove one of the few effective methods of destroying structures. Now the only other methods of destroying structures effectively (aside from builders) are explosive ballista bolts and kegs.

> [Reducing its usefulness] is kinda the point

> If you think it needs to be stronger, we could just increase amount of rocks/tweak their spread.

If your goal is to reduce its usefulness, I suggest you just reduce how far it can cut through blocks instead of removing its unique mechanic. I hate to admit this, but my [pull request](https://github.com/transhumandesign/kag-base/pull/422#pullrequestreview-207719510) might've been the cause of it being so strong. 😅

CatWiz:
> it feels like rocks from a boulder do much more damage, especially to doors/platforms

To better showcase what i mean:

Damage from 5 normal cata shots:
image

Damage from 5 cata boulders:
image

Damage from 5 cata boulders in vanilla (i tried shooting with slightly different power but all boulders magically home into 1 spot):
image

Harrison-Miller:
So I see your screenshot and that looks nice. But after testing it seems really inconsistent how much damage it does to a stone wall. I was build a 6x15 stone wall for testing tried boulders and most of the time it didn't even break a single tile. Additionally there seems to be a bug where the catapult will get into a state where every time it fires a boulder it immediately breaks and does nothing.

CatWiz:
Rocks that spawn from boulder do fly in completely random directions, so that's not surprising. Could make first few rocks fly in the direction of the boulder with low spread.

Turtlebutt:
Should probably test from a higher point(aswell), like a tower with an overhang, rather than on ground where it isn't realistic to have a cata.

PR #896, opened 2020-08-29
Food Ownership Changes

bunniewormy:
## Status

(pick one)

## Description

Food ownership now works similiarly to mines - the "owner" of a food item is the person who bought it or first picked it up. You also become owner of food items taken from enemies. When somebody in your team picks up your food items, you are still the owner. The point of this PR is that the coins for healing people are given to the person who buys the food, rather than the person that throws food at damaged teammates.

Moreover, you can also see owner of a food item when you hover your mouse over a food item. This works the same as my recent mine PR https://github.com/transhumandesign/kag-base/pull/894

### DIFF
--- a/Entities/Items/Food/Eatable.as
+++ b/Entities/Items/Food/Eatable.as
@@ -84,7 +84,7 @@ void onAttach(CBlob@ this, CBlob@ attached, AttachmentPoint @attachedPoint)
 
 	if (isServer())
 	{
-	Heal(attached, this);
+		Heal(attached, this);
 	}
 
 	setHealer(this, attached);
--- a/Entities/Items/Food/Eatable.as
+++ b/Entities/Items/Food/Eatable.as
@@ -80,14 +80,14 @@ void onCollision(CBlob@ this, CBlob@ blob, bool solid)
 
 void onAttach(CBlob@ this, CBlob@ attached, AttachmentPoint @attachedPoint)
 {
-    if (this is null || attached is null) {return;}
+	if (this is null || attached is null) {return;}
 
-    if (isServer())
-    {
-        Heal(attached, this);
-    }
-    
-    setHealer(this, attached);
+	if (isServer())
+	{
+	Heal(attached, this);
+	}
+
+	setHealer(this, attached);
 }
 
 void onThisAddToInventory(CBlob@ this, CBlob@ inventoryBlob)
--- a/Entities/Items/Food/Eatable.as
+++ b/Entities/Items/Food/Eatable.as
@@ -106,34 +106,12 @@ void onAttach(CBlob@ this, CBlob@ attached, AttachmentPoint @attachedPoint)
         Heal(attached, this);
     }
     
-    CPlayer@ player = attached.getPlayer();
-
-    if (player !is null && (this.getTeamNum() != attached.getTeamNum() || !this.exists("healer")))
-    {
-        if (isServer())
-        {
-            this.server_setTeamNum(attached.getTeamNum());
-        }
-
-        this.set_u16("healer", player.getNetworkID());
-        this.Sync("healer", true);
-    }
+    setHealer(this, attached)
 }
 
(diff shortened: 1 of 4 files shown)


Entities/Items/Food/Eatable.as
```diff
@@ -78,20 +97,43 @@ void onCollision(CBlob@ this, CBlob@ blob, bool solid)
...
+        this.Sync("healer", true);
+    }
+}
 
-	this.set_u16("healer", p.getNetworkID());
+void onThisAddToInventory(CBlob@ this, CBlob@ inventoryBlob)
+{
+    CPlayer@ player = inventoryBlob.getPlayer();
+
+    if (player !is null && (this.getTeamNum() != inventoryBlob.getTeamNum() || !this.exists("healer")))
```
AsuMagic, line 127: That `if` condition logic is duplicated with the `onAttach` function, make a function for this, e.g. `void setHealer(CBlob@ food_blob, CPlayer@ player)` so you can call it like `setHealer(@this, @player);`

Entities/Items/Food/Eatable.as
```diff
@@ -10,6 +10,25 @@ void onInit(CBlob@ this)
 	this.addCommandID(heal_id);
 }
 
+void onTick(CBlob@ this)
```
AsuMagic, line 13: The whole hook seems unnecessary as you could get the username in the `onRender` hook instead, without having to use a separate property for the player name and network ID.

Entities/Items/Food/Eatable.as
```diff
@@ -78,20 +97,43 @@ void onCollision(CBlob@ this, CBlob@ blob, bool solid)
...
+
+    if (player !is null && (this.getTeamNum() != inventoryBlob.getTeamNum() || !this.exists("healer")))
+    {
+        if (isServer())
+        {
+            this.server_setTeamNum(inventoryBlob.getTeamNum());
+        }
+
+        this.set_u16("healer", player.getNetworkID());
+        this.Sync("healer", true);
```
AsuMagic, line 135: I don't think the `Sync` is necessary, as I believe this hook should get called both on client and server side already. Should check whether removing that line breaks the behavior on dedicated servers.

Entities/Items/Food/Food.as
```diff
@@ -15,7 +15,26 @@ void onInit(CBlob@ this)
 
 	this.getSprite().SetFrameIndex(index);
 	this.SetInventoryIcon(this.getSprite().getConsts().filename, index, Vec2f(16, 16));
-	this.server_setTeamNum(0); // blue fishy like in sprite sheet
+	// this.server_setTeamNum(0); // blue fishy like in sprite sheet
```
AsuMagic, line 18: Not sure why this is commented out now, should probably clarify why in a comment if intended

AsuMagic (changes requested):

Entities/Items/Food/Eatable.as
bunniewormy: should be fine now

Entities/Items/Food/Eatable.as
bunniewormy: added setHealer to EatCommon.as

Entities/Items/Food/Eatable.as
bunniewormy: should be fine now

Entities/Items/Food/Food.as
```diff
@@ -15,7 +15,44 @@ void onInit(CBlob@ this)
...
+	{
+		u16 healerID = blob.get_u16("healer");
+		CPlayer@ healer = getPlayerByNetworkId(healerID);
+		if(healer !is null)
+		{
+			healerName = healer.getUsername(); 
+		}
+	}
+
+	blob.Sync("healer", true);
```
AsuMagic, line 49: That should not be here

bunniewormy: fixed

AsuMagic (dismissed):
LGTM

Issue #835, opened 2020-06-13
Lock keg ownership for 2 seconds before explosion


bunniewormy:
**Current in-game behaviour.**

Whenever you pick up a keg lit by anybody else, you gain ownership on it. E.g. red knight puts a lit keg next to 4 blue people but in the last second one of the blue people pick up the keg -- the red person won't get any coins or kills.

**Proposed Change**

If any person picks up a keg in the last 2 seconds of its fuse, ownership doesn't change -- previous owner still has the ownership.

PR #897, opened 2020-08-29
Lock keg ownership for 2 seconds before explosion

bunniewormy:
## Status

## Description

Resolves https://github.com/transhumandesign/kag-base/issues/835

### DIFF
--- a/Entities/Items/Explosives/Keg.cfg
+++ b/Entities/Items/Explosives/Keg.cfg
@@ -74,7 +74,6 @@ $name                                             = keg
 													NoPlayerCollision.as;
 													BehindWhenAttached.as;
 													SetTeamToCarrier.as;
-													SetDamageToCarrier.as;
 f32 health                                        = 3.0
 $inventory_name                                   = Keg
 $inventory_icon                                   = -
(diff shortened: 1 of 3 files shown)


AsuMagic (commented):
Should work like that, but I think it would be better if this was implemented in keg scripts instead. i.e. copy over `onAttach` to keg scripts and remove `SetDamageToCarrier.as` from the script list in the keg config file.

Entities/Items/Explosives/Keg.cfg
```diff
@@ -73,8 +73,7 @@ $name                                             = keg
 													ExplodeOnDie.as;
 													NoPlayerCollision.as;
 													BehindWhenAttached.as;
-													SetTeamToCarrier.as;
-													SetDamageToCarrier.as;
+													SetTeamToCarrier.as
```
eps0003, line 76: I may be wrong, but doesn't there need to be a semicolon after each script?

eps0003: I am wrong.

bunniewormy: not needed, but changed for consistency

Issue #628, opened 2019-06-14
Display a coloured message in chat when someone gets a +5 or more


bunniewormy:
**Describe the feature**

Title, makes these kills more satisfying, something like 'Homek got a Pentakill!' in a special colour in chat.

benjamingorman:
Where do you think the message should appear? As part of the kill feed? Or above the player? Or both?

bunniewormy:
did you read the title?

benjamingorman:
Of course. The chat box is not necessarily the best place though. It's perhaps not as visible as text in game - or as obviously linked to the player in question.

PR #899, opened 2020-09-02
Display a coloured message in chat whenever somebody does a killstreak

bunniewormy:
## Status

(pick one)

## Description

Resolves https://github.com/transhumandesign/kag-base/issues/628

Displays message in chat whenever somebody gets a killstreak. I'm not sure if my implementation is good, or if it's bad for performance because I'm still new to modding/programming. The way it currently works is increasing the kill amount for killer blob in onPlayerDie, while onTick displays the message after the killstreak is over and resets the kill amount if 6 seconds have passed.

### DIFF
--- a/Rules/CommonScripts/KillMessages.as
+++ b/Rules/CommonScripts/KillMessages.as
@@ -288,9 +288,32 @@ void onPlayerDie(CRules@ this, CPlayer@ victim, CPlayer@ killer, u8 customdata)
 			CBlob@ victimblob = victim.getBlob();
 			CPlayer@ helper = getAssistPlayer(victim, killer);
 
-			bool kill = killerblob !is null && victimblob !is null && killerblob.isMyPlayer() && killerblob !is victimblob;
+			bool kill = killerblob !is null && victimblob !is null && killerblob !is victimblob;
 			bool assist = helper !is null && helper.isMyPlayer();
+
 			if (kill)
+			{
+				if(!killerblob.exists("kill time"))
+				{
+					killerblob.set_u32("kill time", getGameTime());
+				}
+
+				if(!killerblob.exists("killstreak"))
+				{
+					killerblob.set_u8("killstreak", 1);
+				}
+				else
+				{
+					killerblob.set_u8("killstreak", killerblob.get_u8("killstreak")+1);
+				}
+
+				killerblob.set_u32("kill time", getGameTime());
+
+				killerblob.Sync("kill time", true);
+				killerblob.Sync("killstreak", true);
+			}
+
+			if(kill && killerblob.isMyPlayer())
 			{
 				add_message(KillSpreeMessage(victim));
 			}
@@ -310,6 +333,47 @@ void onTick(CRules@ this)
 	{
 		feed.Update();
 	}
+
+	// display message in chat whenever a killstreak happens
+	for(int a = 0; a < getPlayerCount(); a++) 
+	{ 
+		CPlayer@ player = getPlayer(a);
+		CBlob@ blob = player.getBlob();
+
+		if (blob !is null && player !is null && blob.exists("killstreak"))
+		{
+			if(getGameTime() - blob.get_u32("kill time") > 180 && blob.get_u8("killstreak") > 4)
+			{
+				string multiKill;
+
+				switch (blob.get_u8("killstreak"))
+				{
+					case 5: multiKill = "a Pentakill";
+						break;
+					case 6: multiKill = "a Hexakill";
+						break;
+					case 7: multiKill = "a Septakill";
+						break;
+					case 8: multiKill = "an Octakill";
+						break;
+					case 9: multiKill = "a Nonakill";
+						break;
+					case 10: multiKill = "a Decakill";
+						break;
+					case 11: multiKill = "an " + blob.get_u8("killstreak") + " kill multikill";
+						break;
+					case 18: multiKill = "an " + blob.get_u8("killstreak") + " kill multikill";
+						break;
+					default: multiKill = "a " + blob.get_u8("killstreak") + " kill multikill";
+						break;
+				}
+
+				client_AddToChat(player.getCharacterName() + " got " + multiKill + "!", SColor(255, 180, 24, 94));
+					
+				blob.set_u8("killstreak", 0);
+			}
+		}
+	} 
 }
 
 void onRender(CRules@ this)


Rules/CommonScripts/KillMessages.as
```diff
@@ -310,6 +333,47 @@ void onTick(CRules@ this)
...
+	for(int a = 0; a < getPlayerCount(); a++) 
+	{ 
+		CPlayer@ player = getPlayer(a);
+		CBlob@ blob = player.getBlob();
+
+		if (blob !is null && player !is null && blob.exists("killstreak"))
+		{
+			if(getGameTime() - blob.get_u32("kill time") > 180 && blob.get_u8("killstreak") > 4)
+			{
+				string multiKill;
```
AsuMagic, line 347: this chat stuff should be moved in a `isClient()` branch. it won't behave any differently really, but it makes the intent more clear (and avoids some minor unnecessary processing on the server side)

Rules/CommonScripts/KillMessages.as
```diff
@@ -310,6 +333,47 @@ void onTick(CRules@ this)
...
+						break;
+					case 18: multiKill = "an " + blob.get_u8("killstreak") + " kill multikill";
+						break;
+					default: multiKill = "a " + blob.get_u8("killstreak") + " kill multikill";
+						break;
+				}
+
+				client_AddToChat(player.getCharacterName() + " got " + multiKill + "!", SColor(255, 180, 24, 94));
+					
+				blob.set_u8("killstreak", 0);
```
AsuMagic, line 373: this should be moved in a `isServer()` branch, and it should be `Sync`'d like it is in `onPlayerDie`

Rules/CommonScripts/KillMessages.as
```diff
@@ -288,9 +288,32 @@ void onPlayerDie(CRules@ this, CPlayer@ victim, CPlayer@ killer, u8 customdata)
...
 
-			bool kill = killerblob !is null && victimblob !is null && killerblob.isMyPlayer() && killerblob !is victimblob;
+			bool kill = killerblob !is null && victimblob !is null && killerblob !is victimblob;
 			bool assist = helper !is null && helper.isMyPlayer();
+
 			if (kill)
+			{
+				if(!killerblob.exists("kill time"))
+				{
+					killerblob.set_u32("kill time", getGameTime());
```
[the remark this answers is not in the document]
AsuMagic: looks to me this whole branch can be removed since the `exists` check makes no sense anyway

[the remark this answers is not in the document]
bunniewormy: this one made me laugh at how stupid i am

AsuMagic:
right so actually those syncing changes cause issues, so i thought about it a bit:

i think the most reliable and clean way to implement this would actually be
- only set the "killstreak", "kill time" properties on the server side, do not sync
- make the for loop server-side only and make it send a command that contains the player netid and kill count to the client

the reasons being
- property syncing is annoying to reason about and doesn't work well there so i'm dumb
- server_AddToChat would work, but it doesn't support colors because it's crap,
- and it would make it impossible to do things like implementing translations, or allowing to toggle the printing of those messages,
- and it saves a bit of bandwidth to do it the command way

Rules/CommonScripts/KillMessages.as
```diff
@@ -310,6 +331,54 @@ void onTick(CRules@ this)
...
+
+	// display message in chat whenever a killstreak happens
+	for(int a = 0; a < getPlayerCount(); a++) 
+	{ 
+		CPlayer@ player = getPlayer(a);
+		CBlob@ blob = player.getBlob();
+
+		if (blob !is null && blob.exists("killstreak"))
+		{
+			if(getGameTime() - blob.get_u32("kill time") > 180 && blob.get_u8("killstreak") > 4)
```
AsuMagic, line 343: replacing the `180` constant with `30 * 6` should be more clear

AsuMagic:
> right so actually those syncing changes cause issues, so i thought about it a bit:
> (snip)

oh, there is another issue i didn't notice at the time: if the player doing the killstreak dies it's possible it will not be shown because the blob may be deleted before the message gets shown.  
(the blob gets actually deleted several seconds after its death was requested, actually. note that when you die but your corpse is still there, it's still the same blob)

so the proper way would be to use cplayer properties instead of cblob ones since the player may outlive their blob. cplayer properties are horribly broken when it comes to syncing AFAIK, but the command based solution does not require syncing

the command stuff could be implemented like so:
- a rules script that sets up the command and contains the `onCommand` hook, receiving the payload containing the player netid and kill count, and displaying the message locally using `client_AddToChat` (make sure to use `saferead_xx` methods instead of `read_xx` ones since those don't do proper sanity checking because they're garbage)
- the `for` loop that checks for the killstreak would be executed only if `isServer()` and would send the command

bunniewormy:
killstreaks now increase coin gain for kills too

Rules/CommonScripts/KillMessages.as
```diff
@@ -324,3 +399,101 @@ void onRender(CRules@ this)
...
+				case 8: multiKill = "Octakill";
+					break;
+				case 9: multiKill = "Nonakill";
+					break;
+				case 10: multiKill = "Decakill";
+					break;
+				case 11: multiKill = kill_count + " kill multikill";
+					break;
+				case 18: multiKill = kill_count + " kill multikill";
+					break;
```
eps0003, line 491: These two cases can be removed since they're identical to the default case.

Rules/CommonScripts/KillMessages.as
```diff
@@ -324,3 +399,101 @@ void onRender(CRules@ this)
...
+				case 8: multiKill = "an Octakill";
+					break;
+				case 9: multiKill = "a Nonakill";
+					break;
+				case 10: multiKill = "a Decakill";
+					break;
+				case 11: multiKill = "an " + kill_count + " kill multikill";
+					break;
+				case 18: multiKill = "an " + kill_count + " kill multikill";
+					break;
```
eps0003, line 444:
```suggestion
				case 11:
				case 18: multiKill = "an " + kill_count + " kill multikill";
					break;
```

Rules/CommonScripts/KillMessages.as
```diff
@@ -310,6 +350,41 @@ void onTick(CRules@ this)
...
+			CPlayer@ player = getPlayer(a);
+
+			if (player !is null && player.exists("killstreak"))
+			{
+				if (getGameTime() < player.get_u32("kill time"))
+				{
+					player.set_u8("killstreak", 0);
+				}
+
+				if (getGameTime() - player.get_u32("kill time") > (6 * 30) && player.get_u8("killstreak") > 4)
```
eps0003, line 367: This is very nitpicky but I may as well suggest it while I'm here.
```suggestion
				if (getGameTime() - player.get_u32("kill time") > (6 * getTicksASecond()) && player.get_u8("killstreak") > 4)
```

Rules/CommonScripts/KillMessages.as
```diff
@@ -288,11 +291,48 @@ void onPlayerDie(CRules@ this, CPlayer@ victim, CPlayer@ killer, u8 customdata)
...
+					{
+						killer.set_u32("kill time", getGameTime());
+						
+						if(!killer.exists("killstreak"))
+						{
+							killer.set_u8("killstreak", 1);
+						}
+						else
+						{
+							killer.add_u8("killstreak", 1);
```
eps0003, line 311: I'm pretty sure `add_u8` adds from zero if `"killstreak"` doesn't exist. If this is the case, the surrounding `if else` statement becomes redundant. I may be wrong though.

Rules/CTF/Scripts/CTF_Trading.as
```diff
@@ -100,7 +101,7 @@ void onPlayerDie(CRules@ this, CPlayer@ victim, CPlayer@ killer, u8 customData)
 		{
 			if (killer !is victim && killer.getTeamNum() != victim.getTeamNum())
 			{
-				killer.server_setCoins(killer.getCoins() + coinsOnKillAdd);
+				killer.server_setCoins(killer.getCoins() + coinsOnKillAdd * (1+ (killer.get_u8("killstreak") - 1)*killstreakFactor));
```
eps0003, line 104: This can be simplified.
```suggestion
				killer.server_setCoins(killer.getCoins() + coinsOnKillAdd * killer.get_u8("killstreak") * killstreakFactor);
```

Rules/CommonScripts/KillMessages.as
eps0003: This also applies to line 381.

Rules/CommonScripts/KillMessages.as
```diff
@@ -288,11 +291,48 @@ void onPlayerDie(CRules@ this, CPlayer@ victim, CPlayer@ killer, u8 customdata)
...
+						}
+					}
+
+					if (victim.get_u8("killstreak") > 4)
+					{
+
+						uint16 victim_netid = victim.getNetworkID();
+						uint16 killer_netid = killer.getNetworkID();
+						uint16 kill_count = victim.get_u8("killstreak");
```
eps0003, line 320: `"killstreak"` is a u8 but you're saving it to a u16 variable and writing it to the CBitStream as a u16.

eps0003: This also occurs at line 372.

Rules/CommonScripts/KillMessages.as
```diff
@@ -324,3 +399,101 @@ void onRender(CRules@ this)
...
+		if (cmd == this.getCommandID("killstreak message"))
+		{
+			u16 player_netid, kill_count;
+
+			if (!params.saferead_u16(player_netid)
+			 || !params.saferead_u16(kill_count))
+			{
+				print("failed to parse killstreak message payload");
+				return;
+			}
```
eps0003, line 416: You don't need to use `saferead_u16` because you know what data to expect in the CBitStream. Just use `read_u16` instead.

eps0003: This also applies to line 456.

eps0003 (changes requested):

Rules/CommonScripts/KillMessages.as
eps0003: You also need to fix the places where you read killstreak as a u16.

Rules/CommonScripts/KillMessages.as
bunniewormy: I'll keep it to make the intention clear

bunniewormy:
fixed, thanks for the review

Rules/CommonScripts/KillMessages.as
eps0003: You've fixed writing to the bitstream, but `kill_count` is still a uint16. Simple, probably pointless fix.

Rules/CTF/Scripts/CTF_Trading.as
```diff
@@ -100,7 +101,7 @@ void onPlayerDie(CRules@ this, CPlayer@ victim, CPlayer@ killer, u8 customData)
 		{
 			if (killer !is victim && killer.getTeamNum() != victim.getTeamNum())
 			{
-				killer.server_setCoins(killer.getCoins() + coinsOnKillAdd);
+				killer.server_setCoins(killer.getCoins() + coinsOnKillAdd * (1+killer.get_u8("killstreak")*killstreakFactor));
```
eps0003, line 104: This equation is not correct. When killstreakFactor is 1, I would expect there to be no bonus for killstreaks. And if killstreakFactor is 2, I would expect the coin bonus to double with each kill.

eps0003 (dismissed):

eps0003:
I propose a better equation which exponentially scales the coin bonus received from killstreaks:

> killer.getCoins() + coinsOnKillAdd * Maths::Pow(killstreakFactor, killer.get_u8("killstreak"))

<details>
  <summary>Graph</summary>

  The green line is your equation with a killstreakFactor of 1.05, and the red line is my proposed equation with a killstreakFactor of 1.2.
  
  image

</details>

The benefit of exponential increase over linear increase is that it better rewards larger killstreaks. It's much harder to maintain higher and higher killstreaks, so the coin increase for each kill should become larger with each subsequent kill.

Another issue with your equation is that it gives way too many coins for such a small killstreak. At a 3 killstreak, the player is already receiving triple the number of coins. With my equation, you reach this point after a 7 killstreak which is significantly more balanced than what you propose.

Rules/CTF/Scripts/CTF_Trading.as
eps0003: My mistake, I misinterpreted the equation.

PR #900, opened 2020-09-04
Crate changes: pressing space with mines inside, boobytrap when trying to get in with enemy mines

bunniewormy:
## Status

## Description

Code is probably bad, please review uwu

- Crates are now boobytrapped if they have an enemy mine and you try to get in; the mine instakills you. Previously only opening inventory was boobytrapped (the mine goes up and instantly activates on inventory open, you can avoid it).

- Pressing space with crate in hand no longer dumps mines (previously it was only players)

### DIFF
--- a/Entities/Items/Crate/Crate.as
+++ b/Entities/Items/Crate/Crate.as
@@ -766,9 +766,7 @@ bool DumpOutItems(CBlob@ this, float pop_out_speed = 5.0f, Vec2f init_velocity =
 
 		while (inv !is null && (inv.getItemsCount() > target_items_left))
 		{
-			CBlob@ item;
-
-			@item = inv.getItem(item_num);
+			CBlob@ item = inv.getItem(item_num);
 
 			if (!item.hasTag("player") && item.getName() != "mine")
 			{
--- a/Entities/Items/Crate/Crate.as
+++ b/Entities/Items/Crate/Crate.as
@@ -429,7 +429,7 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
 		CBlob@ carrier = this.getAttachments().getAttachmentPointByName("PICKUP").getOccupied();
 		if (carrier !is null)
 		{
-			DumpOutItems(this, 5.0f, carrier.getVelocity(), false, false);
+			DumpOutItems(this, 5.0f, carrier.getVelocity(), false);
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


Entities/Items/Crate/Crate.as
```diff
@@ -346,10 +346,32 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
...
+					break;
+				}
+			}
+			if (mine !is null)
+			{
+				CBitStream params;
+				params.write_u16(caller.getNetworkID());
+				params.write_u16(mine.getNetworkID());
+				this.SendCommand(this.getCommandID("boobytrap"), params);
+				return;
```
eps0003, line 371: Since you return here, you should just move this inside the loop at line 361.

Entities/Items/Crate/Crate.as
```diff
@@ -740,19 +762,15 @@ bool DumpOutItems(CBlob@ this, float pop_out_speed = 5.0f, Vec2f init_velocity =
...
-			{
-				@item = inv.getItem(1);
-			}
-			else
-			{
-				@item = inv.getItem(0);
-			}
-			if (!item.hasTag("player"))
+
+			@item = inv.getItem(item_num);
```
eps0003, line 771: Variable declaration and assignment can be done in one line.
```suggestion
			CBlob@ item = inv.getItem(item_num);
```

Entities/Items/Crate/Crate.as
```diff
@@ -723,7 +745,7 @@ CBlob@ getPlayerInside(CBlob@ this)
 	return null;
 }
 
-bool DumpOutItems(CBlob@ this, float pop_out_speed = 5.0f, Vec2f init_velocity = Vec2f_zero, bool dump_player = true)
+bool DumpOutItems(CBlob@ this, float pop_out_speed = 5.0f, Vec2f init_velocity = Vec2f_zero, bool dump_player = true, bool dump_mine = true)
```
eps0003, line 748: Instead of having both `dump_player` and `dump_mine`, I suggest you just have one called `dump_special`. In most, if not all cases this method is used, you either want to dump out everything or dump out everything except players and mines. This way it allows more blobs to easily be added if needed in the future.

Entities/Items/Crate/Crate.as
bunniewormy: I've also edited the part at line ~520 too

Entities/Items/Crate/Crate.as
```diff
@@ -496,17 +513,13 @@ void onCreateInventoryMenu(CBlob@ this, CBlob@ forBlob, CGridMenu @gridmenu)
 		}
 		if (item.getName() == "mine" && item.getTeamNum() != forBlob.getTeamNum())
 		{
-			@mine = item;
+			CBitStream params;
+			params.write_u16(caller.getNetworkID());
```
Vam-Jam, line 517: error, caller should be forBlob

Vam-Jam (changes requested):

PR #905, opened 2020-09-13
Boats collision fix

ferrezinhre:
READY: this PR is (to the best of your knowledge) ready to be incorporated into the game.

Description
Fixes collison with enemy boats, I guess no one really care about this one anyway.

Steps to Test or Reproduce
one person on server (or localhost):

1.Ram enemy dinghy with longboat without code change, to see that you are unable to brake it*
2.Ram enemy dinghy with longboat with code change, to see that you are able to brake it*

(why dinghy? because one player in longboat can't generate enough speed to "hurt" other boats. )

This change will affect following vehicles in kag: dinghy, longboat, warboat.

### DIFF
--- a/Entities/Vehicles/Common/FakeBoatCollision.as
+++ b/Entities/Vehicles/Common/FakeBoatCollision.as
@@ -85,5 +85,5 @@ void onTick(CBlob@ this)
 
 bool doesCollideWithBlob(CBlob@ this, CBlob@ blob)
 {
-	return !blob.hasTag("fake boat collision") || (blob.hasTag("fake boat collision") && blob.getTeamNum() != this.getTeamNum()) ;
+	return !blob.hasTag("fake boat collision") || blob.getTeamNum() != this.getTeamNum();
 }


Entities/Vehicles/Common/FakeBoatCollision.as
```diff
@@ -85,5 +85,5 @@ void onTick(CBlob@ this)
 
 bool doesCollideWithBlob(CBlob@ this, CBlob@ blob)
 {
-	return !blob.hasTag("fake boat collision");
+	return !blob.hasTag("fake boat collision") || (blob.hasTag("fake boat collision") && blob.getTeamNum() != this.getTeamNum()) ;
```
eps0003, line 88: This boolean expression can be simplified:
```suggestion
	return !blob.hasTag("fake boat collision") || blob.getTeamNum() != this.getTeamNum() ;
```

eps0003 (changes requested):
Sorry.

eps0003:
Closed because of situations like this causing headaches. The left longboat wouldn't be able to move right because it collides with the player who is wedged between the prow of each longboat.
image

eps0003 (approved):

PR #907, opened 2020-09-20
Blobs recheck their collision with player after they crouched

CatWiz:
## Status

## Description

Crouching makes blobs that are touching player blob recheck their collisions with it. Fixes problem when crouching doesnt do anything if enemy keeps walking into you after touching you/is standing on you.


eps0003:
You should also do the same when a player dies. I've had situations where I cannot walk past a dead body because I killed them while colliding with them.

CatWiz:
RunnerDeath already does checkCollisionsAgain for the body. Should it make all blobs that are touching it recheck their collisions instead?

eps0003:
Not sure what needs to be done to fix it. I just wanted to make it known that similar behaviour occurs with dead bodies.

Entities/Characters/Scripts/CrouchCommon.as
```diff
@@ -12,3 +12,15 @@ bool isCrouching(CBlob@ this)
...
+bool hasJustCrouched(CBlob@ this)
+{
+	return
+	this.isOnGround()
+	&& this.isKeyJustPressed(key_down)
+	//cannot have movement intent
+	&& !this.isKeyPressed(key_left)
+	&& !this.isKeyPressed(key_right)
+	//cannot have banned crouch (done in actor logic scripts)
+	&& !this.hasTag("prevent crouch");
```
eps0003, line 25: There's no need to repeat the same chunk of code
```suggestion
	return isCrouching(this) && this.isKeyJustPressed(key_down);
```

eps0003 (changes requested):

PR #919, opened 2020-09-25
Fixed kaggen sometimes generating blocks inside flag

### LINKED ISSUE #588: Block inside flag
Normally you cannot place blocks inside flags. I think at least until the flag has been captured.

I had a game where enemy had a stone block inside of the flag, and it was not possible to destroy it as builder. So it was kind of game-breaking. We don't know how it occurred.

https://i.imgur.com/Z7rGmHR.png

When I find more info, I will update here.

eps0003:
## Status

## Description

This clears the flag of blocks when the flag's build sector is added.

## Steps to Test or Reproduce

My method:
1. Using code, set the corner tiles of the flag's build sector to any block before the code in this PR
2. Load any CTF map to see that these blocks were removed

Alternative method:
1. Make a CTF map with blocks inside the flag
2. Load the map

### DIFF
--- a/Entities/Special/CTF/FlagBase.as
+++ b/Entities/Special/CTF/FlagBase.as
@@ -51,6 +51,15 @@ void onTick(CBlob@ this)
 
 				map.server_AddSector(pos + Vec2f(-12, -32), pos + Vec2f(12, 16), "no build", "", this.getNetworkID());
 
+				//clear the no build zone so we dont get unbreakable blocks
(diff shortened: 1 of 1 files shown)


eps0003:
This won't fix the linked issue, but it will fix kaggen generating blocks in the flag.

Vam-Jam (approved):

AsuMagic (approved):

Issue #872, opened 2020-07-02
Improve animals


CatWiz:
Both bison and shark work in a weird and unconsistent way. Bisons can be tamed with burgers, but sharks cant (and yet there's an achievement for killing someone with a tamed shark that requires you to ride it, which nobody does). Shark attacks can be fully shielded, but bisons' cant. I propose changing them.
Bisons:
1. Remove superjump. Good meme but a bison shouldnt fly
2. Make bison attacks partially shieldable (so shielding reduces the knockback and damage, but doesnt completely negate them)
3. Make burgers slightly heal them
4. Add some way to tell if a bison is tamed and which team it belongs to (like a team colored saddle)
Sharks:
1. Make them tameable with a burger (but not fish)
2. Make it possible to ride and control a tamed shark in water, adding variety to water combat (faster than warboat, slower than max speed longboat). Maybe a little bit of control over it while its on the ground, like being able to make it jump left or right.
3. Make sharks slowly lose health while on land
4. Make burgers slightly heal them (maybe also fish, but it creates a risk of having immortal self-healing sharks on large water maps)
5. Same as bison, add a way to tell if its tamed by adding a team colored saddle

mugg91:
> 5\. Same as bison, add a way to tell if its tamed by adding a team colored saddle

I looked into this a bit.
It seems taming is a bit more complex than I thought.
The Bison considers the taming player its owner - if the owner blob disappears (can happen if you change class or die), then the Bison stops being friendly. It would be weird if it gets a saddle and nobody knows who the owner is. Then if he somehow disappears, "plop", the saddle disappears.

So I think some rework is in order.
When someone of Team 0 feeds a burger to a Bison, it should turn friendly to Team 0, period. And stay that way until one of these happen:
- Bison gets attacked by its own team -> turns neutral again.
- Bison gets fed by the other team -> What was the previous behavior? (I cannot test it alone) Should switch sides?

mugg91:
Build 3479 - Patch notes from 26th Oct 2020:
(...) - Fixed bison super jump (...)
So I take it is the same as your "Bisons 1)" and it has already been fixed.

PR #921, opened 2020-09-28
Fixed bison superjump

eps0003:
## Status

## Description

The slight upwards force applied while walking was being applied independently for left and right keys.
This addresses one of the suggestions in #872

## Steps to Test or Reproduce

1. Tame a bison using a burger
2. Ride the bison
3. Hold down, left and right to superjump


Trumbles:
I'm of the opinion that a bug like that is more charming than it is detrimental.

PR #923, opened 2020-09-28
Siege no longer shoots if gunner leaves while charging

eps0003:
## Status

## Description

Now the charge is reset when the gunner leaves mid charge. The catapult arm correctly retracts as well.

## Steps to Test or Reproduce

1. Sit in gunner seat of siege
2. Hold left click to charge
3. Press up to leave mid charge

### DIFF
--- a/Entities/Vehicles/Catapult/Catapult.as
+++ b/Entities/Vehicles/Catapult/Catapult.as
@@ -13,8 +13,6 @@ const u8 charge_contrib = 35;
 const u8 cooldown_time = 45;
 const u8 startStone = 100;
 
-bool hadAttached = false;
-
 void onInit(CBlob@ this)
 {
 	Vehicle_Setup(this,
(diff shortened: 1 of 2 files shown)


Vam-Jam (approved):

Entities/Vehicles/Catapult/Catapult.as
```diff
@@ -13,6 +13,8 @@ const u8 charge_contrib = 35;
 const u8 cooldown_time = 45;
 const u8 startStone = 100;
 
+bool hadAttached = false;
```
AsuMagic, line 16: This should be a blob property, as this global variable will be shared across multiple catapults

AsuMagic (changes requested):

Entities/Vehicles/Catapult/Catapult.as
eps0003: So would [this](https://github.com/transhumandesign/kag-base/blob/master/Entities/Items/Sponge/Sponge.as#L6) also be the same issue? This wasn't picked up when I improved sponges ages ago so I thought it was fine.

PR #928, opened 2020-10-05
Grass and bushes grow on dirt

CatWiz:
## Status

## Description

1. Grass and bushes can grow on a dirt tile if tile above is empty
2. Stone blocks and stone backwalls start turning into mossy stone blocks/backwalls 10 minutes after they were placed, starting from blocks that are adjacent to dirt/bedrock

~~Grass and bushes grow really fast, will probably have to further decrease chances.~~ Not relevant anymore after last major commit IMO

Stone turning into mossy stone is disabled by default for 2 reasons:
1. Moss stone has less health than normal stone. I like the idea of bases getting more fragile as match goes on, but without a drill nerf big enough being implemented anytime soon it doesnt seem justified to make bases passively get weaker. Could just increase mossy tiles health so its the same as normal castle tiles to make it purely cosmetic.
2. When a stone block/backwall is replaced with a mossy one, it plays "block built" sound and does a particle effect of previous tile breaking. Is annoying, and cant be fixed script-side afaik


Rules/CommonScripts/plants_regrow.as
```diff
@@ -0,0 +1,199 @@
+#define SERVER_ONLY
+
+const bool moss_stone = false; //should stone turn into mossy stone
+
+const f32 grass_grow_chance = 10;
+const f32 bush_grow_chance = 1;
+const f32 moss_stone_chance = 2;
+
+const u32 moss_time_min = 300*30*30; //10 minutes = 300*30*30
```
eps0003, line 9: This is in ticks, not minutes, so the variable should be named `moss_time_ticks` or just `moss_time`.
Also, how is 300 * 30 * 30 = 10 minutes? It should be 30 * 60 * 10 (ticks a second * number of seconds in a minute * number of minutes).

Rules/CommonScripts/plants_regrow.as
```diff
@@ -0,0 +1,199 @@
...
+
+		Tile tile_left = map.getTile(coords - Vec2f(map.tilesize, 0));
+		Tile tile_right = map.getTile(coords - Vec2f(-map.tilesize, 0));
+		Tile tile_above = map.getTile(coords - Vec2f(0, map.tilesize));
+		Tile tile_below = map.getTile(coords - Vec2f(0, -map.tilesize));
+
+		if (map.isTileGround(tile_below.type) || map.isTileBedrock(tile_below.type) || tile_below.type == CMap::tile_castle_moss || tile_below.type == CMap::tile_castle_back_moss) {luck += 2;}
+		if (map.isTileGround(tile_above.type) || map.isTileBedrock(tile_above.type) || tile_above.type == CMap::tile_castle_moss || tile_above.type == CMap::tile_castle_back_moss) {luck += 2;}
+		if (map.isTileGround(tile_left.type) || map.isTileBedrock(tile_left.type) || tile_left.type == CMap::tile_castle_moss || tile_left.type == CMap::tile_castle_back_moss) {luck += 2;}
+		if (map.isTileGround(tile_right.type) || map.isTileBedrock(tile_right.type) || tile_right.type == CMap::tile_castle_moss || tile_right.type == CMap::tile_castle_back_moss) {luck += 2;}
```
eps0003, line 62: There's a better way of writing this without repeating yourself. This also allows you to add the corner tiles without copying and pasting chunks of code if you choose to.

Ideally, someone should be an engine-side method called `isTileMossy` so you don't have to hardcode specific mossy tiles, but this will do for now I guess.

```suggestion
		Vec2f[] adjacentTilePos = {
			Vec2f(map.tilesize, 0),   //left
			Vec2f(-map.tilesize, 0),  //right
			Vec2f(0, map.tilesize),   //up
			Vec2f(0, -map.tilesize)   //down
		};

		for (u32 i = 0; i < adjacentTilePos.size(); i++)
		{
			Tile tile = map.getTile(coords - adjacentTilePos[i]);

			bool ground = map.isTileGround(tile.type);
			bool bedrock = map.isTileBedrock(tile.type);
			bool mossyCastle = tile.type == CMap::tile_castle_moss || tile.type == CMap::tile_castle_back_moss;

			if (ground || bedrock || mossyCastle)
			{
				luck += 2;
			}
		}
```

Rules/CommonScripts/plants_regrow.as
```diff
@@ -0,0 +1,199 @@
...
+		{
+			for (u32 y = 0; y < map.tilemapheight; y++)
+			{
+				Vec2f coords(x*map.tilesize, y*map.tilesize);
+
+				Tile tile = map.getTile(coords);
+				Tile tile_left = map.getTile(coords-Vec2f(map.tilesize, 0));
+				Tile tile_right = map.getTile(coords-Vec2f(-map.tilesize, 0));
+				Tile tile_above = map.getTile(coords-Vec2f(0, map.tilesize));
+				Tile tile_below = map.getTile(coords-Vec2f(0, -map.tilesize));
```
eps0003, line 84: Left, right and below are not used.
```suggestion
				Tile tile_above = map.getTile(coords-Vec2f(0, map.tilesize));
```

Rules/CommonScripts/plants_regrow.as
```diff
@@ -0,0 +1,199 @@
+#define SERVER_ONLY
+
+const bool moss_stone = false; //should stone turn into mossy stone
+
+const f32 grass_grow_chance = 10;
+const f32 bush_grow_chance = 1;
+const f32 moss_stone_chance = 2;
```
eps0003, line 7: It looks like these variables can range from 0-100 rather than 0-1. Why? It makes more sense if chance / probability / percentages range from 0-1.

Rules/CommonScripts/plants_regrow.as
```diff
@@ -0,0 +1,199 @@
...
+
+void onRestart(CRules@ this)
+{
+	CMap@ map = getMap();
+	tiles.clear();
+
+	if (map !is null)
+	{
+		if (!map.hasScript("grow.as")) map.AddScript("grow.as");
+	}
```
eps0003, line 106: You don't need this here because you have the same code on line 93.
```suggestion
```

eps0003 (changes requested):
Ignoring the obvious error you get when you try to run your code, here is some feedback:
* Add some comments to your code.
* Rename `plants_regrow.as` to `RegrowPlants.as` so it matches KAG's naming convention.
* Have an array of the two castle tiles and another array with their mossy equivalents. Then in your code you should refer to the arrays. This will allow more tiles to be added without needing to add it to several parts of the code.

I'll do a second pass once these are addressed.

Rules/CommonScripts/plants_regrow.as
Vam-Jam: I think this one is needed since its adding a new map script, which will be removed upon loading a new map iirc

Rules/CommonScripts/plants_regrow.as
```diff
@@ -0,0 +1,199 @@
...
+}
+
+void onTick(CRules@ this)
+{		
+	if (getGameTime() % random_time == 0)
+	{
+		CMap@ map = getMap();
+		float tilesize = map.tilesize;
+
+		for (int i = 1; i < tiles.length(); i++)
```
Vam-Jam, line 148: use .size() instead of .length(), stops warnings on asu's build

Rules/CommonScripts/plants_regrow.as
```diff
@@ -0,0 +1,199 @@
...
+	
+	Vec2f coords(x*this.tilesize,y*this.tilesize);
+
+	u32 tindex = findTileByCoords(tiles, coords);
+
+	if ((newtile == CMap::tile_castle || newtile == CMap::tile_castle_back) && tindex == 0) 
+		tiles.insertLast(TileInfo(coords, getGameTime(), this.getTile(coords)));
+	else
+	{
+		print("tindex "+tindex);
```
Vam-Jam, line 124: accidental debug message left in?

Rules/CommonScripts/plants_regrow.as
eps0003: I checked when each one is called. The one in onInit is called every time this one is called plus when the server starts for the first time.

Harrison-Miller:
Probably going to double the random_time

PR #935, opened 2020-10-11
Fixed being able to place blocks in no build sectors

### LINKED ISSUE #588: Block inside flag

eps0003:
## Status

## Description

How people are able to bypass the client buildable check is a mystery to me, but I know this fix will fix this. I just added the same client-side buildable check to server-side.

Resolves #588 

## Steps to Test or Reproduce

1. Replace PlacementCommon.as line 140 with `return isClient();` so you can place blocks in no build sectors on the client but not on the server.
2. Attempt to place a block inside a no build sector (e.g. tent or flag).

### DIFF
--- a/Entities/Common/Building/BlockPlacement.as
+++ b/Entities/Common/Building/BlockPlacement.as
@@ -18,10 +18,16 @@ void PlaceBlock(CBlob@ this, u8 index, Vec2f cursorPos)
 	CBitStream missing;
 
 	CInventory@ inv = this.getInventory();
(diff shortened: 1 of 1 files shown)


Vam-Jam (approved):

Vam-Jam:
No longer needed, pr #980 includes this

PR #955, opened 2020-11-11
[Fix] Fixes clients being unable to copy other teams username

Vam-Jam:
Title, caused by ``mouseWasPressed2 = controls.mousePressed2;`` happening at the end of drawScoreboard which is called for each team segment on the scoreboard. 

Moving to onRenderScoreboard fixes this


eps0003 (approved):
I don't know how we didn't notice this until now.

AsuMagic (approved):

PR #959, opened 2020-11-21
[Addition] Auto rebuild script

Vam-Jam:
Auto rebuild script as title says
Needs engine side PR merged for this to compile and work.

Script gets added via AddScript by localhost.as (which is called when using localhost.bat)
~~This can cause minor stutters, but we modders can disable/change how often its called, and we can work out how to improve it engine side later on.~~ read below

### DIFF
--- a/Scripts/Solo/AutoRebuild.as
+++ b/Scripts/Solo/AutoRebuild.as
@@ -1,0 +1,20 @@
+const bool enabled = true; // if disabled the script will be removed
+const uint checkTimer = 90; // 30 ticks in a second, default is 90 (3 seconds)
+
+void onInit(CRules@ this)
+{
+    print("Auto rebuild has been enabled.\nTo disable/edit this check AutoRebuild.as.\nCan cause stutters.");
+}
+
+void onTick(CRules@ this)
+{
+    if (getGameTime() % checkTimer == 0)
+    {
+        rebuild();
+    }
+
+    if (!enabled)
+    {
+        getRules().RemoveScript("AutoRebuild.as");
+    }
+}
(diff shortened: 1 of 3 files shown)


Vam-Jam:
Added AutoRebuild.cfg file for people to edit.
```
enabled = true # bool - Script will be removed if AutoRebuild has been disabled.

check_x_tick = false# bool - Do we check to rebuild every x amount of ticks (can cause stutters)
check_timer = 90 # s32 - how often is x for rebuild tick check

check_focus = true # bool - Do we check when screen goes into focus
```
Added focus check, which will rebuild every time you alt tab back into the game.

By default focus check is enabled, and tick check is disabled.

Scripts/Solo/AutoRebuild.cfg
```diff
@@ -0,0 +1,9 @@
+# Config settings for AutoRebuild script.
+
+enabled = true # bool - Script will be removed if AutoRebuild has been disabled.
+
+check_x_tick = false # bool - Do we check to rebuild every x amount of ticks (can cause stutters)
+check_timer = 90 # s32 - how often is x for rebuild tick check
```
eps0003, line 6: I suggest removing check_x_tick and making it so check_timer = 0 disables it instead

Scripts/Solo/AutoRebuild.as
```diff
@@ -0,0 +1,50 @@
...
+	check_timer = cfg.read_s32("check_timer", 90);
+	check_focus = cfg.read_bool("check_focus", true);
+
+	if (!enabled)
+	{
+		removeMe(this);
+	}    
+}
+
+void onTick(CRules@ this)
```
eps0003, line 29: Couldn't rebuild() be called twice on the same tick if both check_focus and check_x_tick are enabled? Could that be a problem?

Vam-Jam: side from checking if we should rebuild twice this is not an issue

Scripts/Solo/AutoRebuild.cfg
Vam-Jam: done

AsuMagic (approved):

PR #960, opened 2020-11-22
[Change] Moved some scripts to client/server only

Vam-Jam:
Commits have the list of files changed, more will be added as I find them.

The reason for this is because of an engine PR which will means the server/client will no longer attaches scripts they don't run to blobs/crules, which might result in a small performance improvement in some situations.


Harrison-Miller:
will need to wait for engine PR to be in a good state. And will need to double check these files as some funky stuff happens in a lot of scripts that might be important for either side even if it doesn't seem it.

Vam-Jam:
Will work on fixing the engine PR today and will double check the script files

Harrison-Miller:
@Vam-Jam can you fix the merge conflict.

Vam-Jam:
Done

Entities/Characters/Archer/Archer.cfg
```diff
@@ -183,8 +183,6 @@ $name                                             = archer
 													ClimbArrows.as;
 													ClimbTree.as;
 													Stomp.as;
-													RunnerHead.as;
```
Harrison-Miller, line 186: Why are we removing scripts?

Harrison-Miller: sprite only hooks - got it.

Harrison-Miller:
will need testing before release.

Issue #189, opened 2018-07-27
[Feature Pitch] Spectator AFK kick timer


eps0003:
People in spectator should have the same AFK timer as when being AFK while on a team.
The game should check for camera movement, mouse movement and chat interaction to see if someone isn't AFK.
Admins should be immune to this.

Pros:
- Prevents AFK people from taking up slots

Cons:
- Time-lapse recordings are slightly more difficult to do

AsuMagic:
If spectators using slots is a problem then the proper solution would be to add dedicated spectator slots (which needs in-engine changes however).

Shylie:
at some point (not sure when) an option to vote-kick spectators was added.

PR #963, opened 2020-11-23
[Addition] Spectator support

Vam-Jam:
ENGINE PR IS NEEDED FOR THIS TO WORK

Add's spectator slots, the server api/browser already supports this, so there isnt much work that needs to go in there.

Autoconfig settings:
``sv_spectatorslots``, default value is 4

Binds:
``getPlayersCount_NotSpectator()``, ``getNet().joined_maxplayers``, ``getNet().joined_spectatorslots``, ``getNet().joined_reservedslots``

What does this do:
```
- Players who join will get forced to team 200 (spectator team) if game is full
- Patreons who join will be put into a team AS LONG AS there are enough patreon slots (2 by default)
- People with reserved slots will be put into a team if there's a spare slot, otherwise they will become a spectator
- People with joinfull access will be put into a team no matter what
```

### DIFF
--- a/Rules/CommonScripts/CoreHooks.as
+++ b/Rules/CommonScripts/CoreHooks.as
@@ -5,6 +5,7 @@
 //Make sure you dont forget JoinCoreHooks! :)
 
 #include "GameplayEvents.as"
+#include "SwitchFromSpec.as"
 
 //not server only so that all the players get this
 void onInit(CRules@ this)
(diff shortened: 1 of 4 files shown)


Vam-Jam:
Engine PR is made and finished the script side support

eps0003:
Also resolves #189

Harrison-Miller:
Will need testing before release.

PR #964, opened 2020-11-23
[Change] Client side gui no longer use cmd

### LINKED ISSUE #612: Make opening shop menus clientside

if youre playing with ping you wait 2 years for a shop menu to open

**Proposed Change**

shop menus would open instantly now - just buying would be serverside

this should be the same with menu for changing classes

Vam-Jam:
Title, cmd would result in a delay depending on your ping. I updated the engine to include new binds for CreateGenericButton to have an option for a function callback, so if we do have any client only code we want to call, we can use that.

I tested using clumsy, your ping no longer makes a difference on how fast the gui's appear.

Requires engine pr to be merged to work
```
New funcdef: void CallbackButtonFunc(CBlob@, CBlob@)

New script binds:
CButton@ CreateGenericButton( int _frameNum, Vec2f _offset, CBlob@ attached, CallbackButtonFunc@ cb, const string &in )
CButton@ CreateGenericButton( const string &in iconName, Vec2f _offset, CBlob@ attached, CallbackButtonFunc@ cb, const string &in )
```

Fixes #612

### DIFF
--- a/Entities/Common/Factory/Shop.as
+++ b/Entities/Common/Factory/Shop.as
@@ -57,7 +57,7 @@ void GetButtonsFor(CBlob@ this, CBlob@ caller)
 			this.get_u8("shop icon"),                                // icon token
 			this.get_Vec2f("shop offset"),                           // button offset
 			this,                                                    // shop blob
-			CreateMenu,                                              // func callback
+			createMenu,                                              // func callback
 			getTranslatedString(this.get_string("shop description")) // description
 		);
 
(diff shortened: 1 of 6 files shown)


Vam-Jam:
This fixes lag with opening shop menu's and respawn menu on tents, ballistas, halls and warboats currently. 
Let me know if there's any other gui that needs the same treatment

Vam-Jam:
Accessing storage from other blobs is now instant

AsuMagic (approved):

PR #965, opened 2020-11-25
[Change] Sandbox changes 

Vam-Jam:
- ``!crate blacklisted`` item will no longer spawn said blacklisted item
- Shops will no longer spawn into the ground 
- Mats will decay extremely quickly

### DIFF
--- a/Entities/Common/Decaying/DecayQuantity.as
+++ b/Entities/Common/Decaying/DecayQuantity.as
@@ -6,46 +6,48 @@
 // discarded each step
 
 const uint16 FREQUENCY = 2000;
+const uint16 QUICK_FREQUENCY = 30;
 const uint16 QUARTER = FREQUENCY / 4;
 
 void onInit(CBlob@ this)
 {
-  ScriptData@ script = this.getCurrentScript();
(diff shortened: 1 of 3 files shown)


Entities/Common/Decaying/DecayQuantity.as
```diff
@@ -6,46 +6,48 @@
...
 {
-  ScriptData@ script = this.getCurrentScript();
-  script.runFlags |= Script::tick_not_attached;
-  script.runFlags |= Script::tick_not_ininventory;
-  script.tickFrequency = FREQUENCY;
+	ScriptData@ script = this.getCurrentScript();
+	script.runFlags |= Script::tick_not_attached;
+	script.runFlags |= Script::tick_not_ininventory;
+
+	this.getCurrentScript().tickFrequency = (getRules().hasTag('quick decay') ? QUICK_FREQUENCY : FREQUENCY);
```
AsuMagic, line 18: it would be better to have a function that does this instead of having all those duplicate ternary expressions, but i'll merge anyway for now

AsuMagic (approved):

PR #971, opened 2020-12-08
Saw breaks if it explodes too many bombs at once

CatWiz:
## Status

## Description

Title. Done so sawjumps cost a saw for every attempt

### DIFF
--- a/Entities/Industry/Saw/Saw.as
+++ b/Entities/Industry/Saw/Saw.as
@@ -238,6 +238,33 @@ void onCollision(CBlob@ this, CBlob@ blob, bool solid)
 		Vec2f bpos = blob.getPosition();
 		this.Tag("sawed");
 		this.server_Hit(blob, bpos, bpos - pos, 0.0f, Hitters::saw);
+
+		if (blob.getName() == "bomb")
+		{
+			if (this.get_u8("bombs_exploded") == 0)
+			{
+				this.set_u32("bomb_time", getGameTime());
+			}
+
+			this.add_u8("bombs_exploded", 1);
+		}
+	}
+}
+
+void onTick(CBlob@ this)
+{
+	if (this.get_u32("bomb_time") == 0) return;
+
+	if (getGameTime() - this.get_u32("bomb_time") > 8)
+	{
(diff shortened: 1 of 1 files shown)


Entities/Industry/Saw/Saw.as
```diff
@@ -238,6 +238,33 @@ void onCollision(CBlob@ this, CBlob@ blob, bool solid)
...
+			{
+				this.set_u32("bomb_time", getGameTime());
+			}
+
+			this.add_u8("bombs_exploded", 1);
+		}
+	}
+}
+
+void onTick(CBlob@ this)
```
Vam-Jam, line 254: This can be done without having to use onTick, its not a big deal but imo if we can avoid using it, it would be best

Vam-Jam (approved):

Entities/Industry/Saw/Saw.as
```diff
@@ -238,6 +238,33 @@ void onCollision(CBlob@ this, CBlob@ blob, bool solid)
 		Vec2f bpos = blob.getPosition();
 		this.Tag("sawed");
 		this.server_Hit(blob, bpos, bpos - pos, 0.0f, Hitters::saw);
+
+		if (blob.getName() == "bomb")
+		{
+			if (this.get_u8("bombs_exploded") == 0)
```
AsuMagic, line 244: the `"bombs_exploded"` property is never set before. i assume that `get_u8` (also used by `add_u8` internally) assigns `0` as a default value but i suspect this causes a warning and should not be relied upon anyway.

Entities/Industry/Saw/Saw.as
```diff
@@ -238,6 +238,33 @@ void onCollision(CBlob@ this, CBlob@ blob, bool solid)
...
+			}
+
+			this.add_u8("bombs_exploded", 1);
+		}
+	}
+}
+
+void onTick(CBlob@ this)
+{
+	if (this.get_u32("bomb_time") == 0) return;
```
AsuMagic, line 256: same thing here

AsuMagic (approved):

PR #980, opened 2020-12-14
[Change] Server now checks placeBlob and placeBlock

Vam-Jam:
Server will now do some checks before allowing you to place a tile down.

These include
- Range check (compensates for lag)
- Delay check (^)
- Don't allow user to place on dirt/bedrock
- Don't allow users to place tiles at tiles with no support


Vam-Jam:
Added no build zone check, along with is tile collapsing at the current pos check

PR #985, opened 2020-12-22
Quarry becomes less efficient each time it drops mats

CatWiz:
## Status

## Description

1. Starting output slightly increased (75 => 80)
2. Output decreases by 2 each time stone is dropped, until it reaches 30. Because stone drops are rounded to 5, it should drop same amount of stone 2-3 times before going down by 5
3. Same amount of wood is consumed for all outputs
4. Breaking the quarry doesn't reset it's output
5. Output resets on map change/restart

### DIFF
--- a/Entities/Industry/CTFShops/Quarry/Quarry.as
+++ b/Entities/Industry/CTFShops/Quarry/Quarry.as
@@ -11,7 +11,7 @@ const string rare_ore = "mat_gold";
 const int input = 100;					// input cost in fuel
 const int initial_output = 80;			// output amount in ore
 const int min_output = 30;				// minimal possible output in ore
-const int output_decrease = 3;			// by how much output decreases every time ore is dropped
+const int output_decrease = 2;			// by how much output decreases every time ore is dropped
 const bool enable_rare = false;			// enable/disable
 const int rare_chance = 10;				// one-in
 const int rare_output = 20;				// output for rare ore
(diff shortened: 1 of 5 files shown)


Entities/Industry/CTFShops/Quarry/Quarry.as
```diff
@@ -195,6 +204,8 @@ void spawnOre(CBlob@ this)
 	_ore.server_SetQuantity(!rare ? amountToSpawn : rare_output);
 
 	this.set_s16(fuel_prop, blobCount - actual_input); //burn wood
+	// reduce output if it's higher than minimal output
+	getRules().set_s32("current_quarry_output_" + this.getTeamNum(), Maths::Max(getRules().get_s32("current_quarry_output_" + this.getTeamNum()) - output_decrease, min_output));
```
eps0003, line 208: Would it look more elegant if you used `CRules::add_s32()`?

Rules/CTF/Scripts/CTF.as
```diff
@@ -685,6 +685,11 @@ void Reset(CRules@ this)
 	this.set("core", @core);
 	this.set("start_gametime", getGameTime() + core.warmUpTime);
 	this.set_u32("game_end_time", getGameTime() + core.gameDuration); //for TimeToEnd.as
+
+	for (int team = 0; team < this.getTeamsCount(); team++)
```
eps0003, line 689: What about other gamemodes? Why is this only in CTF.as?

Entities/Industry/CTFShops/Quarry/Quarry.as
Vam-Jam: Could be worth adding a wrapper for get current_output and another for set current output, would look much better

Rules/CTF/Scripts/CTF.as
CatWiz: Because quarries dont exist in TTH, and in sandbox quarry output doesn't matter. I'll make it decrease output only in CTF modes.

Entities/Industry/CTFShops/Quarry/Quarry.as
```diff
@@ -195,6 +204,13 @@ void spawnOre(CBlob@ this)
 	_ore.server_SetQuantity(!rare ? amountToSpawn : rare_output);
 
 	this.set_s16(fuel_prop, blobCount - actual_input); //burn wood
+	const string current_output = "current_quarry_output_" + this.getTeamNum();
+	
+	// reduce output if it's higher than minimal output
+	if (sv_gamemode == "CTF" || sv_gamemode == "SmallCTF")
```
eps0003, line 210: Hardcoding gamemodes like this is bad. Find a better way to only apply it to specific gamemodes.

Rules/CTF/Scripts/CTF.as
eps0003: But what if this changes? You should implement this in a way that works for any gamemode that needs this behaviour.

eps0003 (changes requested):

eps0003 (approved):
Nice work!

PR #991, opened 2020-12-23
Saplings and gold that falls in the void reappears at the top of the map.

Hobey:
- **READY**   (tested on modded CTF server)

[Demo](https://user-images.githubusercontent.com/26771587/102968642-89d3cc00-44f4-11eb-8b48-f61729defa22.mp4)

Destroying enemy gold and saplings is sometimes fun to do and sometimes it's fun to counterplay.
Sometimes I find though that gold and saplings should be an **actually permanent** resource.

Definitely a mixed opinion change. Personally I would prefer gold and saplings to be permanent.

### DIFF
--- a/Entities/Materials/Construction/MaterialGold.as
+++ b/Entities/Materials/Construction/MaterialGold.as
@@ -3,5 +3,47 @@ void onInit(CBlob@ this)
 {
   this.maxQuantity = 50;
 
-  this.getCurrentScript().runFlags |= Script::remove_after_this;
+  //this.getCurrentScript().runFlags |= Script::remove_after_this;
 }
+
+void onDie(CBlob@ blob) {
+    if (isServer()) {
+        if (blob.getQuantity() != 0) {
+            if (!blob.hasTag("merge_killed")) {
+                
+                // NOTE(hobey): blob fell in the void; resummon the blob at the top of the map
+                
+                CMap@ map = getMap();
+                
+                Vec2f pos = blob.getPosition();
+                
+                float min_x = 16;
+                float max_x = (map.tilemapwidth * map.tilesize - 16);
+                if (pos.x < min_x) pos.x = min_x;
+                if (pos.x > max_x) pos.x = max_x;
+                
+                pos.y = 0;
+                bool do_summon = true;
+                while (true) {
+                    if (!map.isTileSolid(pos)) break;
+                    pos.y += 8;
+                    if (pos.y > (map.tilemapheight * map.tilesize)) {
+                        do_summon = false; // NOTE(hobey): if the blob fell into the void, and there's a colomn of blocks from top to bottom of the map, delete the blob for good
+                        break;
+                    }
+                }
+                
+                if (do_summon) {
+                    CBlob@ new_blob = server_CreateBlobNoInit(blob.getName());
+                    if (new_blob !is null) {
+                        new_blob.Tag('custom quantity');
+                        new_blob.Init();
+                        new_blob.server_SetQuantity(blob.getQuantity());
+                        new_blob.setPosition(pos);
+                    }
+                }
+            }
+        }
+    }
+}
+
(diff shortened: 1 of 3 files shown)


CatWiz:
Does it preserve item's momentum? Looks like it doesn't. Maybe it should.

Hobey:
I doesn't preserve momentum atm; the blob actually dies and that's what I'm using to detect that it fell in the void, and I just create a copy of the blob at the top of the map at the same x coordinate; definitely could just set the velocity (would make it feel way more toroidal!) 

dunno if there's maybe a way to keep a blob alive from inside of onDie(); that would preserve all properties (which doesn't really matter, but would be nice; also would be computationally cheaper probably)

Entities/Materials/Construction/MaterialGold.as
```diff
@@ -3,5 +3,47 @@ void onInit(CBlob@ this)
...
   this.maxQuantity = 50;
 
-  this.getCurrentScript().runFlags |= Script::remove_after_this;
+  //this.getCurrentScript().runFlags |= Script::remove_after_this;
 }
+
+void onDie(CBlob@ blob) {
+    if (isServer()) {
+        if (blob.getQuantity() != 0) {
+            if (!blob.hasTag("merge_killed")) {
```
eps0003, line 12: Merge these three if statements into one

Entities/Materials/Construction/MaterialGold.as
```diff
@@ -3,5 +3,47 @@ void onInit(CBlob@ this)
...
+                
+                // NOTE(hobey): blob fell in the void; resummon the blob at the top of the map
+                
+                CMap@ map = getMap();
+                
+                Vec2f pos = blob.getPosition();
+                
+                float min_x = 16;
+                float max_x = (map.tilemapwidth * map.tilesize - 16);
```
eps0003, line 21: Shouldn't you instead use the blob's radius instead of constants?

Entities/Materials/Construction/MaterialGold.as
```diff
@@ -3,5 +3,47 @@ void onInit(CBlob@ this)
...
+                if (pos.x < min_x) pos.x = min_x;
+                if (pos.x > max_x) pos.x = max_x;
+                
+                pos.y = 0;
+                bool do_summon = true;
+                while (true) {
```
eps0003, line 27: Why not use a for loop? Infinite while loops are scary!

Entities/Materials/Construction/MaterialGold.as
```diff
@@ -3,5 +3,47 @@ void onInit(CBlob@ this)
...
+                    if (!map.isTileSolid(pos)) break;
+                    pos.y += 8;
+                    if (pos.y > (map.tilemapheight * map.tilesize)) {
+                        do_summon = false; // NOTE(hobey): if the blob fell into the void, and there's a colomn of blocks from top to bottom of the map, delete the blob for good
```
eps0003, line 31: Maybe instead of deleting the blob, you instead search horizontally outwards for a valid position.

Entities/Natural/Seed/Seed.as
```diff
@@ -159,3 +160,37 @@ bool doesCollideWithBlob(CBlob@ this, CBlob@ blob)
 {
 	return blob.getShape().isStatic() && blob.isCollidable();
 }
+
+void onDie (CBlob@ blob) {
```
eps0003, line 164: You're pretty much repeating the exact same code as in the other file. Figure out a way to reuse most of this code.

eps0003 (changes requested):

Hobey:
I'm won't edit this branch; I'm exhausted.

Harrison-Miller:
1. Should probably find the nearest ground to spawn it over, because it may just keep looping over the void.
2. What happens if a player has gold in their inventory and voids out.

Hobey:
Closing this branch since the meaning of gold will probably change significantly through the changes of CTF Rebuild. 

I might revisit this idea if gold-permanence ends up being a thing people want, but then it would be implemented with parachute crates, or a bird bringing the gold back out of the void, or some continuous natural regeneration of gold in the map up to a certain limit.

PR #992, opened 2020-12-23
Fixed sponges drying water tiles when they are in your inventory.

Hobey:
- **READY**

[Bug](https://user-images.githubusercontent.com/26771587/102972153-e803ad80-44fa-11eb-8ef6-20f6d2919eea.mp4)
[Fix](https://user-images.githubusercontent.com/26771587/102972160-eafe9e00-44fa-11eb-8c01-ed2af814f3ef.mp4)

### DIFF
--- a/Entities/Items/Sponge/Sponge.as
+++ b/Entities/Items/Sponge/Sponge.as
@@ -29,7 +29,7 @@ void onTick(CBlob@ this)
 	f32 tilesize = map.tilesize;
 
 	//absorb water
-	if (absorbed < ABSORB_COUNT)
+	if (absorbed < ABSORB_COUNT && !this.isInInventory())
 	{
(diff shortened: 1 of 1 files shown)


Entities/Items/Sponge/Sponge.as
```diff
@@ -29,7 +29,7 @@ void onTick(CBlob@ this)
 	f32 tilesize = map.tilesize;
 
 	//absorb water
-	if (absorbed < ABSORB_COUNT)
+	if (absorbed < ABSORB_COUNT && !this.isInInventory())
```
Vam-Jam, line 32: Would it not be better to use runflag tick_not_ininventory

Hobey: Right! I changed it now

AsuMagic (approved):

PR #1003, opened 2021-01-05
[Change] Red Barrier Rewrite

Vam-Jam:
Rewrote red barrier from scratch, mainly because the old barrier code got a bit messy and to fix some issues we've been seeing with it.

Changes:
- x1 and x2 is no long f32, instead its u16 (since we didnt really need it to be f32)
- We only sync the 2 x positions instead of 2 x positions and 2 y positions 
- RedBarrierVar.cfg is now an option for servers
- Server vars will only sync if `blob_pushback` was changed in the cfg (only var client needs)
- Spikes and materials will get pushed by the barrier 
- Code might be cleaner?

### DIFF
--- a/Rules/CommonScripts/RedBarrier.as
+++ b/Rules/CommonScripts/RedBarrier.as
@@ -85,7 +85,7 @@ void onTick(CRules@ this)
 			CBlob @b = blobsInBox[i];
 
 			if (!b.getShape().isStatic() && 
-				(b.getTeamNum() < 200 || b.hasTag("no barrier pass") || 
+				(b.getTeamNum() < 100 || b.hasTag("no barrier pass") || 
 				 b.hasTag("material") || b.getName() == "spikes"))
 			{
 				PushBlob(b, middle);
@@ -241,7 +241,6 @@ void SyncToPlayer(CRules@ this, CPlayer@ player)
 // We only send this if we know that the cfg has been edited
 void SyncVarsToPlayer(CRules@ this, CPlayer@ player)
 {
-	print("syncing vars!");
 	// Only send pushback as its the only one client needs
 	CBitStream stream = CBitStream();
 	stream.write_f32(VEL_PUSHBACK);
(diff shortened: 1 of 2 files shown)


Vam-Jam:
Same as the other PR, consider this abandoned unless this is something people want, as its been almost 2 years now.

Ladder issue is a nice catch tho, this is most likely caused by line 209, we add a sector called "barrier" but not "no build". This is intended since the current version of barrier *can* remove flag no build zones (rarely), but i forgot to update builder code.

The fix for this is to update builder to take into account "barrier" flag when deciding if it can place blobs.

Gingerbeard5773:
#1305 is still a problem and this PR would fix it
--> https://github.com/Vam-Jam/kag-base/pull/2 <-- @Vam-Jam

Vam-Jam:
Merged it in, thanks. I'll look into sorting the ladder issue later in that case.

Gingerbeard5773:
-> https://github.com/Vam-Jam/kag-base/pull/3 <- @Vam-Jam

Gingerbeard5773:
Ready for review. With sectors removed there is no possible way for the barrier to desync (to my knowledge). This PR will fix #1305

Rules/CommonScripts/RedBarrier.as
```diff
@@ -1,291 +1,245 @@
...
+
+	CMap@ map = getMap();
+
+	const u16 x1 = this.get_u16("barrier_x1");
+	const u16 x2 = this.get_u16("barrier_x2");
+	const u16 middle = (x1 + x2) * 0.5f;
 
-	if (shouldBarrier(this))
+	CBlob@[] blobsInBox;
+	if (map.getBlobsInBox(Vec2f(x1, 0), Vec2f(x2, map.tilemapheight * map.tilesize), @blobsInBox))
```
AsuMagic, line 81: This makes it so that you can shoot things over the red barrier. It might be a problem for SmallCTF.

AsuMagic (dismissed):

Rules/CommonScripts/RedBarrier.as
Gingerbeard5773: Nice catch. 50 or so tiles extension is much more than enough to catch all objects.
```suggestion
	if (map.getBlobsInBox(Vec2f(x1, -50 * map.tilesize), Vec2f(x2, map.tilemapheight * map.tilesize), @blobsInBox))
```

Gingerbeard5773 (changes requested):

AsuMagic (approved):

PR #1009, opened 2021-01-30
Added -1 as a valid option for accolades

Vam-Jam:
Title


furai:
@Vam-Jam weird change. First of all I could give Asu permanent tier 4 in the database cause he deserves it.

I'm not sure that -1 entirely makes sense there.

Harrison-Miller:
Gonna merge, because I don't see a problem with having this as an option.

PR #1014, opened 2021-02-21
damaging stone and thick stone drop stone at feet like gold does

Kuift:
## Status

**READY**

## Description

Previously, when a knight hit dirt stone or dirt thick stone, it would not drop any of the stone mats. This PR make it so that when a knight damage a dirt stone or dirt thick stone, 4 or 6 stone mats are dropped at the feet of the knight.

## Steps to Test
Go knight, hit dirt stone, notice that 4 stone dropped at your feet.
Go knight, hit dirt thick stone, notice that 6 stone dropped at your feet.

### DIFF
--- a/Entities/Characters/Knight/KnightLogic.as
+++ b/Entities/Characters/Knight/KnightLogic.as
@@ -1288,6 +1288,7 @@ void DoAttack(CBlob@ this, f32 damage, f32 aimangle, f32 arcdegrees, u8 type, in
 				{
 					bool ground = map.isTileGround(hi.tile);
 					bool dirt_stone = map.isTileStone(hi.tile);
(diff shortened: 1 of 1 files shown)


Entities/Characters/Knight/KnightLogic.as
```diff
@@ -1347,6 +1348,28 @@ void DoAttack(CBlob@ this, f32 damage, f32 aimangle, f32 arcdegrees, u8 type, in
 	     								ore.server_SetQuantity(4);
 	     							}
 								}
+								else if (dirt_thick_stone)
```
Harrison-Miller, line 1351: combine this into one if statement and fix  the last bracket indenting

PR #1020, opened 2021-03-08
Admins can view inventories of players and other things while in spectator team

CatWiz:
## Status

## Description

Title. Hold left control and aim at a player or anything that has an inventory to see what's inside.
Works only in spectator team because i think it would be too big of a gameplay advantage to give to just admins, and it can be used for trolling too easily to give it to everyone.
Items are drawn in rows of 2, so it doesn't look good with storages/crates/etc, but there's no way to get inventory dimensions that i know of, and you're most likely to use it on players anyways.


Rules/CommonScripts/AdminInventoryView.as
```diff
@@ -0,0 +1,71 @@
...
+
+	CPlayer@ player = getLocalPlayer();
+	if (player is null) return;
+
+	CControls@ controls = player.getControls();
+	if (controls is null) return;
+
+	Vec2f mouse_pos = controls.getMouseWorldPos();
+
+	if (player.isMod() && player.getTeamNum() == this.getSpectatorTeamNum() && controls.isKeyPressed(KEY_LCONTROL))
```
Vam-Jam, line 14: You should remove this script if player is not a mod, this would save having to all the script every frame.

OR

Remove this script from all gamemode.cfg and add onChangeTeam in core gamemode files. Then check to see if an admin has switched to spec (if so, add file, and remove when they leave). Then you can get rid of checking isMod and getTeamNum every frame.

Along with that, define it client only at top of the script

Rules/CommonScripts/AdminInventoryView.as
```diff
@@ -0,0 +1,71 @@
...
+			}
+		}	
+	}
+}
+
+void DrawInventoryOfBlob(CBlob@ this)
+{
+	CInventory@ inv = this.getInventory();
+	string[] drawn;
+	Vec2f tl = this.getPosition();
```
Vam-Jam, line 38: You're calling this every frame, getPosition only get's updated every tick. This would look stuttery if said blob was moving.

Call getInterpolatedPosition() instead to make it smooth.

Vam-Jam (changes requested):

eps0003:
> Items are drawn in rows of 2, so it doesn't look good with storages/crates/etc, but there's no way to get inventory dimensions that i know of, and you're most likely to use it on players anyways.

You could use the number of items in the inventory to display them in a grouping that looks good. An equation that would work well is `columns = Maths::Ceil(Maths::Sqrt(n))`, where n is the number of items in the inventory. This displays the items within a square grid.

PR #1030, opened 2021-04-14
You can now eat while stunned with hotkey, burger price increased to 30 coins

bunniewormy:
## Status

## Description

Currently you cannot eat by pressing V, but you can by taking an item out of your inventory (which requires a bit more clicks and mouse movement). While this may seem like a bug to some people, fixing it would change the balance of the game quite a bit - water would be harder to counter, you couldn't change your currently held item while stunned (eg take the drill out as a builder while stunned, or a lit bomb, which could be quite annoying). Moreover currently with quick enough reflexes you're able to survive a double slash if you heal quickly enough which can allow for skillful clutches. See discussion in #development from 11-13 April for more.

To compensate for this change, price of burgers is increased from 20 to 30 coins

### DIFF
--- a/Entities/Characters/Scripts/EatFoodButton.as
+++ b/Entities/Characters/Scripts/EatFoodButton.as
@@ -12,7 +12,6 @@ void onTick(CBlob@ this)
 {
 	if (isServer() &&
 		this.isKeyJustPressed(key_eat) &&
-		!isKnocked(this) &&
 		this.getHealth() < this.getInitialHealth()
 	) {
 		CBlob @carried = this.getCarriedBlob();
--- a/Entities/Common/Includes/Costs.as
+++ b/Entities/Common/Includes/Costs.as
@@ -144,7 +144,7 @@ void InitCosts()
 	CTFCosts::beer =                        ReadCost(cfg, "cost_beer"               , 5);
 	CTFCosts::meal =                        ReadCost(cfg, "cost_meal"               , 10);
 	CTFCosts::egg =                         ReadCost(cfg, "cost_egg"                , 30);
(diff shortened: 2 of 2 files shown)


benjamingorman:
Looks good

eps0003 (approved):

ewetcr:
Looks like an undercover idea to regularize something that could be considered a bug. If stuns don't completely stun the balance of it isn't so sound. For instance, bad timing as knight shouldn't be mended with fast burger eating it, adespite how much skill it should require to fast open teh inventory and pick an item.
Moreover, the reasons to hold it seem a roundabout to not to fix the bug. Doubleslashes should be lethal plus changing the price of burgers looks like an ill idea to hold this. Most of the players won't use clutch burger eating nor will want to learn it and will find it unfair that it is even possible.

NotLaikaDawg:
"@AsuMagic another troll issue to close"
-Bunnie

NotLaikaDawg:
for real though this has been mentioned in #1061 and another prior issue, and is a known bug, who would be dumb enough to actually implement a dedicated way to do it, its literally "an engine side bug, ill need to fix it later" my good friend vamist 2021

NotLaikaDawg:
side note: banning people who point out bugs, because a notable(ly toxic) member of the community who owns patreon happens to use those bugs is kinda dumb, and you cant silence the truth

NotLaikaDawg:
I already know bunnie is gonna suck a fat one to get me banned again, or buy another tier 4 patreon sub, because I dared say glitches dont take skill, so we are gonna do my classic witnesses segment again: 
Witness One (Identity Protected for his own safety)
witness one
Witness Two (Blav is so based he doesnt need protection)
witness two
Blav Upon seeing this god awful pr for the first time

PR #1033, opened 2021-04-17
Change respawn times for player numbers

### LINKED ISSUE #746: Shorter respawn time depending on amount of players in a team
Current respawn time (10 seconds) feels too long in small matches. Matches with a small amount of players often end too fast due to mistakes being more costly (less players in team to defend). Reducing respawn time length could fix that issue. My suggestion is:
12-10 players per team - 10 seconds
9-8 - 9 seconds
7-6 - 8 seconds
5-4 - 7 seconds
4-0 - 6 seconds

laws65:
## Status

## Description

Resolves #746 

## Steps to Test or Reproduce

1. Start a private game
2. Spawn bots in
3. Die and note respawn time


Rules/CTF/Scripts/CTF.as
```diff
@@ -255,7 +255,10 @@ shared class CTFSpawns : RespawnSystem
 			CTFTeamInfo@ team = cast < CTFTeamInfo@ > (CTF_core.teams[info.team]);
 
 			info.can_spawn_time = ((old_spawn_time > 30) ? old_spawn_time : tickspawndelay);
-
+			float normalSpawnTimeInSecs = CTF_core.spawnTime / getTicksASecond();
+            float timeAdjustedForTeamNums = Maths::Floor(normalSpawnTimeInSecs - (normalSpawnTimeInSecs - team.players_count) / 2.0);
+            float respawnTimeSecs = Maths::Clamp(timeAdjustedForTeamNums, normalSpawnTimeInSecs/5.0*3.0, normalSpawnTimeInSecs);
+            info.can_spawn_time = respawnTimeSecs * getTicksASecond();
```
eps0003, line 261: Fix indentation

Rules/CTF/Scripts/CTF.as
```diff
@@ -255,7 +255,10 @@ shared class CTFSpawns : RespawnSystem
 			CTFTeamInfo@ team = cast < CTFTeamInfo@ > (CTF_core.teams[info.team]);
 
 			info.can_spawn_time = ((old_spawn_time > 30) ? old_spawn_time : tickspawndelay);
```
eps0003, line 257: Remove the old implementation

eps0003 (changes requested):

eps0003 (approved):

terracraft321:
What about the other gamemodes?

Vam-Jam:
> What about the other gamemodes?

- TTH has far too many issues right now, this change is not suited for it.
- Small CTF uses CTF.as, so it'll use this change.
- TDM does not need this.

Vam-Jam (approved):

AsuMagic:
OK, so I did leave that PR hanging for a while but have not addressed the reasons why as I should have.

I generally don't like the magic value and formula approach taken for implementing this.  
It makes it more difficult for people to understand and edit such values, and makes the assumption that we want to scale respawn time linearly with the amount of players, which is not the case.

It would make much more sense to me to explicitly use some threshold and hardcoded respawn time values.  
This would allow configuring the threshold and respawn times through the CTF configuration file as well.

asumagic:
Closing for inactivity + there hasn't been much demand for it, although we could look in other ways to rebalance low-player count CTF

PR #1038, opened 2021-04-26
Improved lighting bombs with bad ping

### LINKED ISSUE #1037: Lighting and throwing bombs is difficult with bad ping
**Description**

When you try to light a bomb and throw it with bad ping, it sometimes lights two bombs instead of throwing the first. The first bomb drops below you. This is especially an issue with water bombs since you accidentally stun yourself when this happens.

**Reproduction**

1. Have bad ping (join foreign server)
2. Have two bombs in your inventory
3. Press space twice somewhat quickly
4. See the first bomb drop below you instead of being thrown

**Expected Behavior**

The code that attempts to light a bomb should check if the player is already holding a lit bomb (or anything) and throw it instead of lighting a new bomb.

eps0003:
## Status

## Description

Simple fix that just prevents the client from lighting a bomb from the player's inventory if they are holding something. However, the client now needs to wait for the round trip to the server and back before they can throw the bomb. Ideally, the client should tell the server to try to light or throw a bomb and the server decides, but that would require a bit of refactoring. At least this is an improvement from before.

Resolves #1037.

## Steps to Test or Reproduce

1. Run and join a dedicated server
2. Run [clumsy](https://jagt.github.io/clumsy/) with 200ms lag
3. Attempt to light two bombs in quick succession

### DIFF
--- a/Entities/Characters/Knight/KnightLogic.as
+++ b/Entities/Characters/Knight/KnightLogic.as
@@ -1074,6 +1074,9 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
 {
 	if (cmd == this.getCommandID("get bomb"))
 	{
+		if (this.getCarriedBlob() !is null)
+			return;
+
 		const u8 bombType = params.read_u8();
 		if (bombType >= bombTypeNames.length)
 			return;


terracraft321:
Nice. Unrelated but in the other case regarding the refactoring, how should the server decide about the players send action?

eps0003:
Instead of the client checking if it's holding something or needs to light a new bomb, the server should do that. Pretty much move all the checks the client is doing onto the server instead.

Vam-Jam (approved):

PR #1041, opened 2021-05-04
Improved pickup with bad ping

eps0003:
## Status

## Description

The client no longer needs to wait for the player to pick up items to throw them. If you press C twice in quick succession while near an item with bad ping, you pick up and throw the item even though the item is yet to be picked up on the second press.

The downside to this is now items are thrown when C is released instead of on press. This only applies to C and not other controls that throw items like space.

This PR also moves some pickup methods into a new file called PickupCommon.as and throw methods into ThrowCommon.as so the code can potentially be reused.

## Steps to Test or Reproduce

1. Run and join a dedicated server
2. Run [clumsy](https://jagt.github.io/clumsy/) with 200ms lag
3. Attempt to pickup and throw items fast
4. Notice how you don't need to wait for items to visibly be in your hand to throw it

### DIFF
--- a/Entities/Common/Controls/PickupCommon.as
+++ b/Entities/Common/Controls/PickupCommon.as
@@ -1,0 +1,367 @@
+#include "StandardControlsCommon.as"
+
+void GatherPickupBlobs(CBlob@ this)
+{
+	CBlob@[]@ pickupBlobs;
+	this.get("pickup blobs", @pickupBlobs);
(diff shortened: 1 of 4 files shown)


AsuMagic:
I only just noticed this conflicts now because https://github.com/transhumandesign/kag-base/pull/1327 was merged.

eps0003:
Abandoning.

PR #1048, opened 2021-05-10
Dead enemy bodies open team bridges

eps0003:
## Status

## Description

A very simple change that allows dead enemy bodies to open team bridges. This can allow players to throw bodies onto enemy bridges in hopes of opening them underneath an enemy.

## Steps to Test or Reproduce

1. Place team bridges
2. Swap team
3. Spike yourself to death
4. Throw your dead corpse onto the bridges

### DIFF
--- a/Entities/Structures/Platform/Bridge.as
+++ b/Entities/Structures/Platform/Bridge.as
@@ -177,7 +177,7 @@ bool canOpen(CBlob@ this, CBlob@ blob)
 {
 	if (this.getTeamNum() != blob.getTeamNum()
 		&& blob.getShape().getConsts().collidable
-		&& (blob.hasTag("player") || blob.hasTag("vehicle")))
+		&& (blob.hasTag("player") || blob.hasTag("dead") || blob.hasTag("vehicle")))
 	{
 		return true;
 
--- a/Entities/Structures/Platform/Bridge.as
+++ b/Entities/Structures/Platform/Bridge.as
@@ -177,7 +177,7 @@ bool canOpen(CBlob@ this, CBlob@ blob)
 {
 	if (this.getTeamNum() != blob.getTeamNum()
 		&& blob.getShape().getConsts().collidable
-		&& (blob.hasTag("player") || blob.hasTag("dead") || blob.hasTag("vehicle")))
+		&& (blob.hasTag("player") || blob.hasTag("dead player") || blob.hasTag("vehicle")))
 	{
 		return true;
 
--- a/Entities/Characters/Scripts/RunnerDeath.as
+++ b/Entities/Characters/Scripts/RunnerDeath.as
@@ -91,6 +91,7 @@ f32 onHit(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@ hitt
 		// disable tags
 		this.Untag("shielding");
 		this.Untag("player");
+		this.Tag("dead player");
 		this.getShape().getVars().isladder = false;
 		this.getShape().getVars().onladder = false;
 		this.getShape().checkCollisionsAgain = true;


Vam-Jam (approved):

Entities/Structures/Platform/Bridge.as
```diff
@@ -170,7 +170,7 @@ bool canOpen(CBlob@ this, CBlob@ blob)
 {
 	if (this.getTeamNum() != blob.getTeamNum()
 		&& blob.getShape().getConsts().collidable
-		&& (blob.hasTag("player") || blob.hasTag("vehicle")))
+		&& (blob.hasTag("player") || blob.hasTag("dead") || blob.hasTag("vehicle")))
```
AsuMagic, line 173: Cannot check easily right now, but are you 100% sure that the `dead` tag only ever applies to player blobs?

eps0003: I've never seen it used for anything other than players.

CatWiz: Dead fishies have it

eps0003: Does anyone know of a sure-fire way to check for dead bodies? I've always just used this check.

AsuMagic (changes requested):
This does seem to open team bridges because of fishes.

> Does anyone know of a sure-fire way to check for dead bodies? I've always just used this check.

Honestly I don't know and from glancing at the properties (*wink wink* staging blob debug) there is none that is immediately obvious. Maybe that should be added through a tag? knight/archer/builder scripts already tag `"player"` in each of their individual scripts already anyway...

Entities/Structures/Platform/Bridge.as
mugg91: Since fishies are team 255, you can do this.

```
        if ((this.getTeamNum() != blob.getTeamNum() && blob.getTeamNum() != 255)
		&& blob.getShape().getConsts().collidable
		&& (blob.hasTag("player") || blob.hasTag("dead") || blob.hasTag("vehicle")))
	{
		print(blob.getTeamNum() + " " + blob.getName());
		return true;
	}
	return false;
```

And if [#1251](https://github.com/transhumandesign/kag-base/pull/1251) gets merged, which adds [TeamChecking.as](https://github.com/transhumandesign/kag-base/blob/130db67cdc411575bf722d5e2d5177faaa19fe96/Entities/Common/Scripts/TeamChecking.as), you can do this:

```
#include "TeamChecking.as"

// (...)

        if (isDifferentTeam(blob, this)
		&& blob.getShape().getConsts().collidable
		&& (blob.hasTag("player") || blob.hasTag("dead") || blob.hasTag("vehicle")))
	{
		print(blob.getTeamNum() + " " + blob.getName());
		return true;
	}
	return false;
```

Migrants, Princess and Trader will not have normal collision with Bridge and fall through in the current build. 
[#1283](https://github.com/transhumandesign/kag-base/pull/1283) may change that. But that is less of a problem since standalone Migrants, Princess and Trader won't show up in normal gameplay.

eps0003: Player blobs are now tagged as `"dead player"` when they die, and this tag is used in the check instead of `"dead"`.

AsuMagic (approved):

mugg91:
Ok, you added a new tag `"dead player"`. 
Was my other suggestion not sufficient? With that, a new tag and clutter could be avoided.

I was considering if `this.Tag("dead player");` should be placed further up, after line 23 `this.Tag("dead");`, instead of in the section that says `// disable tags`. 

Assuming we keep this new tag, you may consider doing changes to other files:

**PowerUp.as**

Change `if (blob !is null && blob.hasTag("player") && !blob.hasTag("dead"))` 
to `if (blob !is null && !blob.hasTag("dead player"))`.

**VehicleConvert.as**

 Change `if (b !is this && b.hasTag("player") && !b.hasTag("dead"))`
to `if (b !is this && !b.hasTag("dead player"))`.

**Hall.as**

Change `if (b !is this && b.hasTag("player") && !b.hasTag("dead") && !b.hasTag("migrant"))`
to `if (b !is this && !b.hasTag("dead player") && !b.hasTag("migrant"))`.

eps0003:
> Was my other suggestion not sufficient? With that, a new tag and clutter could be avoided.

Using a tag like this is more reliable and straightforward than checking for teams as you have suggested.

> Assuming we keep this new tag, you may consider doing changes to other files

Feel free to do this yourself if you feel it's necessary.

PR #1049, opened 2021-05-10
Players playing on duplicate instances of KAG cannot vote

### LINKED ISSUE #655: Prevent accounts with ~ from having the option to vote
**Description**

When you join a server with your username with 2 different kag instances, you get a ~2 on the 2nd user, ~3 and so on for the third. 
This can be abused to force vote to kick people out the server or force a map to change

**Additional context**
Its like just the admin username exploit with joining with 1 accounts ages ago, nobody has used it yet, but its safer to prevent it.

eps0003:
## Status

## Description

Players with ~ in their username cannot start votes or contribute to existing votes. I also moved the "Voting for surrender is not available as a spectator" reason above most other reasons for not being able to vote.

Resolves #655

## Steps to Test or Reproduce

1. Join a server with two clients with the same username
2. Attempt to vote with the client who joined last

### DIFF
--- a/Rules/CommonScripts/DefaultVotes.as
+++ b/Rules/CommonScripts/DefaultVotes.as
@@ -463,7 +463,7 @@ void onMainMenuCreated(CRules@ this, CContextMenu@ menu)
 
 	bool can_skip_wait = getSecurity().checkAccess_Feature(me, "skip_votewait");
 
-	bool duplicatePlayer = me.getUsername().find("~") > -1;
+	bool duplicatePlayer = isDuplicatePlayer(me);
 
 	//kick menu
 	if (getSecurity().checkAccess_Feature(me, "mark_player"))
--- a/Rules/CommonScripts/DefaultVotes.as
+++ b/Rules/CommonScripts/DefaultVotes.as
@@ -476,7 +476,9 @@ void onMainMenuCreated(CRules@ this, CContextMenu@ menu)
 				getTranslatedString(
 					"Voting to kick a player\n" +
 					"is not allowed when playing\n" +
-					"with a duplicate instance of KAG"
+					"with a duplicate instance of KAG.\n\n" +
+					"Try rejoining the server\n" +
+					"if this was unintentional."
 				)
 			);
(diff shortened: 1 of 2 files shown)


Rules/CommonScripts/DefaultVotes.as
```diff
@@ -578,7 +600,19 @@ void onMainMenuCreated(CRules@ this, CContextMenu@ menu)
...
-		if (g_lastNextmapCounter < 60 * getTicksASecond()*required_minutes_nextmap
+		if (duplicatePlayer)
+		{
+			Menu::addInfoBox(
+				mapmenu,
+				getTranslatedString("Can't Start Vote"),
+				getTranslatedString(
+					"Voting for next map\n" +
+					"is not allowed when playing\n" +
+					"with a duplicate instance of KAG"
```
Vam-Jam, line 611: Might be worth telling the user to rejoin the server. Most the time people rejoin a server after they lag out (before their last player gets removed), causing a ~ to appear. In that case those people might not be aware they are a duplicate.

Rules/CommonScripts/DefaultVotes.as
```diff
@@ -455,10 +463,24 @@ void onMainMenuCreated(CRules@ this, CContextMenu@ menu)
 
 	bool can_skip_wait = getSecurity().checkAccess_Feature(me, "skip_votewait");
 
+	bool duplicatePlayer = me.getUsername().find("~") > -1;
```
Vam-Jam, line 466: replace with ``bool duplicatePlayer = VoteCheckFunctor::PlayerCanVote(me);``

eps0003: Can't do that. That line isn't within a class that inherits from `VoteCheckFunctor`.
image

Vam-Jam: Ah that makes sense :+1:

Vam-Jam (approved):

AsuMagic (approved):

PR #1051, opened 2021-05-18
Custom head auto uploading

Vam-Jam:
## Description:
Patreon user's can use their heads without having to wait for an update.

## Status:

## What does this do:
- When a player joins, server sends them a cmd telling them to send their head (if they are allowed to)
- Player searches for `CustomHead.png` and checks the size (64x16)
- Player will send alpha of every pixel (48x16) (note: we do not need to do full 64x16 since last frame is not used)
  - if pixel > 50 alpha (needs tweaking), client also sends rgb
- Server keeps a copy of the head, and will send it to every client (and new client that joins)

Notes: 
- Players who have won a permanent head without patreon can still use their custom heads (However the head still needs to be put inside CustomHeads folder like before (so it requires an update))

Needs doing:
- Validate that it works with multiple people

Future stuff:
- Cache player heads on client(?)
- Put a template CustomHead.png somewhere in the base folder for user's to write over
- Toggle to force user heads to default locally (binded to f6 or something) and save to cache folder
- Admin toggle to force player's head to default (binded to f6 or something)

### DIFF
--- a/Entities/Characters/Scripts/RunnerHead.as
+++ b/Entities/Characters/Scripts/RunnerHead.as
@@ -124,6 +124,9 @@ CSpriteLayer@ LoadHead(CSprite@ this, int headIndex)
 	int headsPackIndex = getHeadsPackIndex(headIndex);
 	HeadsPack@ pack = getHeadsPackByIndex(headsPackIndex);
 	string texture_file = pack.filename;
+	
+	// Custom head texture
+	CSpriteLayer@ head = null;
 
 	bool override_frame = false;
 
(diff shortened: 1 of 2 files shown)


Vam-Jam:
Taking this out of WIP. It work's fine from my testing but I do not plan to add any of the 'future features' any time soon.

This PR should work, but needs testing with multiple people.

Gingerbeard5773:
This is a pretty nice idea, so I tested it for a while. Despite some editing from me to get the code to work, I couldn't get the headstream to correctly de-serialize, which causes visual issues (see image below). There's also a few other networking problems that I'm surprised vam didn't catch. Vamist pls.

some other problems as well
1) The heads don't show on the scoreboard.
2) The heads don't change team colors.

image

Vam-Jam:
Worked a year ago lol, and yea, the code is pretty bad looking at it now. 
If this is a feature that people are interested in, I will do a new PR with a rewrite and what ever featured are requested, otherwise consider this abandoned.

asumagic:
Closing given the abandoned status at the moment

PR #1058, opened 2021-05-26
Admins can view inventories of players and other things while in spectator team

CatWiz:
## Status

- 
## Description

New pr with new branch because i fucked up the previous one. Fixed drawing everything only in rows of 2, script now removes itself if player is not admin.

Title. Hold left control and aim at a player or anything that has an inventory to see what's inside.
Works only in spectator team because i think it would be too big of a gameplay advantage to give to just admins, and it can be used for trolling too easily to give it to everyone.

### DIFF
--- a/Rules/CommonScripts/AdminInventoryView.as
+++ b/Rules/CommonScripts/AdminInventoryView.as
@@ -1,6 +1,8 @@
 #define CLIENT_ONLY
 
-void onRender(CRules@ this)
+CBlob@ toDrawInventory = null;
+
+void onTick(CRules@ this) // pick which blob's inventory will be drawn_blobs only on game update
 {
 	CMap@ map = getMap();
 	if (map is null) return;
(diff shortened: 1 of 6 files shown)


Rules/CommonScripts/AdminInventoryView.as
```diff
@@ -0,0 +1,79 @@
...
+			{
+				CBlob@ blob = targets[i];
+				if (blob.getInventory() !is null)
+				{
+						DrawInventoryOfBlob(blob);
+						break;
+				}
+			}
+		}	
+	}
```
eps0003, line 39: The code that gets the blob should be moved to `onTick()`. `onRender()` should only include the code for rendering the inventory of that blob.

Rules/CommonScripts/AdminInventoryView.as
```diff
@@ -0,0 +1,79 @@
...
+}
+
+void DrawInventoryOfBlob(CBlob@ this)
+{
+	CInventory@ inv = this.getInventory();
+	string[] drawn;
+	Vec2f tl = this.getInterpolatedScreenPos();
+	Vec2f offset (-40, -120);
+	u8 j = 0;
```
eps0003, line 48: This variable should have a better name. It's not clear what this variable is for at first glance.

eps0003 (dismissed):
Apologies for the late review.

Rules/CommonScripts/AdminInventoryView.as
CatWiz: done

Mazey (approved):
Seems good.

mugg91:
Changed files says "AdminInvetoryView.as;" instead of "AdminInventoryView.as;"...

PR #1063, opened 2021-06-22
Bucket: prevent accidental splashing when clicking when inventory is open

Nananas:
## Status

## Description

This should fix accidentally splashing when the inventory is open, and clicking. This happens for example when trying to put away or take a bucket from inventory by clicking.

## Steps to Test or Reproduce

1. Get and hold a full bucket
2. Open your inventory
3. Try putting (or taking out) the bucket in the inventory by clicking
4. The bucket should now not splash water

### DIFF
--- a/Entities/Items/Bucket/Bucket.as
+++ b/Entities/Items/Bucket/Bucket.as
@@ -51,7 +51,11 @@ void onTick(CBlob@ this)
 		{
 			this.set_u8("water_delay", water_delay - 1);
 		}
-		else if (point.getOccupied() !is null && point.getOccupied().isMyPlayer() && point.getOccupied().isKeyJustPressed(key_action1) && !this.isInWater())
+		else if (point.getOccupied() !is null && 
+				 point.getOccupied().isMyPlayer() && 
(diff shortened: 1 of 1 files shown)


Entities/Items/Bucket/Bucket.as
```diff
@@ -51,7 +51,11 @@ void onTick(CBlob@ this)
 		{
 			this.set_u8("water_delay", water_delay - 1);
 		}
-		else if (point.getOccupied() !is null && point.getOccupied().isMyPlayer() && point.getOccupied().isKeyJustPressed(key_action1) && !this.isInWater())
+		else if (point.getOccupied() !is null && 
+				 point.getOccupied().isMyPlayer() && 
+				 point.getOccupied().isKeyJustPressed(key_action1) && 
+				 !this.isInWater() &&
+				 !point.getOccupied().isKeyPressed(key_inventory)) // prevent splash when doing stuff with inventory
```
eps0003, line 58: It might be worth getting the occupied blob once and storing it in a variable instead of doing it four times.

Nananas: Done

eps0003 (approved):

PR #1076, opened 2021-07-21
[fix] Game freezing while trying to join a server

Vam-Jam:
# How does this happen?
This happens when a tree is covered by a tile. This does not happen often in vanilla (but has done in the past), but mostly happens on modded servers. When a client joins the game, they get stuck in a while loop trying to call DoGrow, except it wont work because the tree refuses to grow due to the tile blocking it. 

# How does this fix it?
Client will attempt to call do grow a finite amount of times before stopping. Ideally we add some custom logic in DoGrow for when your client is joining to ignore tiles, but for such a rare edge case I feel like this is fitting enough.

### DIFF
--- a/Entities/Natural/Trees/TreeSync.as
+++ b/Entities/Natural/Trees/TreeSync.as
@@ -61,9 +61,9 @@ void InitTree(CBlob@ this, TreeVars@ vars)
 		// recreate growth
 		u8 grown_times = this.get_u8("grown_times");
 
-		while (vars.grown_times < grown_times)
+		for (int a = vars.grown_times; a < grown_times; a++)
 		{
-			DoGrow(this, vars);
+			DoGrow(this, var);
 		}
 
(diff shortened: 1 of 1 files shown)


eps0003:
How were you able to test this?

Vam-Jam:
> How were you able to test this?

If you remove the code that forces tree's to instant grow on map load, you should be able to repeat the issue.

This code is used in TC and other mods where the bug happens. It was first found a few years ago when pirate-rob had this issue in his mod.

eps0003:
> If you remove the code that forces tree's to instant grow on map load, you should be able to repeat the issue.

I am unable to reproduce. I commented out the code in TreeSync.as lines 45-74 and edited the map to put a pillar of dirt above a tree.

image

When you say it "happens when a tree is covered by a tile", do you mean the seed needs to be inside the block or grow up into a block?

AsuMagic (approved):

PR #1083, opened 2021-07-27
Knights cant slash picked up items

CatWiz:
## Status

## Description

Title. Also added a tag to prevent this for certain blobs and added it to keg has it so it's still possible to hit them off enemies' backs

### DIFF
--- a/Entities/Characters/Knight/KnightLogic.as
+++ b/Entities/Characters/Knight/KnightLogic.as
@@ -1223,7 +1223,8 @@ void DoAttack(CBlob@ this, f32 damage, f32 aimangle, f32 arcdegrees, u8 type, in
 			if (b !is null && !dontHitMore) // blob
 			{
 				if (b.hasTag("ignore sword")) continue;
-
+				
+				// don't hit picked up items
 				CAttachment@ att = b.getAttachments();
 				if (att !is null)
 				{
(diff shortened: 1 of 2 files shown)


Entities/Characters/Knight/KnightLogic.as
```diff
@@ -1223,6 +1223,15 @@ void DoAttack(CBlob@ this, f32 damage, f32 aimangle, f32 arcdegrees, u8 type, in
 			if (b !is null && !dontHitMore) // blob
 			{
 				if (b.hasTag("ignore sword")) continue;
+				
+				// don't hit picked up items
+				CAttachment@ att = b.getAttachments();
+				if (att !is null)
+				{
+					AttachmentPoint@ point = att.getAttachmentPointByName("PICKUP");
+					if (point !is null && !point.socket &&
```
eps0003, line 1232: Why do you need to check if the point isn't a socket? I barely understand how attachment points work.

CatWiz: Because if I don't, any player who's holding anything will count as "attached to pickup point" and will be unslashable

eps0003: I still don't really understand. I tested this and it works exactly the same as your code. I only tested this with a boulder and a keg.
```angelscript
if (b.isAttachedToPoint("PICKUP") && !b.hasTag("slash_while_in_hand")) continue;
```

CatWiz: Were you able to slash the player/bot holding the item? Did you test it on a server (shouldn't affect anything but still)? The issue i had was if a player is holding mat_wood, for example, it was be impossible to slash them

eps0003: Yep you're right.

eps0003 (approved):

AsuMagic:
Er, I didn't notice that those changes were merged in as part of #1102 already. Closing.

PR #1084, opened 2021-07-27
Ballista crates drop gold

CatWiz:
## Status

## Description

Ballista crates drop gold if they are destroyed before they get unpacked (for example, if they fall into a saw)


eps0003 (dismissed):
How did you miss the typo I fixed?

PR #1085, opened 2021-07-27
Fix RegrowPlants not working

bunniewormy:
## Status

(pick one)

## Description

next_check_time was being set to gametime, so in next matches it didnt work until the match time reached the gametime it was last set at


bunniewormy:
closing pr because cat found 69 other issues

eps0003 (approved):

PR #1086, opened 2021-07-30
Knight crouch anims

CatWiz:
## Status

## Description

Adds additional animations for knight: crouching+shielding and crouching+charging a slash. No slash/jab+crouch anims because crouching is impossible in those states. Unused carry anims for caped knight were moved to the bottom of the spritesheets.


eps0003:
Why are his legs so derpy when shielding straight up? Why aren't his legs positioned the same as when shielding to the side?

image

CatWiz:
Those sprites aren't mine (original knightmale sprite was made by dome, then adapted to other spritesheets by Skemonde. i really should've credited them when making the pr), but i find it funny. Also it's because knight is already sitting in other crouch anims when shielding up without crouch

eps0003 (approved):

AsuMagic (approved):

mugg91:
Shield-crouching doesn't look like you are humping something when holding stuff. For better or worse. But I'm just saying it means an old meme is gone.

PR #1087, opened 2021-07-31
Builder can now hit team crate

bunniewormy:
## Status

(pick one)

## Description

builder can now hit crate, previously couldn't which was quite annoying sometimes


eps0003 (approved):
There's probably a better way of doing this so we don't need to keep hardcoding them in here, but this will do for now.

PR #1101, opened 2021-09-19
More server-side checks for correct blob/block placement

bunniewormy:
## Status

## Description

- Removed the warn prints (they were there for finding hackers more easily, but it just prints a lot in every single match)
- Added a check to prevent blob-inside-blob placement
- Added a check to prevent placing the same tile in the same tilespace a few times in a row when lagging (eg. you could try to place a stone backwall on one tilespace 4 times and lose 8 stone)
- Modified a check to prevent blob-inside-block and block-inside-blob placement

## Steps to Test or Reproduce

Get a 333 ping friend to join a server with you

Make them spam left mouse button while having a stone block/door selected in the spot of a door you're about to destroy; place a block or a door in the same spot after destroying it. Currently it'd result in a door and a block occupying the same spot, but it's fixed in this PR.

Go on a 333 ping server and try to place 4 stone backwalls in the same spot; you will get 8 stone taken away from you. Fixed with this PR

### DIFF
--- a/Entities/Common/Building/BlobPlacement.as
+++ b/Entities/Common/Building/BlobPlacement.as
@@ -58,7 +58,7 @@ bool serverBlobCheck(CBlob@ blob, CBlob@ blobToPlace, Vec2f cursorPos)
 	CMap@ map = getMap();
 	Tile backtile = map.getTile(cursorPos);
 
-	if (map.isTileBedrock(backtile.type) || map.isTileSolid(backtile.type) && map.isTileGroundStuff(backtile.type)) 
+	if (map.isTileBedrock(backtile.type) || map.isTileSolid(backtile.type))
 		return false;
 
 	// Make sure we actually have support at our cursor pos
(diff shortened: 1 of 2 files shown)


Entities/Common/Building/BlobPlacement.as
```diff
@@ -78,6 +78,17 @@ bool serverBlobCheck(CBlob@ blob, CBlob@ blobToPlace, Vec2f cursorPos)
 			return false;
 	}
 
+	CBlob @blobAtPos = map.getBlobAtPosition(cursorPos + Vec2f(1, 1));
+
+	// Are we trying to place a blob on a door/ladder/platform/bridge (usually due to lag)?
+	if (blobAtPos !is null && (
```
AsuMagic, line 84: this should probably check for whether there is any solid tile/blob tile instead of this?

Entities/Common/Building/BlockPlacement.as
```diff
@@ -88,6 +79,25 @@ bool serverTileCheck(CBlob@ blob, u8 tileIndex, Vec2f cursorPos)
...
+	BuildBlock @blockToPlace = getBlockByIndex(blob, tileIndex);
+	// Are we trying to place a tile on the same tile (usually due to lag)?
+	if (backtile.type == blockToPlace.tile)
+	{
+		return false;
+	}
+
+	CBlob @blobAtPos = map.getBlobAtPosition(cursorPos + Vec2f(1, 1));
+
+	// Are we trying to place a solid tile on a door/ladder/platform/bridge (usually due to lag)?
```
AsuMagic, line 91: duplicate logic with BlobPlacement

Entities/Common/Building/BlobPlacement.as
bunniewormy: fixed, added a function to placementcommon.as

PR #1102, opened 2021-09-24
Raycasted knight slashes (and jabs)

CatWiz:
## Status

## Description

Knight's DoAttack now does a getHitInfosFromRay to every hittable blob it finds (starting from the furthest ones). This raycast hits every hittable blob in its way and stops if it runs into a tile or a "large" blob (usually a door). 
This allows knights to hit through 1 block holes in doors, removing the mechanic or doors "eating" slashes. In addition to that, it allows knights to hit multiple doors with one slash without doing 360 cursor rotation, and it removes the ability to hit multiple logs with a single slash by doing same type of cursor rotation.

## Steps to Test or Reproduce

- Go knight
- Try hitting people in any of door formations that are shown [here](https://deynarde.github.io/kag-knight-guide/playing-ctf.html#hitting-enemies-at-specific-door-angles) and [here](https://deynarde.github.io/kag-builder-guide/builder-iv-offensive-playstyle.html#using-doors-at-certain-corner-angles-to-damage-knights) normally
- Succeed

## Additional notes

 - Being able to hit multiple doors is a huge boost to the speed at which knights can break them, and I think it is a good thing.
 - This is worse than old knight attack logic performance-wise, however we did not notice any lag caused by slashing or overall increase in lag on servers that had this change (few 6v6 matches on EU Captains and 8v8-12v12 matches on CTF Rebuild)
 - Changes same part of code as #1083, I'll probably just merge it into this branch later. **- DONE**


bunniewormy:
Mildly related to PR - opinions of 39 people regarding doors eating slashes, also i wrote 1k word comment regarding the entire thing: https://docs.google.com/document/d/1XO3f6Rm7zvzleq-l0-rRVudEPfPyu-ZS7MZ2rEbNsMA/edit?usp=sharing

AsuMagic (approved):

PR #1122, opened 2021-10-31
Allow shooting hearts with arrows like food

### LINKED ISSUE #1017: Make hearts behave like food and also make  knights hit food like archers

Currently shooting burger or grain is just a funny feature that almost never comes into play... even though it is a very interesting feature

**Proposed Change**

To make it more of an consistent feature i think it should apply to hearts too and maybe would make sense to let knights be able to slash food too so that it flies a short (not als long as with arrows) distance. 

**Additional idea**

An alternative and maybe interesting way  to shoot food would be to make arrows "spike" food and the food would travel the same path as the arrow... and would stick on a wall. this would not only be cool but also make it way easier to aim food to your teammates.

laws65:
## Status

## Description
Makes it so that you can shoot hearts as an archer like you can with food. Resolves #1017

## Steps to Test or Reproduce
- Spawn in a heart
- Throw and shoot it

### DIFF
--- a/Entities/Items/Heart/Heart.as
+++ b/Entities/Items/Heart/Heart.as
@@ -3,6 +3,5 @@ void onInit(CBlob@ this)
 	this.set_string("eat sound", "/Heart.ogg");
 	this.getCurrentScript().runFlags |= Script::remove_after_this;
 	this.server_SetTimeToDie(40);
-	this.Tag("ignore_arrow");
 	this.Tag("ignore_saw");
 }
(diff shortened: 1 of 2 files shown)


eps0003:
This feature was added at one point but removed to avoid archers accidentally shooting dropped hearts.

AsuMagic:
> This feature was added at one point but removed to avoid archers accidentally shooting dropped hearts.

Yeah, I don't think this will be merged for now. On the field this is very likely to end up into accidentally shooting hearts. That rarely happens with food. Plus a heart heals only one heart it doesn't feel nearly as useful as the same mechanic for burgers.

PR #1146, opened 2021-11-09
[fixed] frames added to door animation unrestricted

Mazey:
## Status

## Description

Fixed door closing animation.

When a door closes, it will play the following frames: 2 -> 1 -> 0. 0 being the frame of an undamaged closed door.

When a door takes damage (frame 4 to 6, 6 being the final frame before the door is destroyed), the animation is supposed to change to: 2 -> 1 -> 4(/5/6). Currently, rather than replacing the final door frame, it will only add frames to the animation. Right now on production a fully damaged door closing will play the following animation: 2 -> 1 -> 0 -> 4 -> 5 -> 6.

After this PR, it will simply play 2 -> 1 -> 6 when the door is nearly destroyed.

Additionally, moved animation stuff to onHealthChange, as there is no reason to do this in onHit, this means calculating "future health" is no longer necessary, this is in line with other blob blocks (thanks bunnie)

### DIFF
--- a/Entities/Structures/Door/SwingDoor.as
+++ b/Entities/Structures/Door/SwingDoor.as
@@ -187,6 +187,11 @@ f32 onHit(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@ hitt
 	if (customData == Hitters::bomb)
 		damage *= 1.3f;
 
+	return damage;
+}
+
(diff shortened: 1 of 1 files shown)


Vam-Jam (approved):

AsuMagic (approved):
Right, I had this tested and got sidetracked by an unrelated crash issue at the time. Seemed to work well and the fix is sensible.

PR #1148, opened 2021-11-10
[CTF Rebuild] Water changes, water arrow nerf

bunniewormy:
## Status

(pick one)

## Description

[changed] water arrows come in stack of 1 instead of 2
[changed] new cost of water arrow: 15 instead of 10
[added] knockback from water bombs and water arrows
[changed] splash from water buckets now does a bit of knockback

### DIFF
--- a/Entities/Materials/Ammunition/MaterialWaterArrow.cfg
+++ b/Entities/Materials/Ammunition/MaterialWaterArrow.cfg
@@ -19,7 +19,7 @@ $sprite_animation_start                = *start*
 $sprite_animation_default_name         = default
 u16 sprite_animation_default_time      = 0
 u8 sprite_animation_default_loop       = 0
-@u16 sprite_animation_default_frames   = 20; 28;
+@u16 sprite_animation_default_frames   = 20;
 
 $sprite_animation_end                  = *end*
 
--- a/Entities/Common/Attacks/Hitters.as
+++ b/Entities/Common/Attacks/Hitters.as
@@ -73,3 +73,8 @@ bool isExplosionHitter(u8 type)
 {
 	return type == Hitters::bomb || type == Hitters::explosion || type == Hitters::mine || type == Hitters::bomb_arrow;
 }
+
+bool isWaterHitter(u8 type)
+{
+	return type == Hitters::water || type == Hitters::water_stun_force || type == Hitters::water_stun;
+}
--- a/Entities/Common/Attacks/KnockBack.as
+++ b/Entities/Common/Attacks/KnockBack.as
@@ -46,7 +46,7 @@ f32 onHit(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@ hitt
 
 	Vec2f f(x_side, y_side);
 
-	if (damage > 0.125f)
+	if (damage > 0.125f || isWaterHitter(customData))
 	{
 		this.AddForce(f * 40.0f * scale * Maths::Log(2.0f * (10.0f + (damage * 2.0f))));
 	}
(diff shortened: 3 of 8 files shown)


bunniewormy:
Changed the cost from 25 to 15 due to this probably being too significant of a nerf. 
Old: 20 coins for stack of 2 arrows
This PR: 15 coins for stack of 1 arrows

Entities/Materials/Ammunition/MaterialWaterArrow.cfg
```diff
@@ -19,7 +19,7 @@ $sprite_animation_start                = *start*
 $sprite_animation_default_name         = default
 u16 sprite_animation_default_time      = 0
 u8 sprite_animation_default_loop       = 0
-@u16 sprite_animation_default_frames   = 20; 28;
+@u16 sprite_animation_default_frames   = 20;
```
AsuMagic, line 22: Was this done because a full stack of 1 would show as the 2 water arrows sprite?  
If not this would be fine to keep so that it is easier for mods to change the max amount imo

AsuMagic (approved):

Entities/Materials/Ammunition/MaterialWaterArrow.cfg
bunniewormy: yeah, full stack of 1 shows 2 water arrows sprite without this

mugg91:
I haven't looked into it much but I noticed `onThisAddToInventory()` in Arrow.as has a return for bomb arrows:

```	
if (arrowType == ArrowType::bomb)
	{
		return;
	}
```
presumably because bomb arrows won't be merged together due to being a stack of 1.
Since water arrows would now also be in stacks of 1, maybe it is necessary to add a return for water arrows, too.

bunniewormy:
> I haven't looked into it much but I noticed `onThisAddToInventory()` in Arrow.as has a return for bomb arrows:
> 
> ```
> if (arrowType == ArrowType::bomb)
> 	{
> 		return;
> 	}
> ```
> 
> presumably because bomb arrows won't be merged together due to being a stack of 1. Since water arrows would now also be in stacks of 1, maybe it is necessary to add a return for water arrows, too.

no, as far as i know that's legacy code that is not used anywhere. the arrow blob is different from the mat_arrows, mat_waterarrows, mat_bombarrows blobs.

`void onThisAddToInventory(CBlob@ this, CBlob@ inventoryBlob)
{
	if (!getNet().isServer())
	{
		return;
	}

	const u8 arrowType = this.get_u8("arrow type");
	if (arrowType == ArrowType::bomb)
	{
		return;
	}

	// merge arrow into mat_arrows

	for (int i = 0; i < inventoryBlob.getInventory().getItemsCount(); i++)
	{
		CBlob @blob = inventoryBlob.getInventory().getItem(i);

		if (blob !is this && blob.getName() == "mat_arrows")
		{
			blob.server_SetQuantity(blob.getQuantity() + 1);
			this.server_Die();
			return;
		}
	}

	// mat_arrows not found
	// make arrow into mat_arrows
	CBlob @mat = server_CreateBlob("mat_arrows");

	if (mat !is null)
	{
		inventoryBlob.server_PutInInventory(mat);
		mat.server_SetQuantity(1);
		this.server_Die();
	}
}`

the arrow blob should never be put in someone's inventory; arrows don't even have an attachment slot, and the fletchArrow function in ArcherLogic.as creates new mat_arrows.

what this code does is give you +1 arrow (normal arrow) if a water arrow (not mat_waterarrows) or a fire arrow enters your inventory. but it shouldn't ever happen and i think that code is just some very old hacky workaround. anyway, that's all completely unrelated to this PR

PR #1150, opened 2021-11-10
[CTF Rebuild] Coin changes: more coin gain from combat. 600 coin limit. Less coin loss on death

bunniewormy:
## Status

(pick one)

## Description

[change] more coin gain from combat (around 15-20% more total compared to current)
[change] less coin loss on death (15% instead of 20)
[change] increased coin gain from killing people as a builder (around 30% compared to new values for other classes)
[change] less coin gain from hitting siege
[removed] remove coin loss from teamkill
[added] added a 600 coin limit (automatically disabled on Christmas holiday for festivity)


MonkeyFeats:
Removing coin loss from team kills is good.
Less coins from seige is good.

Builders already get more money than everyone else. Another example of buffing builders...
What is the purpose of a 600 coin limit?

bunniewormy:
> 
> 
> Removing coin loss from team kills is good. Less coins from siege is good.
> 
> Builders already get more money than everyone else. Another example of buffing builders... What is the purpose of a 600 coin limit?

The purpose of a 600 coin limit is because currently builders are able to reach ridiculous amount of money by simply building and not going outside of the base at all. You can farm 600+ coins if you spend 15 minutes on only digging stone, resupplying and base expanding which is pretty bs. I believe builders should instead be able to get coins from combat and playing offensively instead of having to rely only on building blocks for their income.

And yes, the point of this change is to buff offensive builders and increase the amount of items they're able to buy if they'll kill enemies (such as kegs or other explosives for their teammates or siege), which I believe will result in more fun, faster-paced gameplay.

Rules/CTF/Scripts/CTF_CoinCap.as
```diff
@@ -0,0 +1,17 @@
+// hacky
+#define SERVER_ONLY
+
+const int coin_cap = 600;
+
+void onTick(CRules@ this)
+{
+	// coin cap won't exist during the christmas holiday
+	if(this.hasTag("remove coincap")) return;
```
AsuMagic, line 9: This PR doesn't seem to set that tag anywhere?

AsuMagic (approved):

Rules/CTF/Scripts/CTF_CoinCap.as
AsuMagic, line 9: This PR doesn't seem to set that tag anywhere?

AsuMagic (changes requested):

Rules/CTF/Scripts/CTF_CoinCap.as
bunniewormy: derp, fixed

AsuMagic (approved):

AsuMagic (approved):

PR #1151, opened 2021-11-10
[CTF Rebuild] Faster building in build time

bunniewormy:
## Status

(pick one)

## Description

[added] building during build time is now almost 2 times faster (build delay: 7 > 4)
[fix] fixed bug when you could accidentally damage backwall behind a door when lagging or building fast

### DIFF
--- a/Entities/Characters/Builder/BuilderLogic.as
+++ b/Entities/Characters/Builder/BuilderLogic.as
@@ -142,26 +142,44 @@ bool RecdHitCommand(CBlob@ this, CBitStream@ params)
 
 	if (blobID == 0)
 	{
-		// block
 		CMap@ map = getMap();
 		if (map !is null)
 		{
-			uint16 type = map.getTile(tilepos).type;
-			if (!inNoBuildZone(map, tilepos, type))
+			if (map.getSectorAtPosition(tilepos, "no build") is null)
 			{
-				if (getNet().isServer())
-				{
-					map.server_DestroyTile(tilepos, 1.0f, this);
+				uint16 type = map.getTile(tilepos).type;
 
-					Material::fromTile(this, type, 1.0f);
+				CBlob@[] blobs_here;
+				map.getBlobsAtPosition(tilepos + Vec2f(1, 1), blobs_here);
+
+				bool no_dmg = false;
+
+				// dont dmg backwall if there's a door here
+				for(int i=0; i < blobs_here.size(); ++i)
+				{
+					CBlob@ current_blob = blobs_here[i];
+					if (current_blob !is null && (current_blob.hasTag("door") || current_blob.getName() == "bridge" || current_blob.getName() == "wooden_platform"))
+					{
+						no_dmg = true;
+					}
 				}
 
-				if (getNet().isClient())
+				if (!no_dmg)
 				{
-					if (map.isTileBedrock(type))
+					if (getNet().isServer())
 					{
-						this.getSprite().PlaySound("/metal_stone.ogg");
-						sparks(tilepos, attackVel.Angle(), 1.0f);
+						map.server_DestroyTile(tilepos, 1.0f, this);
+
+						Material::fromTile(this, type, 1.0f);
+					}
+
+					if (getNet().isClient())
+					{
+						if (map.isTileBedrock(type))
+						{
+							this.getSprite().PlaySound("/metal_stone.ogg");
+							sparks(tilepos, attackVel.Angle(), 1.0f);
+						}
 					}
 				}
 			}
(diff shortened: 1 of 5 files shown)


MonkeyFeats:
Seems like unnecessary code, just increase the time limit. 
Bases can already be massive with the current amount of time.

bunniewormy:
> 
> 
> Seems like unnecessary code, just increase the time limit. Bases can already be massive with the current amount of time.

I don't really see massive bases early-game on CTF. From what I've seen, bases are pretty shit for the first 5-10minutes of the game because people take a lot of time to build. This can result in a simple push with a drill and some knights to end the games very early, which this PR is supposed to change. The main philosophy, as I mentioned in another post, is to get bigger bases at start of the game, but with offensive builder playstyle being more viable so that you won't be able to constantly repair them for 2 hours. https://github.com/transhumandesign/kag-base/pull/1159#issuecomment-965079815

The faster building, combined with resupplies every 30 sec instead of 40 and not having to visit a builder shop, in my opinion makes build time much smoother and fun. I don't think most people enjoy build time -- it's just about making a base, once you know what you're doing there's nothing special to it and it's the same every single time -- so increasing the build time would make people even more bored.

AsuMagic:
> The faster building, combined with resupplies every 30 sec instead of 40 and not having to visit a builder shop, in my opinion makes build time much smoother and fun. I don't think most people enjoy build time -- it's just about making a base, once you know what you're doing there's nothing special to it and it's the same every single time -- so increasing the build time would make people even more bored.

Agreed, the faster block building makes the build phase a lot more bearable. It was also one of the most uncontroversial changes in voting on the CTF rebuild discord.

Extending the build phase would have the same results but would make it a lot more annoying from a gameplay perspective. It would also not have the same effects, because unless resupplies are rebalanced, the teams would end up with more materials by the end of the build phase. Early match is already often flooded with excess resources on mid, so this would be a problem. This PR probably slightly improve that situation as well.

Entities/Common/Building/BlockPlacement.as
```diff
@@ -44,7 +44,7 @@ void PlaceBlock(CBlob@ this, u8 index, Vec2f cursorPos)
 		server_TakeRequirements(inv, bc.reqs);
 		getMap().server_SetTile(cursorPos, bc.tile);
 
-		u32 delay = this.get_u32("build delay");
+		u32 delay = (getRules().hasTag("faster building") ? this.get_u32("warmup build delay") : this.get_u32("build delay"));
```
AsuMagic, line 47: This statement is duplicated in a few places, this should be refactored into a function in a common script (`getCurrentBuildSpeed()` or something)

Entities/Characters/Builder/BuilderLogic.as
```diff
@@ -142,26 +142,44 @@ bool RecdHitCommand(CBlob@ this, CBitStream@ params)
 
 	if (blobID == 0)
 	{
-		// block
 		CMap@ map = getMap();
 		if (map !is null)
 		{
-			uint16 type = map.getTile(tilepos).type;
-			if (!inNoBuildZone(map, tilepos, type))
+			if (map.getSectorAtPosition(tilepos, "no build") is null)
```
AsuMagic, line 148: What's the difference between using `inNoBuildZone` and checking for the no build map sectors here?  
And assuming this is the fix for backwall destruction part: why exactly does this bug happen and how does this fix it? Is there no better way? Might be worth documenting as comments in the code what the intents and possible sequence of events are.

AsuMagic (changes requested):

Entities/Characters/Builder/BuilderLogic.as
bunniewormy: no, this is legacy code that i forgot to change -- sections of builderlogic that used getsectoratposition were replaced in last patch and i forgot to do it here  
no idea why exactly it happens, it's just that with fast building if you hold LMB+RMB at once while placing wooden doors you will be able to damage the backwall behind them pretty consistently, and this prevents that

Entities/Common/Building/BlockPlacement.as
bunniewormy: fixed

AjaxAjaxAjax:
> Vaguely related, but if merged, it could be nice to add some sort of visual indication of the differences present during the build phase (higher resupplies and faster building), which could also serve as a better visual indication that the build phase is still going on for newbies. Maybe even just a temporary banner ("Build phase" with building icons and a short descriptive text, "Capture the flag!" when the match begins) similar to the "X team won" banner.

Would it also be possible to add text in the build menu during build time to indicate the increased speed? I think this would help greatly with new players, also just a nice QoL thing to have. 
Not sure if this has been brought up, but if resource collection speed is maintained then I could see some potential issues with players running out of stone and wood much quicker with these changes, forcing a lot of back and forth between gathering and building.

AsuMagic (approved):

AsuMagic:
FWIW, if in practice this+the build time resupply change cause teams to get excessively large towers or too much excess materials to be lost at mid early in the game, we could just shrink the CTF build time a bit. This would only lead to less frustration compared to before anyway.

PR #1152, opened 2021-11-10
[CTF Rebuild] Drill balance changes

bunniewormy:
## Status

(pick one)

## Description

[change] receives slightly more heat from hitting tiles like dirt and ore
[change] receives less heat from hitting blobs like players or doors
[change] increased max heat
[change] slower cool down 
[change] water buckets now cool 50% of drill heat instead of 70%
[change] you can't pick up overheated drills anymore (previously you could and have them instantly drop to the ground]
[change] drills now ignore bushes

### DIFF
--- a/Entities/Industry/Drill/Drill.as
+++ b/Entities/Industry/Drill/Drill.as
@@ -196,12 +196,6 @@ void onTick(CBlob@ this)
 
 		AimAtMouse(this, holder); // aim at our mouse pos
 
-		// cool faster if holder is moving
-		if (heat > 0 && holder.getShape().vellen > 0.01f && getGameTime() % 4 == 0)
-		{
-			heat--;
-		}
-
 		if (int(heat) >= heat_max - (heat_add * 1.5))
 		{
 			makeSteamPuff(this, 1.5f, 3, false);
--- a/Entities/Industry/Drill/Drill.as
+++ b/Entities/Industry/Drill/Drill.as
@@ -23,6 +23,8 @@ const u8 heat_add_constructed = 2;
 const u8 heat_add_blob = 6;
 const u8 heat_cool_amount = 2;
 
+const f32 heat_reduction_water = 0.5f;
+
 const u8 heat_cooldown_time = 8;
 const u8 heat_cooldown_time_water = u8(heat_cooldown_time / 3);
 
@@ -44,7 +46,7 @@ void onInit(CSprite@ this)
 			anim.AddFrames(frames);
 		}
 		heat.SetAnimation(anim);
-		heat.SetRelativeZ(1.0f);
+		heat.SetRelativeZ(0.1f);
 		heat.SetVisible(false);
 		heat.setRenderStyle(RenderStyle::light);
 	}
@@ -438,7 +440,7 @@ f32 onHit(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@ hitt
 
 	if (customData == Hitters::water)
 	{
-		s16 current_heat = this.get_u8(heat_prop) - heat_max * 0.5f;
+		s16 current_heat = this.get_u8(heat_prop) - heat_max * heat_reduction_water;
 		if (current_heat < 0) current_heat = 0;
 		this.set_u8(heat_prop, current_heat);
 		makeSteamPuff(this);
--- a/Entities/Industry/Drill/Drill.as
+++ b/Entities/Industry/Drill/Drill.as
@@ -14,6 +14,7 @@ const string buzz_prop = "drill timer";
 
 const string heat_prop = "drill heat";
 const u8 heat_max = 150;
+const u8 heat_drop = 140;
 const u8 high_damage_window = 40; // at how much heat before max drill deals increased damage
 
 const string last_drill_prop = "drill last active";
@@ -92,7 +93,7 @@ bool canBePutInInventory( CBlob@ this, CBlob@ inventoryBlob )
 
 bool canBePickedUp(CBlob@ this, CBlob@ byBlob)
 {
-	return (this.get_u8(heat_prop) < heat_max - heat_add * 1.5f);
+	return (this.get_u8(heat_prop) < heat_drop);
 }
 
 void onThisRemoveFromInventory( CBlob@ this, CBlob@ inventoryBlob )
@@ -198,7 +199,7 @@ void onTick(CBlob@ this)
 
 		AimAtMouse(this, holder); // aim at our mouse pos
 
-		if (int(heat) >= heat_max - (heat_add * 1.5))
+		if (int(heat) >= heat_drop)
 		{
 			makeSteamPuff(this, 1.5f, 3, false);
 			this.server_Hit(holder, holder.getPosition(), Vec2f(), 0.25f, Hitters::burn, true);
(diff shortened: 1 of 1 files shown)


MonkeyFeats:
Drills were added to assist in mining ores, that should be kept as their main use.

bunniewormy:
> Drills were added to assist in mining ores, that should be kept as their main use.

Drills are never used to mine ore and it was never their main purpose. At first they only gave 50% of mats (currently 75%), and always overheated after 3-4 seconds of mining ore. Because of that it's literally almost never worth it to use drills to mine ore rather than doing it by hand because you lose materials. FWIW, drills at first were also usable by knights and archers. They were always used either for ratting underground or offensively overground to attack the enemy base, or as a combat tool. The point of the change is to reduce their usability for ratting (hence the water bucket change and more heat from tiles like dirt), which I believe a lot of people dislike, and to make them more useful as an offensive tool for attacking the enemy base.

MonkeyFeats:
Yeah I stand corrected, maybe I just felt like they were for ore mining.

ewetcr:
[change] you can't pick up overheated drills anymore (previously you could and have them instantly drop to the ground]
this means, correct me if I'm wrong, that you can't pick an enemy drill when it overheated and fell off his/her hands?

Just to add to the conversation, drills shouldn't be more effective against players than against dirt/blocks, otherwise it's like giving a weapon to builders making it a compulsory item for any pushing builder.

CatWiz:
> you can't pick an enemy drill when it overheated and fell off his/her hands?

not immediately, only after it has cooled down a bit. both players still have chance to grab it in that situation, just need to time it. less obvious result of this change is that you cant spam pickup+lmb on overheated drill to hit something right before drill auto drops. haven't seen anyone pull this off in combat but have seen players using it to quickly dig logs when there's no saw.

and i am going to sound like a broken recorder but #1102 makes combat builders a lot weaker and more reliant on knights whether those builders use door angles intentionally or not

Entities/Industry/Drill/Drill.as
```diff
@@ -87,6 +88,11 @@ bool canBePutInInventory( CBlob@ this, CBlob@ inventoryBlob )
 	return true;
 }
 
+bool canBePickedUp(CBlob@ this, CBlob@ byBlob)
+{
+	return (this.get_u8(heat_prop) < heat_max - heat_add * 1.5f);
```
AsuMagic, line 93: This expression is copy-pasted from `onTick` (`if (int(heat) >= heat_max - (heat_add * 1.5))`), factor those out in a single global constant as is done with other values

Entities/Industry/Drill/Drill.as
```diff
@@ -43,7 +44,7 @@ void onInit(CSprite@ this)
 			anim.AddFrames(frames);
 		}
 		heat.SetAnimation(anim);
-		heat.SetRelativeZ(0.1f);
+		heat.SetRelativeZ(1.0f);
```
AsuMagic, line 47: What does this change? Document it in the PR if it wasn't accidentally left there

Entities/Industry/Drill/Drill.as
```diff
@@ -191,7 +197,7 @@ void onTick(CBlob@ this)
 		AimAtMouse(this, holder); // aim at our mouse pos
 
 		// cool faster if holder is moving
-		if (heat > 0 && holder.getShape().vellen > 0.01f && getGameTime() % 3 == 0)
+		if (heat > 0 && holder.getShape().vellen > 0.01f && getGameTime() % 4 == 0)
```
AsuMagic, line 200: Should we even keep this mechanic at all since it is rather obscure imo?

Entities/Industry/Drill/Drill.as
```diff
@@ -301,7 +307,7 @@ void onTick(CBlob@ this)
 
 								if (isServer())
 								{
-									if (int(heat) > heat_max * 0.7f) // are we at high heat? more damamge!
+									if (heat_max - int(heat) < high_damage_window) // are we at high heat? more damamge!
```
AsuMagic, line 310: `heat_max - int(heat) < high_damage_window`  
is equivalent to  
`int(heat) >= heat_max - high_damage_window`  
which is a lot more readable

Entities/Industry/Drill/Drill.as
```diff
@@ -438,8 +444,8 @@ f32 onHit(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@ hitt
 
 	if (customData == Hitters::water)
 	{
-		s16 current_heat = this.get_u8(heat_prop) - heat_max*0.7f;
-		if (current_heat < 0) current_heat= 0;
+		s16 current_heat = this.get_u8(heat_prop) - heat_max * 0.5f;
```
AsuMagic, line 447: Overall it might be a good idea to refactor all those heat vars as global constants (e.g. `water_heat_reduction` or whatever). Not necessarily in this PR, though

AsuMagic (changes requested):

Entities/Industry/Drill/Drill.as
AsuMagic: ... or alternatively replace `high_damage_window` into a `high_damage_threshold` which is equal to this `heat_max - high_damage_window` value

Entities/Industry/Drill/Drill.as
bunniewormy: why not? it doesn't really change that much but it's pretty cool to have as a little detail imo

AsuMagic: There's no immediate feedback that it is possible, so this is one of the (various) things drills have that you need to be told about to know.

Entities/Industry/Drill/Drill.as
bunniewormy: done

Entities/Industry/Drill/Drill.as
bunniewormy: done the 1st

Entities/Industry/Drill/Drill.as
bunniewormy: removed

Entities/Industry/Drill/Drill.as
bunniewormy: i have no idea why this is here 

either was testing something and left it accidentally or it was changed to fix some obscure visual issue that was happening only on staging, but i have no idea what it could have been and don't know if i'm not misremembering

Entities/Industry/Drill/Drill.as
bunniewormy: done

AsuMagic (approved):

Blavier:
None of these changes make sense, and why they are being approved is beyond me. Just add drills ignoring bushes and scrap the rest of them.

AsuMagic:
Copypasta: Deciding to merge this for now. IMO the change makes sense as part of the rest of the balance goals, but it will be needed to track what meta/balance this results in after players have some experience playing with all the changes/adapt to whatever new meta. The feedback votes should be useful to characterize that, and revert or adapt part of the changes if needed.

Drill specific comment: The natural tile nerf is useful, and the buff isn't shockingly noticeable when testing in singleplayer, but as this might be one of the most sensitive parts of balance we'll see how the building/destruction changes makes this okay.

PR #1153, opened 2021-11-10
[CTF Rebuild] Bomb tornado in saws to replace saws exploding bombs (sawjumps removed)

bunniewormy:
## Status

(pick one)

## Description

[added] (water)bomb tornado to saws to replace saws exploding bombs

Removes sawjumps in a slightly fun way

Videos:

https://user-images.githubusercontent.com/39147836/141695725-f0dff5d2-be87-44d5-9c04-f8564223069c.mp4

https://user-images.githubusercontent.com/39147836/141695713-c61e7c72-33b2-4d3b-93c5-0a14350b41ed.mp4

https://user-images.githubusercontent.com/39147836/141695705-b637e498-57f3-404b-8b12-6692e30bc17d.mp4

https://user-images.githubusercontent.com/39147836/141695700-eaddc8c1-62d3-4df0-a957-2f0b8788b31b.mp4

https://user-images.githubusercontent.com/39147836/141695693-c4bb022b-0244-4753-a07c-4692cd60d071.mp4


MonkeyFeats:
A better method might be to reduce the max force that can be applied to a player, to just above however much a double bomb jump gives you, then have an exponential fall off after that.

bunniewormy:
> 
> 
> A better method might be to reduce the max force that can be applied to a player to just above however much a double bomb jump gives you; then have an exponential fall off after that.

I've thought of that, but I think max force limits aren't really fun and feel unnatural. Moreover, I believe that the most meaningful issue with sawjumps is that it was a niche trick that required specific game knowledge and which many people simply didn't understand - because of that, they should be entirely removed rather than continue to exist in a nerfed form (although personally I think they were quite fun and useful for breaking stalemates if you did knight+builder in crate meta)

Also, adding a max force limit will also nerf the knockback from having your bomb + enemy bomb explode at the same (which currently results in an impressive bomb jump), which would be a shame because it's fun whenever it happens

Mazey:
Saws killing most of the blobs is a fun mechanic and shouldn't be changed for this. Instead, fix the underlying issue, especially considering sawjumps are at the moment nowhere to be seen in public games. I'm in the same boat as @MonkeyFeats. 

When 2 bombs explode at the same time the force is quite high and anything higher than that causes abnormalities (as seen with f.i. sawjumps or bundled explosions that cause you to fly annoyingly high in the sky only to die instantly after).

>Also, adding a max force limit will also nerf the knockback from having your bomb + enemy bomb explode at the same

No, as @MonkeyFeats suggests (and if not, I am) limiting the force to 2 bombs (and dropping off quickly after that) will _exactly_ still allow this to happen (which is good, because it's indeed fun!), and with good timing & ping, you can mostly do this by yourself already, making saws redundant for bomb jumps).

bunniewormy:
Updated PR with new functionality - Saws now interact with bombs by deflecting them upwards. This also removes sawjumps.

Mazey:
It's fully functional but water bombs and this mechanic will absolutely need (proper gameplay) testing to make sure there are no weird quirks or side effects (though I already removed a couple I could find).

MrPotatobird:
I kinda like this weird bouncing thing too (is it just for team saws?) but couldn't you simply have a saw turn off when a teammate's bomb goes into it? Kinda like how a saw turns off when it teamkills. That way it would only pop _one_ of the bombs if you tried to sawjump, but still preserves a little bit of the original meme strat potential. And you wouldn't have to mess with the laws of physics.

You could even have it turn off after two bombs instead of just one if you wanted to try nerfing the saw jump somewhat without removing it completely.

AsuMagic:
Added some changes which match the version that was briefly tested on a server.

- Added spark particles on hit.
- Modified the audio, added some pitch adjustment for better audio cues.
- Tweaked physics, when hitting a saw on top the bombs will preserve *some* horizontal momentum. Nothing that allows using saws for shooting bombs far, though.
- Bombs hitting a saw have their fuse shortened quite significantly. This can be used to a strategical advantage from my testing.
That fuse shortening is semi-random, but it is possible to have bombs consistently blow up by stacking saws, which makes it possible to sawjump again. It becomes a lot more unpractical and expensive though so I don't think it's a big deal.
- Fixed some saw syncing which makes bloody saws work on net now.

Would like feedback on this too.

AsuMagic (approved):

PR #989, opened 2020-12-23
Resupply message tweaks; resupply on ballista without switching; builder and archer have seperate resupply timers.

Hobey:
- **READY**

[Demo Video 1](https://user-images.githubusercontent.com/26771587/102963130-abc75180-44e8-11eb-876b-3d3c24bd766c.mp4)
[Demo Video 2](https://user-images.githubusercontent.com/26771587/102963134-aec24200-44e8-11eb-947a-0f682dbe3f92.mp4)

---

Archer's resupply (for 30 arrows) and builder's resupply (for 300/100 wood + 100/30 stone) are now on seperate timers, meaning that if you resupply as builder, then 10 seconds later switch to archer, the builder resupply will not be delayed because you got 30 arrows. _Archers can now dropmats each respawn in a less annoying way to get wood+stone for their team and dropmats at their tower/midbase more effectively and without having to worry about the arrow-resupply stalling their wood+stone resupply if you messed up your retreat/resupply/death timings as archer._

You can now just walk to of a ballista/warboat when your resupply timer runs out and will get a resupply (just like with the spawn tent/shops). _Just for quality of life_

---

Changed the resupply message to help people that haven't figured out resupplies yet; see Demo Video 1 above.
Tweaked color of resupply message, disabled the text wobbling effect. _There's a lot of players that have played the game a lot but still don't know about resupplying or don't understand it fully. Resupplying and "dropping mats" is an awesome, fundamental game mechanic. It should be like the first thing you learn when you start playing CTF. At the moment, it is harder than it should be to notice the mechanic and figure it out. I hope the new resupply message + help text helps players learn it._

---

If all players in a 6v6-12v12 CTF match drop mats, 100 wood + 30 stone is actually too much. And in a 1v1-4v4, there's not really enough. I'll (hopefully find time and energy to) PR a resupply nerf (something similar to what we have on Captains, less mats the more players there are in a team; or some kind of diminishing return system maybe) at some point (feel free to do it, CatWiz/Bunnie, if you feel like it).

---

collapsed doGiveSpawnMats, canGetSpawnmats and GiveSpawnResources into one function; 
cleaned up onTick in CTF_GiveSpawnItems a bit for readability.
stopped using playerinfo.items_collected since there's now 2 resupply timers anyway (left the variable in the struct to minimize mods breaking)

PR #1154, opened 2021-11-10
[CTF Rebuild] Slight resupply rework. You always get resupplies in build time.

bunniewormy:
## Status

(pick one)

## Description

It's basically this PR https://github.com/transhumandesign/kag-base/pull/989

[change] resupplies every 30 seconds during build time rather than every 40
[change] changed the resupply message to help people that haven't figured out resupplies yet
[added] you get resupplies without having to go to a builder shop or a tent in build time
[added] text above inventory that tells you when you're able to receive builder resupplies 
[added] builder and archer resupply now use different timers
[added] easier resupplying on ballista


Rules/CTF/Scripts/CTF_GiveSpawnItems.as
```diff
@@ -4,99 +4,51 @@
...
-	if (blob !is null && player !is null)
-	{
-		RulesCore@ core;
-		this.get("core", @core);
-		if (core !is null)
-		{
-			doGiveSpawnMats(this, player, blob, core);
-		}
-	}
+	if (!getNet().isServer()) return;
```
Vam-Jam, line 46: isServer() instead of getNet().isServer()

Rules/CTF/Scripts/CTF_GiveSpawnItems.as
```diff
@@ -105,90 +57,89 @@ void onPlayerDie(CRules@ this, CPlayer@ victim, CPlayer@ attacker, u8 customData
...
 
+void SetCTFTimerBuilder(CRules@ this, CPlayer@ p, s32 time)
+{
+	string property = getCTFTimerPropertyNameBuilder(p);
+	this.set_s32(property, time);
+	this.SyncToPlayer(property, p);
 }
 
-string getCTFTimerPropertyName(CPlayer@ p)
+string getCTFTimerPropertyNameArcher(CPlayer@ p)
```
Vam-Jam, line 85: imo get rid of these funcs

do getCTFTimerPropertyName, which takes a player and a blob. you can just use the blob name instead of hard coding different string names and funcs

Rules/CTF/Scripts/CTF_GiveSpawnItems.as
```diff
@@ -216,48 +169,58 @@ void onTick(CRules@ this)
...
+	if (this.isWarmup()) 
+	{
+		// during building time, give everyone resupplies no matter where they are
+		for (int i = 0; i < getPlayerCount(); i++) 
+		{
+			CPlayer@ player = getPlayer(i);
+			CBlob@ blob = player.getBlob();
+			if (blob !is null) 
+			{
+			doGiveSpawnMats(this, player, blob);
```
Vam-Jam, line 187: missing tab

Rules/CTF/Scripts/CTF_GiveSpawnItems.as
```diff
@@ -267,32 +230,101 @@ void onRender(CRules@ this)
 {
 	if (g_videorecording || this.isGameOver())
 		return;
-
+	
 	CPlayer@ p = getLocalPlayer();
-	if (p is null || !p.isMyPlayer()) { return; }
-
-	string propname = getCTFTimerPropertyName(p);
+	if (p is null || !p.isMyPlayer()) return;
```
Vam-Jam, line 235: this check is not needed, safe to remove

(this is called once client side per frame, p will always be my player, and p will always exist as long as im in game)

Rules/CTF/Scripts/CTF_GiveSpawnItems.as
```diff
@@ -267,32 +230,101 @@ void onRender(CRules@ this)
...
+			
+			text = getTranslatedString("Next resupply of 30 arrows in {SEC}{TIMESUFFIX}.")
+				.replace("{SEC}", "" + secs)
+				.replace("{TIMESUFFIX}", getTranslatedString(units));
+
+			x = getScreenWidth() / 2;
+			y = getScreenHeight() / 3 - 16.0f;
 		}
+		
+		GUI::DrawTextCentered(text, Vec2f(x, y), color);
```
Vam-Jam, line 328: this could be made shorter, you shouldnt need to copy paste the code on a per class basis

Rules/CTF/Scripts/CTF_GiveSpawnItems.as
bunniewormy: fixed

Rules/CTF/Scripts/CTF_GiveSpawnItems.as
bunniewormy: fixed

Rules/CTF/Scripts/CTF_GiveSpawnItems.as
bunniewormy: this is vanilla code, but removed

~~if this will cause dumb issues in the future i will headbonk you~~

Rules/CTF/Scripts/CTF_GiveSpawnItems.as
bunniewormy: The point is to allow for displaying resupply text of a class regardless of what class the player is though (in this case specifically, display builder resupply text even if you're archer or knight. Archer resupply isn't displayed for other classes, but it's done the same way for consistency)

With that, should it still be modified? I could combine the functions into one and just add a string parameter instead of adding a blob parameter (string "archer" or string "builder")?

Rules/CTF/Scripts/CTF_GiveSpawnItems.as
bunniewormy: I've done some refactoring for this, could you please give a look if it's fine now?

eps0003:
The resupply text above your inventory displays for only a split second during build time since you immediately receive resupply. This might be fixable by adding a small delay before the text appears.

AsuMagic:
If/when merged - let's also attribute this change to Hobey as this is based upon his work.

Rules/CTF/Scripts/CTF_GiveSpawnItems.as
```diff
@@ -105,101 +63,168 @@ void onPlayerDie(CRules@ this, CPlayer@ victim, CPlayer@ attacker, u8 customData
...
-		}
-	}
+	return SPAWN_ITEMS_TIMER_BUILDER + p.getUsername();
+}
 
-	return false;
+s32 getCTFTimerBuilder(CRules@ this, CPlayer@ p)
+{
+	string property = getCTFTimerPropertyNameBuilder(p);
+	if (this.exists(property))
```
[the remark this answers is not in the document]
bunniewormy: yes

mugg91:
I think it isn't too big of an issue to just build a build shop and go there occasionally or to go to your tent.
Why not just add a help text that tells new players that they should be building their base and that they get materials at tent/buildshop/ballista etc.

AsuMagic:
> I think it isn't too big of an issue to just build a build shop and go there occasionally or to go to your tent. Why not just add a help text that tells new players that they should be building their base and that they get materials at tent/buildshop/ballista etc.

A lot of people are aware they need to go to the tent/buildershop to resupply at build time but a lot of people don't and just decide to mine for ages. This is generally a massive waste of resources for the team and I feel like it is hardly a good gameplay mechanic to require it as build time except for consistency (which there isn't really since the resupply amounts differ).

AsuMagic (approved):

mugg91:
One problem I see commonly in build 3900 is when I do get resupplies, sometimes it's just given to my teammate instead.
I guess when the materials spawn inside my character, and I'm overlapping with someone else, it's sometimes given to the wrong player. Maybe use "server_putininventory" (if it isn't already)?
It should be checked if this specific issue is fixed.

PR #1165, opened 2021-11-14
Revert halloween sprites

bunniewormy:
## Status

## Description

Title

Also, a lot of people seemed to hate gregs so we probably should think of something more fun for next year?


AsuMagic:
Reverting the two merges should be enough rather than reverting individual commits like that. Also while we don't have a proper way to make those holiday sprite changes automatically it should be cleaner to pack up everything in a branch that should be reversible with only one commit without resorting to dubious git magic every time.

Issue #1170, opened 2021-11-19
Distinction between gold and coins should be more obvious


AsuMagic:
**Current in-game behaviour.**

Newbies often get confused with the difference between coins and gold.

**Proposed Change**

Coins and gold should be distinguished better. This could be improved in a few areas:

- Including coin gain and loss in the hover messages (same as with kills and materials). This sounds like a good idea to me, but I don't know if it would be too distracting in practice.
- Mentioning coins more explicitly in some descriptions.
- Hovering coins/materials in the bottom bar could show a tooltip?

PR #1171, opened 2021-11-19
[added] hover message on coin count change next to materials

AsuMagic:
## Status

## Description

**Shows coin changes as part of the hover messages.**

This should be tested by more people as this could feel distracting, but in my testing hover messages are subtle enough that it is not an issue.

The fact it shows just next to materials makes it rather obvious what actions give you coins, such as building, fighting, and collecting coins - which is not necessarily obvious to a newbie since there is no other clear visual feedback, mind you. Plus it should help give some indication how much coins are actually earned.  
Since it aims towards fixing #1170, this should also help distinguishing coins and gold, because there has been various instances of newbies getting confused between the two.

This is implemented using a MaterialMessage, with the same timeout and category. The timeout being low could prove being an issue, and if it is, we could adjust it to increase it.

## Steps to Test or Reproduce

1. Join Sandbox or a local CTF server.
2. Earn coins through building, spend coins on stuff, etc.
3. Earned/lost coins should appear with behavior very similar to blob materials.

### DIFF
--- a/Entities/Common/HoverMessage/DrawHoverMessages.as
+++ b/Entities/Common/HoverMessage/DrawHoverMessages.as
@@ -43,6 +43,23 @@ const string[] ignored_material_losses = {
 	"mat_waterbombs",
 };
 
+void updateCoinMessage(CPlayer@ player)
+{
+	const string prop_name = "old coin count";
+
+	if (player.exists(prop_name))
+	{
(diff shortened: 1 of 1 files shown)


Vam-Jam (approved):

AsuMagic:
I'll merge it for now. We'll see in testing if this is too distracting.

PR #1175, opened 2021-11-25
Update accolades for Eluded

benjamingorman:
I am an OKSA so should have moderation accolade.

I've had 2 PRs merged so should have github accolade:
https://github.com/transhumandesign/kag-base/pulls?q=is%3Apr+author%3Abenjamingorman+is%3Aclosed


eps0003 (changes requested):
You need to add the comments that state why the accolades were given

eps0003 (approved):

PR #1176, opened 2021-11-26
Birthday

Mazey:
## Status

- **READY**
## Description

- Adds birthday event on KAG's birthday
Birthday heads and balloons on various blobs
image
image
image

- Changes the way heads are chosen for holidays
Head is now set from within the relevant holiday script for Halloween and Christmas. Also support for custom textures (as seen in the new Birthday event)


Rules/CommonScripts/Holiday/Birthday/Birthday.as
```diff
@@ -1 +1,48 @@
...
+	for (u8 i = 0; i < holiday_blobs.length(); i++)
+	{
+		getBlobsByName(holiday_blobs[i], @blobs);
+	}
+
+	for (u8 i = 0; i < blobs.length(); i++)
+	{
+		CBlob@ b = blobs[i];
+		AddBalloons(b.getSprite(), b.getName());
+	}
```
Vam-Jam, line 25: Wont this need to be inside onRestart, since onInit is only called once on script is created (e.g. on first map load or after addScript). So unless the script get's removed and re-added every new round, this should be moved.

AsuMagic (approved):

PR #1191, opened 2021-12-25
Christmas music!

butterdome:
i replaced all currently used music files with my updated christmas versions (peaceful and battle themes)

## Status

- **READY**: haven't tested it yet (bcause i dont know how ;_; ) but it should work, since i didnt change any code, just replaced the files.

## Description

should all be the correct filetype (.ogg) , they all have minimal fade ins and outs, are slightly louder than the previous ones.


butterdome:
ok i tested it. works fine.
i mean as fine as the current music system works (very buggy)
but it works

soren-null (dismissed):
Tested in solo, sandbox and main menu, seems to work as intended and ready to be incorporated

AsuMagic:
Rebased against current `master` to fix conflicts.

AsuMagic (approved):
Tested ingame, works

PR #1196, opened 2022-01-04
High ping building

laws65:
## Status
- **READY**

## Description
-Adds server side check that player has requirements when building blobs
-Fixes typo in server side check for player being build delayed
-Allows quickly switching blocks on high ping so this situation doesn't happen:

> -player sends input to change block
> -player sends input to place block
> -server receives input to change block
> -server receives input to place block and since it's using outdated info from client, places old block
> -player is told they have selected another block
> -placed block is shown to player


laws65:
Lol nvm

PR #1197, opened 2022-01-04
[modified] Enabled + fixed builder quickswap to last block (tap F)

MrPotatobird:
## Status

## Description

Builders can tap F to cycle to last block, like other classes can cycle ammo types

Quickswap for builders was added a while ago, but quickly reverted because of a bug: if you opened + clicked the menu to select an option _really fast_, it still counted as a "tap" and thus immediately cancelled the selection you clicked on. So it kind of broke the game for some good builders who used the click method. The bug was actually preexisting for quickswap in general, but knights and archers would never actually click on the buttons that fast. Anyway it's fixed now

## Steps to Test or Reproduce

- Be builder
- Select a block (stone for instance)
- Select another block (like door)
- Tap F - switches back to first block

To see that it's not bugged:
- Be builder
- Select stone block (or whatever) in build menu
- Open build menu with F, click on ladder, and release F quickly
- It should select ladder without quickswapping back to previous block

### DIFF
--- a/Entities/Characters/Builder/BuilderInventory.as
+++ b/Entities/Characters/Builder/BuilderInventory.as
@@ -42,7 +42,7 @@ const u8 GRID_PADDING = 12;
 const Vec2f MENU_SIZE(3, 4);
 const u32 SHOW_NO_BUILD_TIME = 90;
 
-const bool QUICK_SWAP_ENABLED = false;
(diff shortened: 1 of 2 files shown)


MrPotatobird:
Ah this doesn't really work at the moment anyway. I'm still planning on fixing it and making it optional eventually

Issue #1194, opened 2022-01-02
Make fire arrows hitting shield play shield hit sound


NoNamedNN:
**Description**

Fire arrow doesn't make any sound when it hits enemy knight's shield.

**Reproduction**

1. Hit enemy knight's shield with fire arrow.
2. Notice that there is no sound unlike normal arrows.

**Expected Behavior**

Fire arrow hitting enemy knight's shield makes shield hit sound.

PR #1198, opened 2022-01-04
Make fire arrows hit like normal arrows, fix sound

Nananas:
## Status

## Description

This change makes fire arrows behave more like normal arrows, by first registering an arrow hit (so they will make a sound when hitting a knight's shield and allow stun when shooting fully charges) and then do the usual fire hit with, but with 0 initial damage, if not shielding.

I have also removed a redundant `damage = 0` for shield hits. This is already handled in `ShieldHit.as`, and seemed to not do much afaik.

## Steps to Test or Reproduce

Shoot a shielding knight with a fire arrow and notice the sound, and stun (for full charge), and no fire 
Shoot a non-shielding knight and notice the damage and fire 🔥, and possible stun


Nananas:
Should fix #1194

AsuMagic (approved):

PR #1199, opened 2022-01-04
[modified] Team bombs disable saws when milled (no sawjumps)

MrPotatobird:
## Status

## Description

An alternative to #1153 

Saws will turn off after detonating one teammate's bomb, effectively preventing sawjumps while allowing most "normal" usage of saws to pop bombs. Enemy saws will not turn off

## Steps to Test or Reproduce

- Get saw
- Throw a bomb in there


AsuMagic:
Currently closing as #1153 was merged. That being said, if #1153 turns out to be too derpy or simply not worth it we can consider reverting and merging this instead.

PR #1074, opened 2021-07-21
Trampoline has 360 rotation

CatWiz:
## Status

- **READY**: this PR is (to the best of your knowledge) ready to be incorporated into the game.

## Description

Attempt 2.
When held, trampoline freely rotates like drill. 
Holding lmb makes it rotate in 45 degree steps, and holding rmb prevents it from rotating completely (allows players to throw tramps at any angles in different directions and place horizontal tramps more easily)

PR #1203, opened 2022-01-29
Trampoline spacebar activation to freeze tramp holding angle (with 360 degree rotation)

MrPotatobird:
## Status

## Description

Alternative to mouse button behavior of #1074 that seems more intuitive + nice to use to me than the mouse buttons
Press spacebar to freeze the angle until it's dropped, spacebar again throws
Doesn't have a way to snap to 45-degree steps cuz it didnt seem that necessary to me

## Steps to Test or Reproduce

- Get trampoline
- Hold at desired angle
- Press spacebar

### DIFF
--- a/Entities/Structures/Trampoline/TrampolineLogic.as
+++ b/Entities/Structures/Trampoline/TrampolineLogic.as
@@ -43,10 +43,26 @@ void onTick(CBlob@ this)
 	ray.Normalize();
 
 	f32 angle = ray.Angle();
-	angle = angle > 135 || angle < 45? (holder.isFacingLeft()? 135 : 45) : 90;
-	angle -= 90;
 
-	this.setAngleDegrees(-angle);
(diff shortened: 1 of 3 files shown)


Entities/Structures/Trampoline/TrampolineLogic.as
```diff
@@ -134,6 +136,27 @@ void onCollision(CBlob@ this, CBlob@ blob, bool solid, Vec2f normal, Vec2f point
 	}
 }
 
+void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
+{
+	if (cmd == this.getCommandID("activate"))
+	{
+		if (getNet().isServer())
```
laws65, line 143: Use `isServer()` instead of `getNet().isServer()`

laws65:
Is it intended behaviour that the trampoline is thrown when it's frozen and you press spacebar instead of it just being unfrozen?

PR #993, opened 2020-12-23
The map vote screen now has 4 maps as options

Hobey:
- **READY**

Example:
Example

The map vote screen now has 4 maps as options, and no 'random map' option.
In case of tie, it will pick one of the maps with the most votes at random.

_I think this is a more fun map vote system than the current one._

---
sorry about the inconsistent indentation causing extra line diffs in the commit

PR #1204, opened 2022-01-30
Removed random map vote

CatWiz:
## Status

- **HOLD**: This changes a tcpr command sent to kagstats, probably better not to add this until it's updated

## Description

Replaces "random map" vote option with a third map. Seems like quite a few players want this.


butterdome:
i'd suggest 4 maps to replace the random vote with more variety =)

Mazey:
Already implemented by #993 which is waiting to be merged/reviewed. Should it be 3 maps or 4? Let’s keep discussion there.

PR #1205, opened 2022-01-30
Script-side support for staging off-screen tick throttling

AsuMagic:
## Status

## Description

Staging exposes a `throttleInterval` field to `CBlob` which allows scripts to skip ticking of blobs entirely by only ticking every N ticks. It aims for maximal overhead reduction by cutting ticking much earlier than script ticking intervals would allow. The engine also attempts to spread out entity throttling in order to reduce tick time fluctuation.  
However, this remains largely experimental because skipping ticks on the client-side could have many unintended side effects. Currently, Staging uses that `throttleInterval` unconditionally but it may be disabled by default behind an `autoconfig.cfg` switch until this is largely proven to be stable.

I tested client-side throttling with a proof-of-concept on live release servers in busy matches, and found this to allow for significant tick time reduction.

This PR provides the required script-side infrastructure to allow blobs to opt in for tick throttling through a very simple and generic ruleset, through a newly introduced `OffscreenThrottle.as` script. It also enables that support for several static blobs.  
By default, tick throttling is performed on all clients, including localhost. Mods can trivially override this behavior.

On release KAG, this PR causes no functional change.  
This can be applied to current Release servers, and will only apply to Staging clients.

## Steps to Test or Reproduce

- Use the Staging build
- Join a local server (localhost or dedicated works)
- Press F5
- Click "Debug windows", then "Blobs"
- Certain blobs that are out of the screen (doors, bushes, platforms, etc.) should show up as `ThrXXX` (ticking only every XXX ticks). You should see "waves" of these going to `Ok` for a single tick.

### DIFF
--- a/Entities/Common/Scripts/OffscreenThrottle.as
+++ b/Entities/Common/Scripts/OffscreenThrottle.as
@@ -1,0 +1,49 @@
+// Add this script to the script list of a blob to allow it to skip ticks when offscreen.
+// This is useful to improve performance.
+//
+// This does significantly more than just not calling onTick hooks!
+// Nearly _nothing_ will be done to update the blob while it is throttled.
+// This should really only be used on somewhat static blobs (which, in particular, do _not_ increment a counter onTick!).
+//
+// NOTE: This optimization will *not* be done on servers, but it will be done in localhost.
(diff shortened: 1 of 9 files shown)


Vam-Jam (approved):
:o

PR #1209, opened 2022-03-26
Scoreboard country flags

eps0003:
## Status

## Description

This adds country flags on the scoreboard for each player. It achieves this by using TCPR to fetch location data from an external program. The data received is the player's country code, country, and region. Location data is synced to clients using the player's network ID rather than their IP address to ensure players' IP addresses remain only accessible from the server.

The flag sprites were taken from [here](http://www.famfamfam.com/lab/icons/flags/) and are "available for free use for any purpose with no requirement for attribution".

The external program can be found [here](https://github.com/eps0003/kag-player-location).

## Caveats 

- Players joining a server hosted on the same network won't have a flag since the server only has their local IP address.
- Not all country codes have been tested to see if they properly correspond with a flag, so some players from lesser-known countries may not have a flag.

## Screenshots

image

## Steps to Test or Reproduce

1. Set `sv_tcpr = 1` in server `autoconfig.cfg` and run the server.
2. Configure and run the external program, ensuring it connects with the server.
3. Join the server with people from different countries.


AsuMagic:
A few thoughts:
- I don't feel comfortable with merging this if this is not explicitly opt-in
- I'm not sure how GDPR friendly it is to have the TCPR program send all IPs to a third-party service (there _are_ local IP database libs to my knowledge which may be better)
- If made opt-in, we can drop the reliance on an external program by just allowing the user to pick their country/country flag anyways

AsuMagic (changes requested):
A few thoughts:
- I don't feel comfortable with merging this if this is not explicitly opt-in
- I'm not sure how GDPR friendly it is to have the TCPR program send all IPs to a third-party service (there are local IP database libs to my knowledge which may be better) without the user consent or at least awareness
- (If made opt-in, we can drop the reliance on an external program by just allowing the user to pick their country/country flag anyways)

eps0003:
I will close this since your third point would be a simpler, less intrusive way of doing this.

PR #1212, opened 2022-03-30
Force all players to be a builder until the last 20 seconds of warmup

### LINKED ISSUE #738: Force all players to be a builder until the last 20 seconds of warmup
I need my team mats

thank you for coming to my ted talk

laws65:
## Status

## Description

This PR stops players from switching to other classes in build time until the last 20 seconds (changeable via ctf_vars.cfg). Resolves #738

It works by tagging CRules with "class switching disabled" which all class switching places like the tent, class shops, etc. check for. If switching is disabled, the button is greyed out and text is displayed saying that it's disabled. 

When class switching is disabled, the players will spawn in based on a CRules string property called "default class". If it isn't set, players will spawn in as knights.

The "class switching disabled" tag and "default class" string works for all gamemodes, which for example makes all knight or all archer TDM easy to do.

This PR also fixes the inconsistencies of the class swap buttons text, which is now "Change Class". Before it was that or "Swap Class" or one of the two but missing capitalisation.

Note: The text displaying when the builder only time ends is trash and should be changed but I can't think of better wording.

## Steps to Test or Reproduce

- Start a game, notice you spawn in as builder automatically
- Try to switch to another class, see class button is different and pressing it doesn't do anything
- See build time text
- Wait until the builder only time is over and see that you can now change classes again

### DIFF
--- a/Entities/Industry/Hall/Hall.as
+++ b/Entities/Industry/Hall/Hall.as
@@ -377,33 +377,45 @@ void GetButtonsFor(CBlob@ this, CBlob@ caller)
 	if (!canSeeButtons(this, caller) || !caller.isOverlapping(this))
 		return;
 
-	if (this.getTeamNum() != 255)
+	if (!canChangeClass(this, caller))
+		return;
+	
+	if (this.getTeamNum() == 255)
+		return;
+	
+	CBitStream params;
+	params.write_u16(caller.getNetworkID());
+
+	if (!getRules().hasTag("class switching disabled"))
 	{
-		CBitStream params;
-		params.write_u16(caller.getNetworkID());
+		caller.CreateGenericButton("$change_class$", Vec2f(12, 7), this, buildSpawnMenu, getTranslatedString("Change Class"));
+	}
+	else
+	{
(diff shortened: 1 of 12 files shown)


Entities/Industry/Hall/Hall.as
```diff
@@ -377,33 +377,45 @@ void GetButtonsFor(CBlob@ this, CBlob@ caller)
...
+	if (!canChangeClass(this, caller))
+		return;
+	
+	if (this.getTeamNum() == 255)
+		return;
+	
+	CBitStream params;
+	params.write_u16(caller.getNetworkID());
+
+	if (!getRules().hasTag("class switching disabled"))
```
eps0003, line 389: What's preventing this (and all the other occurrences) from being moved into the `canChangeClass` function?

Gingerbeard5773:
I think this has been resolved, now that mats drop from all classes during warmup.

AsuMagic:
Agreed, though there is still this typo fix:

> This PR also fixes the inconsistencies of the class swap buttons text, which is now "Change Class". Before it was that or "Swap Class" or one of the two but missing capitalisation.

PR #1225, opened 2022-04-09
Fix bad snapshot error that was caused by archer changing class on tent or other respawn points

GoldenGuy:
## Status

## Description

When player changes class from archer to archer in tent or any other respawn, there might be a bad sync, that would cause a bad snapshot, or sometimes bad delta spam that would crash the server. Here i replaced Sync with Command to prevent this from happening.

## Steps to Test or Reproduce

Quickly press E and pick archer class couple of times, you will see errors in console.

### DIFF
--- a/Entities/Characters/Archer/ArcherLogic.as
+++ b/Entities/Characters/Archer/ArcherLogic.as
@@ -37,6 +37,7 @@ void onInit(CBlob@ this)
 	this.getSprite().SetEmitSound("Entities/Characters/Archer/BowPull.ogg");
 	this.addCommandID("shoot arrow");
 	this.addCommandID("pickup arrow");
+	this.addCommandID("has arrows");
 	this.getShape().getConsts().net_threshold_multiplier = 0.5f;
 
 	this.addCommandID(grapple_sync_cmd);
@@ -342,7 +343,12 @@ void ManageBow(CBlob@ this, ArcherInfo@ archer, RunnerMoveVars@ moveVars)
 		if (hasarrow != this.get_bool("has_arrow"))
 		{
 			this.set_bool("has_arrow", hasarrow);
-			this.Sync("has_arrow", isServer());
+			CBitStream params;
+			params.write_bool(hasarrow);
+			if(isServer())
+				this.SendCommand(this.getCommandID("has arrows"), params);
+			else
+				this.SendCommandOnlyServer(this.getCommandID("has arrows"), params);
 		}
 
 	}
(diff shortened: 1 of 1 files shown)


Entities/Characters/Archer/ArcherLogic.as
```diff
@@ -342,7 +343,12 @@ void ManageBow(CBlob@ this, ArcherInfo@ archer, RunnerMoveVars@ moveVars)
 		if (hasarrow != this.get_bool("has_arrow"))
 		{
 			this.set_bool("has_arrow", hasarrow);
-			this.Sync("has_arrow", isServer());
+			CBitStream params;
+			params.write_bool(hasarrow);
+			if(isServer())
```
laws65, line 348: Add a space between if and (

Entities/Characters/Archer/ArcherLogic.as
```diff
@@ -429,7 +435,12 @@ void ManageBow(CBlob@ this, ArcherInfo@ archer, RunnerMoveVars@ moveVars)
 			if (responsible)
 			{
 				this.set_bool("has_arrow", hasarrow);
-				this.Sync("has_arrow", isServer());
+				CBitStream params;
+				params.write_bool(hasarrow);
+				if(isServer())
```
laws65, line 440: Same here

laws65 (changes requested):

GoldenGuy:
@laws65 here

Entities/Characters/Archer/ArcherLogic.as
```diff
@@ -1061,6 +1072,14 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
 			CycleToArrowType(this, archer, type);
 		}
 	}
+	else if (cmd == this.getCommandID("has arrows"))
+	{
+		bool hasarrow;
+		if(params.saferead_bool(hasarrow))
```
laws65, line 1078: Also here, sorry I missed it before

Entities/Characters/Archer/ArcherLogic.as
```diff
@@ -342,7 +343,12 @@ void ManageBow(CBlob@ this, ArcherInfo@ archer, RunnerMoveVars@ moveVars)
 		if (hasarrow != this.get_bool("has_arrow"))
 		{
 			this.set_bool("has_arrow", hasarrow);
-			this.Sync("has_arrow", isServer());
+			CBitStream params;
+			params.write_bool(hasarrow);
+			if (isServer())
+				this.SendCommand(this.getCommandID("has arrows"), params);
+			else
+				this.SendCommandOnlyServer(this.getCommandID("has arrows"), params);
```
eps0003, line 351: I haven't tested this but couldn't these lines be simplified to something like this?
```angelscript
this.SendCommand(this.getCommandID("has arrows"), params, isServer());
```

GoldenGuy: only CRules SendCommand has this bool at the end, CBlob function doesnt have it

PR #1226, opened 2022-04-10
Prevent knight attack when using shops

GoldenGuy:
## Status

## Description

Normally, when knight click shop buttons, he does the attack at the same time, this is bad, because you can use shops from different team, and this attack will hit the shop. This pr disables knight attacks when using shops. Code can also be resued for other classes, but it doesnt seems useful for them as much as for knight.

## Steps to Test or Reproduce

- use shop as knigth and click shop buttons, you will see that knight doesnt do his attacks

### DIFF
--- a/Entities/Characters/Knight/KnightLogic.as
+++ b/Entities/Characters/Knight/KnightLogic.as
@@ -458,7 +458,7 @@ class NormalState : KnightState
 
 	bool TickState(CBlob@ this, KnightInfo@ knight, RunnerMoveVars@ moveVars)
 	{
-		if (this.isKeyPressed(key_action1) && !moveVars.wallsliding)
+		if (this.isKeyPressed(key_action1) && !moveVars.wallsliding && !InShop(this))
 		{
 			knight.state = KnightStates::sword_drawn;
 			return true;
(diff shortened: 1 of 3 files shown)


Entities/Characters/Knight/KnightLogic.as
```diff
@@ -1745,3 +1745,8 @@ void CheckSelectedBombRemovedFromInventory(CBlob@ this, CBlob@ blob)
 		SetFirstAvailableBomb(this);
 	}
 }
+
+bool InShop(CBlob@ this)
```
laws65, line 1749: Use camel case for functions that return stuff

Entities/Common/Factory/Shop.as
```diff
@@ -105,6 +103,13 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
 
 		CBlob@ caller = getBlobByNetworkID(callerID);
 		if (caller is null) { return; }
+
+		if(this.hasTag(SHOP_AUTOCLOSE))
+			caller.set_bool("shop prevent attack", false);
```
laws65, line 108: Add curly braces

asumagic:
Tempted to instead fix this by eating the click when it lands inside cgridmenus instead. This might break some mods which would rely on detecting click events landing in cgridmenus (cough cough bunnie's clientside placing) and would have to be able to prevent sending the action key update to the server (if it even uses that, I don't remember).

We really need some sort of API that both scripts and the engine could use to determine regions of the screen that represent clickable GUI, with custom defined depth/priority levels.

PR #1236, opened 2022-04-22
Fireplace emits sound from the start and doesn't spin

mugg91:
## Status

## Description

Fireplace had an issue where the CampfireSound.ogg didn't play unless you extinguished and reignited the fireplace.
Also fireplace could spin around if falling on ledges and fire burning downwards looks odd.
This change makes it so Fireplace will always play the campfire sound when it is on fire and will not spin around anymore.

## Steps to Test or Reproduce

- Go to solo mode "Save the princess" and notice the campfire at the beginning is emitting sound correctly.
- Go to Sandbox and spawn Fireplaces, notice they emit sound correctly.
- Spawn a fireplace above a ledge, notice it doesn't spin around anymore.


eps0003:
I remember purposefully allowing fireplaces to rotate so they wouldn't look unusual when they hang off a block, but I guess it's even more unusual for them to burn while upside down. 😄

mugg91:
I increased friction from 0.6 to 1.9 so water bomb and such things can't push fireplace far anymore. Fireplace is supposed to be stationary.

AsuMagic (approved):

Issue #1231, opened 2022-04-16
Spikes no longer attached to stone will not do damage


mugg91:
**Description**

If a spike was attached to stone and then is attached back to dirt/wood, it will not deal damage.

This happens in single player and online.

**Screenshots**

https://i.imgur.com/mQu6iPj.jpeg

**Reproduction**

1. Place spikes and stone block so the spikes are attached to the stone block.
2. Remove the stone block, make sure there is dirt or wood next to the spike block.
3. The spikes become attached to the wood/dirt.
4. Jump into or run against the spikes, notice it doesn't do damage.

**Expected Behavior**

Spikes should do damage.

PR #1239, opened 2022-04-24
Spikes not attached to stone anymore are doing damage

mugg91:
## Status

## Description

This fixes [issue #1231](https://github.com/transhumandesign/kag-base/issues/1231).
Spikes will now do damage again when attached to dirt/wood after no longer being attached to stone.

The bug was that the spikes thought they are still in state "hidden", therefore the onCollision() function didn't execute the damage.

I added a line in onTick() so if placedOnStone is false, then set the spikes to state "normal".

I play-tested the change a bit offline and there seems to be no side-effects. The spikes are behaving as usual.
Before this gets into the next patch, please test this change online, too. (Although I don't think there are side-effects online.)

## Steps to Test or Reproduce

- Place spikes next on stone.
- Remove the stone and make sure there is dirt/wood next to the spikes that it can attach itself to.
- Run, jump or fall against the spikes. Notice they are doing damage now.

### DIFF
--- a/Entities/Structures/Spikes/Spikes.as
+++ b/Entities/Structures/Spikes/Spikes.as
@@ -256,6 +256,7 @@ void onTick(CBlob@ this)
 	{
 		this.getCurrentScript().tickFrequency = 25;
 		this.getSprite().SetAnimation("default");
+		this.set_u8(state_prop, 0);
 		this.set_u8(timer_prop, 0);
 	}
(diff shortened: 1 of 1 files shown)


soren-null (approved):
Fixes the issue, tested on dedicatedserver

AsuMagic (approved):

PR #1215, opened 2022-04-03
Allow seeds to be loaded into catapult and show correct icon for seeds and food

laws65:
## Status

- **READY**: this PR is (to the best of your knowledge) ready to be incorporated into the game.

## Description

Allows seeds to be loaded into catapult and makes the correct icon show for them and food when loading.

## Steps to Test or Reproduce
-Spawn in a catapult
-Load different types of seeds
-Wow, it shows the correct icon
-Load different foods (Cake, Bread, Cooked Fish or Cooked Steak)
-Wow, it shows the correct icon

PR #1241, opened 2022-04-27
Loading Raft on Catapult shows the correct icon, Raft blocks players from using actions

mugg91:
## Status

## Description

This is an addition/enhancement to [Pull Request #1215](https://github.com/transhumandesign/kag-base/pull/1215). You need to apply PR #1215 and then this to see the changes.

There are two changes. Please let me know what you think.

1)
Loading Raft on Catapult now shows the correct icon.
To do this, I edited VehicleIcons.png to add a small Raft sprite, replacing frame 7. (I did so because I think frame 7 was unused or was a placeholder anyway?)

2)
Players carrying a Raft will now be unable to use actions like slash, shield, bow, grapple(...)
Dinghy already has this implemented and I felt that Raft (which is supposed to be heavier weight) should implement this logic as well.
To do this, I added "BoatCommon.as" to "Raft.cfg".

## Steps to Test or Reproduce

- Try to load a Raft on a Catapult, notice it will use the correct sprite icon.
- Try to use actions while carrying a Raft, notice it will not work.


eps0003:
It appears that you are manually uploading your changes to GitHub. This is not ideal since it could lead to human error when copying over your changes. Please look into how to set up a local dev environment so you can test and commit without needing to copy and paste. You can ask for help in #development channel on Discord if you are unsure how to do this.

mugg91:
I'm not doing this on a highly professional level by any means. I'm using notepad++ and I'm having a separate dev install where I edit the files. If I come up with a solution I use the github file finder and copy paste the entire content of a file there. Before I PR, I copy paste the content of files back into my install and test if things are still going as expected. So there should not be a problem. I triple check everything. But I will take what you said into consideration for the future.

Entities/Vehicles/Boats/Raft.as
```diff
@@ -1,4 +1,4 @@
-// Boat logic
+#include "GenericButtonCommon.as"
```
AsuMagic, line 1: This does not appear to be used?

AsuMagic (changes requested):
`VehicleIcons.png` is now conflicting because of the outpost addition

Entities/Vehicles/Boats/Raft.as
Gingerbeard5773: It is not used. This line can be removed safely.

Gingerbeard5773 (changes requested):
VehicleIcons

The PR has an outdated image. Here's an updated image ^ that will not conflict.
I tested and found no other problems.

AsuMagic (approved):

Issue #592, opened 2019-05-21
Obstructor block breaks background


mugg91:
As builder, in a sandbox game, stand on top of background and then select the obstructor. Then select something else. The game will immediately destroy 1 tile of background for no reason.

mugg91:
Also happens with motion sensor

mugg91:
I confirmed that the bug is still present in the current build 3900.

mugg91:
I fixed Motion Sensor and now I will try to fix Obstructor.

PR #1243, opened 2022-04-29
Obstructor and Motion Sensor no longer break background

mugg91:
## Status

## Description

This fixes [issue #592](https://github.com/transhumandesign/kag-base/issues/592).
Background and grass will no longer break when selecting obstructor/sensor and then something else in the build menu.

The same code in DummyOnStatic.as is executed when "selecting obstructor/sensor and then something else" and when the tile actually breaks. So I check the tile's health to make sure the code only executes in the latter case, when the health is 0.

I play-tested the change and the two components behave the same as before.

### DIFF
--- a/Entities/Structures/Common/DummyOnStatic.as
+++ b/Entities/Structures/Common/DummyOnStatic.as
@@ -5,7 +5,7 @@
 #include "DummyCommon.as";
 
 void onSetStatic(CBlob@ this, const bool isStatic)
-{
+{	
 	if (this.exists(Dummy::TILE))
 	{
 		const Vec2f POSITION = this.getPosition();
@@ -19,9 +19,12 @@ void onSetStatic(CBlob@ this, const bool isStatic)
 				server_setDummyGridNetworkID(map.getTileOffset(POSITION), this.getNetworkID());
 			}
 			else
-			{
-				map.server_SetTile(POSITION, CMap::tile_empty);
-				server_setDummyGridNetworkID(map.getTileOffset(POSITION), 0);
+			{	
+				if (this.getHealth()==0) //this check fixes Obstructor and Sensor breaking background
+				{ 
+					map.server_SetTile(POSITION, CMap::tile_empty);
+					server_setDummyGridNetworkID(map.getTileOffset(POSITION), 0);
+				}
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


Entities/Structures/Common/DummyOnStatic.as
```diff
@@ -19,10 +19,13 @@ void onSetStatic(CBlob@ this, const bool isStatic)
 				server_setDummyGridNetworkID(map.getTileOffset(POSITION), this.getNetworkID());
 			}
 			else
-			{
-				map.server_SetTile(POSITION, CMap::tile_empty);
-				server_setDummyGridNetworkID(map.getTileOffset(POSITION), 0);
+			{	
+				if (this.getHealth()==0) //this check fixes Obstructor and Sensor breaking background
```
eps0003, line 23: Be sure to use whitespace around comparison operators.
```suggestion
				if (this.getHealth() == 0) //this check fixes Obstructor and Sensor breaking background
```

eps0003, line 23: Couldn't this be merged with the `else` instead of it being nested inside?

eps0003 (changes requested):
I am yet to test in-game.

mugg91:
I found that both components will remove wood/stone background (if it existed) when getting built.
Obstructor will also remove it whenever goes into background mode.
Not sure if it needs addressing, but it seemed odd.

Entities/Structures/Common/DummyOnStatic.as
```diff
@@ -5,7 +5,7 @@
 #include "DummyCommon.as";
 
 void onSetStatic(CBlob@ this, const bool isStatic)
-{
+{
```
Gingerbeard5773, line 8: tab

Entities/Structures/Common/DummyOnStatic.as
```diff
@@ -18,11 +18,11 @@ void onSetStatic(CBlob@ this, const bool isStatic)
 				map.server_SetTile(POSITION, this.get_TileType(Dummy::TILE));
 				server_setDummyGridNetworkID(map.getTileOffset(POSITION), this.getNetworkID());
 			}
-			else
-			{
+			else if (this.getHealth() == 0) //this check fixes Obstructor and Sensor breaking background
+			{
```
Gingerbeard5773, line 22: tab

Gingerbeard5773 (approved):
Works as intended

AsuMagic (approved):

Issue #1242, opened 2022-04-27
Fireplace that has been there from the start cannot be extinguished


mugg91:
**Description**

Fireplaces that have been there from the start, such as the ones in Save the Princess, cannot be extinguished by Water Bomb and Water Arrow. But they can be extinguished by Water Bucket and water tiles.

**Reproduction**

1. Go to Save the Princess first map. Notice the fireplace at the beginning.
2. /sv_test 1 in console.
3. !allbombs
4. Throw water bomb at the fireplace. Notice it will not be extinguished.
5. !archershop and change class to Archer.
6. !allarrows
7. Shoot water arrow at the fireplace. Notice it will not be extinguished.

**Expected Behavior**

Fireplace should be possible to extinguish by all water sources.

**Additional context**

Posting this issue more for future reference so it doesn't become forgotten.

Issue #1206, opened 2022-03-15
Fireplace being re-ignited


mugg91:
I extinguished a fireplace and then tried shooting it with fire arrows, but it doesn't ignite.

So I think it would be cool if it was possible to ignite it again with fire arrows.

Perhaps it's also an idea to make a builder be able to re-ignite a fireplace.

mugg91:
Ok, I found out few minutes later that a fireplace can in fact be re-ignited if you place a wood background on it and then shoot that with the fire arrow. 
So at least there is a way, but it's not so straight-forward. 
I let you decide if my suggestion still has worth.

eps0003:
I think I purposefully didn't allow any arrows to collide with campfires because that would be annoying for archers shooting near them. There are two alternative ways this could be implemented:

1. Reigniting it when a fire arrow is stuck on top of the block underneath the campfire.
2. Reignite it when a fire arrow passes through a campfire

mugg91:
Yes, either one of those would be more straight-forward than it requiring a wood background, I guess.

PR #1244, opened 2022-04-30
Fireplace that has been there from the start can be extinguished, Fireplace can be set on fire with fire arrow

mugg91:
## Status

## Description

This fixes [issue #1242](https://github.com/transhumandesign/kag-base/issues/1242) and [issue #1206](https://github.com/transhumandesign/kag-base/issues/1206).

Fireplace that has been there from the start - such as the ones in Save the Princess' first map - will now be able to be extinguished by water bomb and water arrow. It was not possible before because with those two (water arrow and water bomb) the game assumed hitter was "water_stun" rather than "water" but fireplace only checked for "water" in onHit().

Fireplace can now be set on fire when shooting fire arrow through them. 
This was done by checking in onCollision() if a blob with fire source is colliding with fireplace.
It works well for the most part. Shooting fire arrow through many fireplaces will ignite all of them most of the time. But shooting from top to bottom may fail sometimes.

I discarded the idea that arrows should have direct collision with fireplace (similar behavior like how vehicles can accept arrow hits) for these reasons: 
* Arrows usually force things to get knocked back, and fireplace is not supposed to get knocked back by an arrow, in my opinion.
* Arrows hitting fireplace would ignite themselves and burn nearby grass/stuff unexpectedly.
* If making it so that arrows despawn when hitting fireplace (so stuff doesn't burn), then shooting arrows through the fireplace to ignite the arrows would not be possible anymore. And I don't want inconsistencies like "turned-off fireplace accepts and despawns fire arrow but turned-on fireplace lets arrows fly through".

Even though fire arrow can't directly hit fireplace, I still added a check "isIgniteHitter" in "onHit()" anyway.

I play-tested and didn't encounter regressions.

## Steps to Test or Reproduce

* Go to Save The Princess, throw water bomb or shoot water arrow at the fireplace at the beginning. Notice it gets extinguished now.
* Extinguish a fireplace and shoot fire arrow through it. Notice that it gets ignited.

### DIFF
--- a/Entities/Industry/Fireplace/Fireplace.as
+++ b/Entities/Industry/Fireplace/Fireplace.as
@@ -46,17 +46,20 @@ void onTick(CBlob@ this)
 
 void onCollision(CBlob@ this, CBlob@ blob, bool solid)
 {	
-	if (blob !is null && this.getSprite().isAnimation("fire"))
-	{
-		CBlob@ food = cookFood(blob);
-		if (food !is null)
+	if (blob !is null)
+	{	
+		if (this.getSprite().isAnimation("fire"))
 		{
-			food.setVelocity(blob.getVelocity().opMul(0.5f));
+			CBlob@ food = cookFood(blob);
+			if (food !is null)
+			{
+				food.setVelocity(blob.getVelocity().opMul(0.5f));
+			}
 		}
-	}
-	else if (blob !is null && blob.hasTag("fire source")) //fire arrow works
-	{
-		Ignite(this);
+		else if (blob.hasTag("fire source")) //fire arrow works
+		{
+			Ignite(this);
+		}
 	}
 }
 
@@ -125,7 +128,7 @@ f32 onHit(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@ hitt
 	{
 		Extinguish(this);
 	}
-	if (isIgniteHitter(customData))
+	else if (isIgniteHitter(customData)) 
 	{
 		Ignite(this);
 	}
--- a/Entities/Industry/Fireplace/Fireplace.as
+++ b/Entities/Industry/Fireplace/Fireplace.as
@@ -5,6 +5,7 @@
 #include "MakeFood.as";
 #include "FireParticle.as";
 #include "FireCommon.as";
+#include "FireplaceCommon.as";
 #include "Hitters.as";
 
 void onInit(CBlob@ this)
@@ -42,7 +43,6 @@ void onTick(CBlob@ this)
 		Ignite(this);
 	}
 }
-
 
 void onCollision(CBlob@ this, CBlob@ blob, bool solid)
 {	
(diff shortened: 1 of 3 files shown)


Entities/Industry/Fireplace/Fireplace.as
```diff
@@ -53,6 +54,10 @@ void onCollision(CBlob@ this, CBlob@ blob, bool solid)
 			food.setVelocity(blob.getVelocity().opMul(0.5f));
 		}
 	}
+	else if (blob !is null && blob.hasTag("fire source")) //fire arrow works
```
eps0003, line 57: Move the blob null check outside of this if statement so it isn't being repeated by both if statements.

Entities/Industry/Fireplace/Fireplace.as
```diff
@@ -116,9 +121,13 @@ void Ignite(CBlob@ this)
 
 f32 onHit(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@ hitterBlob, u8 customData)
 {
-	if (customData == Hitters::water)
+	if (isWaterHitter(customData)) 
 	{
 		Extinguish(this);
 	}
+	if (isIgniteHitter(customData))
```
eps0003, line 128: Use `else if` since logically something cannot both extinguish and ignite a fireplace.

eps0003 (changes requested):
I am yet to test this in-game.

mugg91:
As said, a fire arrow shot from above the fireplace or from when you stand inside the fireplace, can sometimes miss and not ignite the fireplace. So I added a check to Arrow.as that makes it so that a fire arrow hitting the base of an unlit fireplace will also ignite it.

Just doing SetAnimation("fire") will not properly ignite the fireplace, so I had to copy paste the entire code from fireplace's Ignite(). I didn't manage to find a more clean way. Maybe someone else finds a better way?

eps0003:
> Just doing SetAnimation("fire") will not properly ignite the fireplace, so I had to copy paste the entire code from fireplace's Ignite(). I didn't manage to find a more clean way. Maybe someone else finds a better way?

The way to avoid duplicating code in this case is to do the following:

1. Create `FireplaceCommon.as`
2. Create a function like `void IgniteFireplace(CBlob@ blob)` in this file
3. Move the ignite code into this function
4. Use this function in the two places you need to ignite the fireplace

mugg91:
Thanks eps, I will consider to change it like you said.

Just want to say, I noticed with the current change, an unlit fireplace will get lit by a lantern. Is this ok or bad? Please say your opinions.

eps0003:
> Just want to say, I noticed with the current change, an unlit fireplace will get lit by a lantern. Is this ok or bad? Please say your opinions.

I don't have any opinions on this considering how rare it is to see a fireplace. Feel free to ask others in Discord.

terracraft321:
Doesn't sound bad to me

Entities/Industry/Fireplace/Fireplace.as
```diff
@@ -98,27 +106,18 @@ void Extinguish(CBlob@ this)
...
-	this.getSprite().SetEmitSoundPaused(false);
-	this.getSprite().PlaySound("/FireFwoosh.ogg");
-	
-	CSpriteLayer@ fire = this.getSprite().getSpriteLayer("fire_animation_large");
-	if (fire !is null)
-	{
-		fire.SetVisible(true);
-	}
+	IgniteFireplace(this);
 }
```
eps0003, line 110: You don't need to wrap `IgniteFireplace` in another function. Just use `IgniteFireplace` directly.

Entities/Industry/Fireplace/FireplaceCommon.as
```diff
@@ -0,0 +1,17 @@
+﻿void IgniteFireplace(CBlob@ this)
+{
+	if (this.getSprite().isAnimation("fire")) return;
```
eps0003, line 3: Add a null check for the sprite just to be safe. And get the sprite once instead of getting it multiple times in this function.

Entities/Industry/Fireplace/Fireplace.as
```diff
@@ -42,15 +44,21 @@ void onTick(CBlob@ this)
...
 
-
 void onCollision(CBlob@ this, CBlob@ blob, bool solid)
 {	
-	if (blob !is null && this.getSprite().isAnimation("fire"))
-	{
-		CBlob@ food = cookFood(blob);
-		if (food !is null)
+	if (blob !is null)
```
eps0003, line 49: It would be better to change this to an early return to reduce the indentation.

eps0003 (dismissed):

AsuMagic (approved):

PR #1248, opened 2022-05-01
Cannot toggle lantern underwater, Lantern checks for water every tick

mugg91:
## Status

## Description

This change makes it so that you cannot toggle the lantern on or off (= action 3) when it is in water.
Additionally, I set tickFrequency from 24 to 1, so the game checks if lantern is touching water much more reliably.
Since onTick() only checks for two bools, there should not be a drain on performance.
It is also possible to not set the tickFrequency, since the default seems to be 1.

It should be tested if the "lantern not getting extinguished" bug still happens in online servers with this change.

## Steps to Test or Reproduce

Hold a Lantern and stand in water. It will be extinguished by the water.
Press Action 3. Notice it will not turn light on anymore.

Throw a bunch of lanterns in water and notice they all get extinguished immediately rather than in 0.0 to 0.8 sec.

### DIFF
--- a/Entities/Items/Lantern/Lantern.as
+++ b/Entities/Items/Lantern/Lantern.as
@@ -16,7 +16,7 @@ void onInit(CBlob@ this)
 	this.Tag("ignore fall");
 
 	this.getCurrentScript().runFlags |= Script::tick_inwater;
-	this.getCurrentScript().tickFrequency = 24;
+	this.getCurrentScript().tickFrequency = 1;
 }
 
 void onTick(CBlob@ this)
@@ -46,6 +46,8 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
 {
 	if (cmd == this.getCommandID("activate"))
 	{
+		if (this.isInWater()) return;
+		
 		Light(this, !this.isLight());
 	}
 
--- a/Entities/Items/Lantern/Lantern.as
+++ b/Entities/Items/Lantern/Lantern.as
@@ -16,7 +16,6 @@ void onInit(CBlob@ this)
 	this.Tag("ignore fall");
 
 	this.getCurrentScript().runFlags |= Script::tick_inwater;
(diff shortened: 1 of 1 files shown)


Entities/Items/Lantern/Lantern.as
```diff
@@ -16,7 +16,7 @@ void onInit(CBlob@ this)
 	this.Tag("ignore fall");
 
 	this.getCurrentScript().runFlags |= Script::tick_inwater;
-	this.getCurrentScript().tickFrequency = 24;
+	this.getCurrentScript().tickFrequency = 1;
```
eps0003, line 19: You can remove this line since this is the default value.

eps0003 (dismissed):

Entities/Items/Lantern/Lantern.as
eps0003: On second thought, it might be best to revert this back to 24. Lanterns are very cheap to make from the builder shop so making it tick so often could make it easier/quicker to lag out the server if a player spams tonnes of them. Plus, it's not really that important for it to be extinguished immediately upon contact with water.

mugg91: I tried lagging in offline and with both tick Frequency 1 and 24, it needs about 700 lanterns.
Compare to chicken, which needs about 280 of them to start noticing lag.
So the tick Frequency doesn't seem to make a difference. The blob itself being there is much more a drain on performance, than the function each one executes (which only consists of checking two booleans).
Chicken is much more a performance issue.

I think it is unlikely someone will actually reach 700 lanterns in CTF because they are destroying themselves due to overlapping with one another. And there is not enough space on the map to place 700 ones apart from each other.

This is why I created [ItemLimits.as](https://github.com/transhumandesign/kag-base/pull/1271). Assuming it does become a problem, I could change it to make CTF have its own item limits to cap things like Lantern and Bucket. The default would be infinite.

AsuMagic (approved):

PR #1249, opened 2022-05-02
Dummy takes damage by other classes, some other minor changes

mugg91:
## Status

## Description

This changes the Dummy strawman which appears in the Tutorial (Basics) and in some TDM maps.

* Before this change, Dummy could only get hit by knight attacks (if in a different team), and other classes couldn't interact with it.
* Now, Dummy can now be shot by arrows (if in a different team), can be grappled (any team), can be hit by a builder (any team).
* Changed the way the Dummy rotates when getting hit. It will rotate around the base pole rather than around its own center.
* Changed the y offset from -4 to -8 so the Dummy stands taller.
* Made Dummy able to burn.
* Changed the Map loader, so Dummy is in no team rather than in Red team. I didn't find an easy way to make Dummy in no team while still giving him red color so it will be grey color for now.

Thanks bunnie for helping a noob like me.

I tested this thoroughly, so barring any mistakes, it should be working as described.

## Steps to Test or Reproduce

Go to any map, then in console type /loadmap Ni_TrainingGround_Revised.
Shoot arrow at a Dummy, notice it will get hit.
Hit a Dummy as a builder, notice it will get hit.
Shoot fire arrow at a Dummy, notice it will burn.
Notice the Dummies are placed slightly higher above ground and will rotate around their pole.

### DIFF
--- a/Entities/Special/Dummy/Dummy.as
+++ b/Entities/Special/Dummy/Dummy.as
@@ -2,7 +2,6 @@ void onInit(CBlob@ this)
 {
 	this.getSprite().SetZ(-20.0f);
 	this.getSprite().animation.frame = (this.getNetworkID() * 31) % 4;
-
 	this.SetFacingLeft(((this.getNetworkID() + 27) * 31) % 18 > 9);
 }
 
@@ -31,20 +30,32 @@ void onGib(CSprite@ this)
 		int r = (((blob.getNetworkID() + 1) * 31) % 3);
 		CParticle@ Large1   = makeGibParticle(filename, pos, vel + getRandomVelocity(90, hp - 0.2 , 80), 7 - (r % 2), 1 - (r / 2), Vec2f(16, 16), 2.0f, 20, "/material_drop", team);
 	}
+}
 
+bool doesCollideWithBlob(CBlob@ this, CBlob@ blob)
+{	
+    if (blob.getName() == "arrow" && this.getTeamNum() != blob.getTeamNum())
+    {
+        return true;
+    }
+    return false;
 }
 
 f32 onHit(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@ hitterBlob, u8 customData)
 {
 	if (damage > 0.0f)
-	{
+	{	
+		CSprite@ sprite = this.getSprite();
+	
 		if (worldPoint.x > this.getPosition().x)
 		{
-			this.setAngleDegrees(this.getAngleDegrees() - 3 - XORRandom(10));
+			//this.setAngleDegrees(Maths::Max(this.getAngleDegrees() - 3 - XORRandom(10), -30));
+			sprite.RotateBy(-2 - XORRandom(7), Vec2f(0.0f, 12.0f));
 		}
 		else
 		{
-			this.setAngleDegrees(this.getAngleDegrees() + 3 + XORRandom(10));
+			//this.setAngleDegrees(Maths::Min(this.getAngleDegrees() + 3 + XORRandom(10), 30));
+			sprite.RotateBy(2 + XORRandom(7), Vec2f(0.0f, 12.0f));
 		}
 	}
 	return damage;
--- a/Entities/Special/Dummy/Dummy.cfg
+++ b/Entities/Special/Dummy/Dummy.cfg
@@ -76,7 +76,6 @@ $name                                      = dummy
 						IsFlammable.as;
 						Wooden.as;
 						GenericHit.as;
-						NoPlayerCollision.as;
 						AlignToTiles.as;
 f32_health                                 = 2.0
 # looks & behaviour inside inventory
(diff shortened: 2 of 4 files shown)


Entities/Special/Dummy/Dummy.as
```diff
@@ -31,20 +30,32 @@ void onGib(CSprite@ this)
...
 	}
+}
 
+bool doesCollideWithBlob(CBlob@ this, CBlob@ blob)
+{	
+    if (blob.getName() == "arrow" && this.getTeamNum() != blob.getTeamNum())
+    {
+        return true;
+    }
+    return false;
```
eps0003, line 41: You can just return what is in the if statement.

Entities/Special/Dummy/Dummy.as
```diff
@@ -31,20 +30,32 @@ void onGib(CSprite@ this)
...
 f32 onHit(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@ hitterBlob, u8 customData)
 {
 	if (damage > 0.0f)
-	{
+	{	
+		CSprite@ sprite = this.getSprite();
+	
 		if (worldPoint.x > this.getPosition().x)
 		{
-			this.setAngleDegrees(this.getAngleDegrees() - 3 - XORRandom(10));
```
eps0003, line 43: Remove this line rather than commenting it out. Same with the other line lower down.

eps0003 (changes requested):

Entities/Special/Dummy/Dummy.as
mugg91: Will do. I just wanted to convey this alternative way of doing it, too.
RotateBy() doesn't use angle, so getAngleDegrees() always returns 0. Therefore I cannot implement a threshold after which the Dummy no longer tilts further.

mugg91:
In order to make Dummy have red color (which is better than grey color), we could make it neutral to all teams by always receiving damage by slashes/jabs/arrows, regardless what team the Dummy and the attacker is on. I'd like to see opinions.

eps0003:
That sounds good, but their color should still match the team they are set to so it can be customised. The dummies in the tutorial should have their team set to red team because they are used to teach the player to attack the enemy (the player always spawns on blue team in the tutorial iirc).

mugg91:
> That sounds good, but their color should still match the team they are set to so it can be customised. The dummies in the tutorial should have their team set to red team because they are used to teach the player to attack the enemy (the player always spawns on blue team in the tutorial iirc).

This PR sets dummy to spawn in team -1. This was done so it would receive damage from both teams (good for TDM).
But team -1 means its grey color which is lame. If Dummy is made to spawn in Red Team, then, with these changes, it would receive damage from Blue Team but not from Red Team (bad for TDM).

Ideas to go forward:

- Make Dummy spawn in Red Team, but receive damage from any team - I don't think any other entity does this, so eh...
- Or make Dummy spawn in no team, make it change to Red Team in the tutorial maps (via rules script).
It remains neutral and grey colored in TDM.
- Or make Dummy spawn in no team but have red color - you disliked this idea.

mugg91:
In this PR, right now, Knight and Archer can hit enemy Dummy. Arrows will get stuck to the enemy Dummy. Therefore it is important to make maps fair to not favor a team over the other.

BasePNGLoader will load Dummy as team 255.

I made a commit just now which will have TDM.as change Dummy to team 2-6.
- In TDM there is team 0 and 1, so when Dummy is team 2-6 there is no problem.
- Also random colored Dummy is nice.

Challenge.as will check if the loaded map contains string "Tutorial", if yes, change Dummy to team 1.
- Red colored Dummy, as intended for Tutorial01.

Of course there are other ways to do it. Like Tag("tutorial") in Tutorial01.as and then checking for that. Let me know what is the best way of doing it.

I'm not sure if there is Dummy in CTF and TTH maps. They will still be grey. Their rules script can be changed accordingly to make Dummy random colored there, too.

Entities/Special/Dummy/Dummy.cfg
```diff
@@ -7,12 +7,13 @@
...
 @$sprite_scripts                           = Dummy.as;
-                      						 Wooden.as;
+                      				Wooden.as;
+						FireAnim.as;
 $sprite_texture                            = Dummy.png
 s32_sprite_frame_width                     = 24
 s32_sprite_frame_height                    = 24
 f32 sprite_offset_x                        = 0
-f32 sprite_offset_y                        = -4
+f32 sprite_offset_y                        = -8
```
AsuMagic, line 16: This makes their hitbox severely mismatch their sprite, which looks weird given that they can now be shot:
image

Not sure what would be the best way to fix this. The circle shape also makes stuck arrows look derpy.

AsuMagic (changes requested):

Entities/Special/Dummy/Dummy.cfg
Gingerbeard5773: the proper way to fix this would be to set the sprite offset back to normal and then offset the blob's spawn position in map loading so the blob

mugg91:
I'm not looking at this anytime soon but might eventually come back to this.

PR #1254, opened 2022-05-09
Shark doen't show seats or healthbar when in inventory

mugg91:
## Status

## Description

When putting Shark in an inventory (Storage/Crate/Dinghy/Longboat/Warboat/Bomber/Airship), the shark's seat arrow and healthbar won't be shown. 

## Steps to Test or Reproduce

Go to Sandbox, !shark, carry it, !crate, put shark in crate, notice there is no seat arrow and no healthbar.

### DIFF
--- a/Entities/Common/GUI/HealthBar.as
+++ b/Entities/Common/GUI/HealthBar.as
@@ -6,6 +6,10 @@ void onRender(CSprite@ this)
 		return;
 
 	CBlob@ blob = this.getBlob();
+	
+	if (blob.isInInventory())
+		return;
+	
 	Vec2f center = blob.getPosition();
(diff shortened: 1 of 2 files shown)


eps0003 (approved):
Irrelevant to your changes: I'm curious as to why there is no null check for the blob in these two places, but I guess it is not needed if they have been working without error so far.

mugg91:
I made another commit that will make it so Seat arrow won't show up on Dinghy when it is picked up.
I figured this was similar and small enough to include it in this PR.

Gingerbeard5773 (approved):
Works as intended

AsuMagic (approved):

PR #1210, opened 2022-03-27
Chat command manager

eps0003:
## Status

- **READY**: this PR is (to the best of your knowledge) ready to be incorporated into the game.

## Description

This is a complete rewrite of `ChatCommands.as`. It splits each command into its own file and makes it easy to write client-side and server-side chat commands. It also enables mods or specific gamemodes to register their own commands without needing to modify or overwrite any of the command handling code.

Server owners can customise which commands are enabled, admin only and `sv_test` only by copying `ChatCommands.cfg` to the Cache folder and modifying it.

This also removes the ability to send empty messages by typing only spaces.

## New commands

- `!help` - lists available commands
- `!tip [tip #]` - shows useful tips
- `!class <name>` - change class
- `!heal [amount]` - heal yourself
- `!spawn <blob>` - spawns blobs (used to be `!<blob>`)
- `!nextmap` - load next map
- `!restartmap` - restart current map

## Steps to Test or Reproduce

Type `!help` in chat

PR #1255, opened 2022-05-09
Spawn Sign with your own text in Sandbox

mugg91:
## Status

Will set to "READY" after addressing these:
- Make the game look for bad words in the Sign's text, don't spawn it or censor it.
- Support line breaks with /n like in "TutorialCTF01.as".
- Make sign hittable by builder.
- Default to the shaggy chocolate text when no text is found after "!sign".
- Fix that text can be out of bounds if a word is too long.
- Fix that an icon can be out of bounds. 

## Description

Before this change, in Sandbox, you could type "!sign" and it spawned a sign saying "The big brown fox jumped over the shaggy chocolate". That's all text you could have. "!sign blah" didn't do anything.

This change makes it so !sign lets you put your own text when spawning a sign in Sandbox.
Type in "!sign your text blah blah $icon$".

https://i.imgur.com/0TFCMoa.png

Let me know what you think.

### DIFF
--- a/Rules/CommonScripts/ChatCommands.as
+++ b/Rules/CommonScripts/ChatCommands.as
@@ -6,6 +6,7 @@
 #include "MakeSeed.as";
 #include "MakeCrate.as";
 #include "MakeScroll.as";
+#include "MakeSign.as";
 
 const bool chatCommandCooldown = false; // enable if you want cooldown on your server
 const uint chatCommandDelay = 3 * 30; // Cooldown in seconds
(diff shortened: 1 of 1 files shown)


mugg91:
For future reference, the current build lets you spawn crate with a custom description that is not checked for swear words.

So if I or anyone ever do manage to check for swear words in a self-created custom Sign, then Crates might as well get checked, too.

Rules/CommonScripts/ChatCommands.as
```diff
@@ -302,11 +303,16 @@ bool onServerProcessChat(CRules@ this, const string& in text_in, string& out tex
...
-				else if(tokens[0] == "!coins")
+				else if (tokens[0] == "!coins")
 				{
 					int money = parseInt(tokens[1]);
 					player.server_setCoins(money);
 				}
+				else if (tokens[0] == "!sign")
+				{
+					CMap@ map = getMap();
+					createSign(pos, text_in.substr(tokens[0].length+1, -1));
```
eps0003, line 314: The default value for the second argument is -1 so this line can be simplified to the following:
```suggestion
					createSign(pos, text_in.substr(tokens[0].length + 1));
```

eps0003 (changes requested):

AsuMagic:
Will conflict with #1210

mugg91:
This is currently not anywhere near done and there are issues I ran into that I haven't figured out to fix. It would seem some things are engine-side and cannot be done anyway. Please ignore (or close if you must) this PR for the time being.

Issue #1252, opened 2022-05-08
Saw kills you when changing class


mugg91:
**Description**

If you hold a Saw and change class at a Tent or TDM Spawn, there is a chance you will get killed by your Saw.
It doesn't seem to happen when changing class at a building.
If the Saw is lying on the ground, the chance is higher.

On a related note, it may be worth to discuss if Saw lying at a tunnel should kill anyone exiting it. IMO, it makes more sense if the exiting person falls for a bit (1 or 2 blocks deep) before being eligible of getting killed by a Saw.
The main issue is what I wrote above, though. It happened to me in CTF and I confirmed it while testing in Sandbox.

**Reproduction**

Hold a Saw while changing class at a Tent or TDM Spawn.
If the Saw doesn't kill you, you can try to jump once, land on the ground, try again.

In this video it happens twice:
https://www.youtube.com/watch?v=l-cN5sSjISU

**Expected Behavior**

It should not kill you. Place stuff you were holding back in your hand, or disable Saw for a few ticks?

PR #1257, opened 2022-05-13
Saw doesn't kill players on the same floor when they change class or use tunnel

mugg91:
## Status

## Description

Fixes [issue #1252](https://github.com/transhumandesign/kag-base/issues/1252).

Player will not get killed by an active Saw on the same ground when changing class or using tunnel.
If a player or a friendly mine is touching the Saw blade but has a velocity of 0.5 or less, then they are not getting sawed.

https://i.imgur.com/jnw8Ink.png

### DIFF
--- a/Entities/Industry/Saw/Saw.as
+++ b/Entities/Industry/Saw/Saw.as
@@ -182,7 +182,7 @@ bool canSaw(CBlob@ this, CBlob@ blob)
 
 		f32 dot = off * (Vec2f(0, -1).RotateBy(this.getAngleDegrees(), Vec2f()));
 
-		if (dot > 0.8f && blob.getVelocity().y > 0.5f)
+		if (dot > 0.8f && (blob.getVelocity().y - this.getVelocity().y) > 0.1f)
 		{
 			if (getNet().isClient() && !g_kidssafe) //add blood gfx
 			{
--- a/Entities/Industry/Saw/Saw.as
+++ b/Entities/Industry/Saw/Saw.as
@@ -182,7 +182,7 @@ bool canSaw(CBlob@ this, CBlob@ blob)
 
 		f32 dot = off * (Vec2f(0, -1).RotateBy(this.getAngleDegrees(), Vec2f()));
 
-		if (dot > 0.8f)
+		if (dot > 0.8f && blob.getVelocity().y > 0.5f)
 		{
 			if (getNet().isClient() && !g_kidssafe) //add blood gfx
 			{


Mazey:
Has this been thoroughly tested? Code-wise this feels a bit too general and there may be side effects.

mugg91:
This has been tested locally with some cases like seen in the image, but not in crowded matches yet.

The first check `dot > 0.8f` is (as far as I understand) to see if a player is touching the Saw blade. The second check I added is to see if the player is moving downward ever so slightly.

Upon checking again, I did notice that the behavior is not as seen as in the image. I apologize.
When travelling to a tunnel with a Saw on the ground, or when the Saw is 1 tile deeper, then y velocity of player is ~0.867 in both cases so the kill is still happening.

Here is the updated image of how it's behaving after the change: https://i.imgur.com/yNqkuPi.png

Since "Saw on the same ground as tunnel" and "Saw 1 tile deeper than the tunnel" kill players at the same y velocity (~0.867), the kill in the former case cannot be prevented based on y velocity.

We could discuss if it's fine like this, or if we want to prevent kills for when the Saw is on the same ground as the tunnel.

laws65:
This stops you from dying when someone throws a saw up to you. Like if the saw was thrown in the picture, the knights wouldn't die.
image

mugg91:
I changed it to `if (dot > 0.8f && (blob.getVelocity().y - this.getVelocity().y) > 0.1f)`. The knights are dead.

Entities/Industry/Saw/Saw.as
```diff
@@ -182,7 +182,7 @@ bool canSaw(CBlob@ this, CBlob@ blob)
 
 		f32 dot = off * (Vec2f(0, -1).RotateBy(this.getAngleDegrees(), Vec2f()));
 
-		if (dot > 0.8f)
+		if (dot > 0.8f && (blob.getVelocity().y - this.getVelocity().y) > 0.1f)
```
Mazey, line 185: Simply checking velocity for any blob is not a reliable way to fix this and may cause too many (unforeseen) issues.

Mazey (changes requested):

Entities/Industry/Saw/Saw.as
mugg91: I found that player can stand on ledge and slowly fall into a Saw without getting killed, because their velocity will be less than 0.1f. This can be fixed by changing the check to:

`blob.getVelocity().y - this.getVelocity().y) > 0.01f`

I have not found any more issues. Climbing down a ladder into the Saw will also kill the player as expected.
Are there cases where a player should get killed without moving towards the Saw?

There can be unforeseen issues with any change. That's why play-testing needs to be done.

What else can be a fix for "getting killed while changing class"?
You could make the player invincible for the first one or five ticks after changing class. Should we pursue that solution instead?

Mazey: At least checking if the blob name is knight, archer, or builder would already eliminate a lot potential issues with blobs that interact with the saw, for example bombs or trees/logs. Regardless, relying on the velocity is not how this should be approached.

mugg91: There are checks to see if the blob is tagged "flesh" or if it is an enemy mine, before the code reaches the velocity check.
Bombs/trees/logs are unaffected.

mugg91:
Devs were not happy with this solution which added a velocity check to Saw.

My suggestion is, to fix Saw killing players in erroneous ways, players should be given a very short-lasting invincibility time. You could add this invincibility when changing class at a tent. And if deemed necessary, you could add this invincibility when traveling via tunnel so you don't get killed by a Saw that is not even below you.

Issue #1233, opened 2022-04-17
Can't use tunnel while carrying a dinghy


mugg91:
**Description**

While carrying a dinghy, you can go through a tunnel if there are 2 tunnels in your team.
But if there are 3+ tunnels, it is not possible to click to select the tunnel you want to go to.
This happened in CTF and I verified it in Sandbox.

**Reproduction**

1. Have two tunnels in your team.
2. Carry a dinghy.
3. Use the tunnel. Notice that there is no problem.
4. Have another tunnel added to your team, you now have 3 tunnels or more.
5. Use the tunnel. Notice that you cannot go through the tunnel anymore.

**Expected Behavior**

Tunnel should be possible to use while carrying a dinghy while there are 3 tunnels or more.

mugg91:
You also cannot click the class icons when changing class while carrying a dinghy.
Note that you can click on the tunnel and class selection when carrying anything else (keg, crate), it's just the dinghy not working.

mugg91:
At line 35 in kag-base/Entities/Vehicles/Common/BoatCommon.as :

`			// disable acion keys when carrying this
			if (oar.name == "PICKUP")
			{
				oar.SetKeysToTake(key_action1 | key_action2);
			}`

If this code is removed, you could travel through tunnel or change class at a tent while carrying a Dinghy.
But Knight would be able to slash and shield while carrying Dinghy.
I didn't manage to make a fix.

mugg91:
Some further suggestions.

Dinghy:
- 1000 shape mass (set in Dinghy.cfg)
- "medium weight" (set in Dinghy.as)

Raft:
- 800 shape mass (Raft.cfg)
- "heavy weight" (Raft.as)

Dinghy and Raft should possibly have their shape mass values switched, so the medium/heavy weight thing makes sense.

Raft should have BoatCommon.as added to Raft.cfg (scripts section) to block players from using slashes, shield, etc. while carrying a Raft. This is to make behavior between Dinghy and Raft more consistent.

PR #1260, opened 2022-05-14
Can use tunnel, shops and change class while carrying Dinghy

mugg91:
## Status

## Description

This fixes [issue #1233](https://github.com/transhumandesign/kag-base/issues/1233).

- Dinghy is set from "medium weight" to "heavy weight" and mass reduced from 1000 to 800. This fixes the inconsistency that Raft had more mass while being considered less weight.
- Disabling action keys is now done in RunnerMovement.as instead of BoatCommon.as. 
It applies not just for Dinghy, but for "heavy weight" tagged things in general now (Dinghy, Raft, Crate with exploding Keg in it).
- You can do one action every 30 ticks, therefore allowing using shops and tunnels, while still disabling slashes/bow.

Applying the key disable for heavy things in general made the most sense to me.
Setting Dinghy to "heavy weight" had to be done, otherwise why disable keys for Dinghy but not for other "medium weight" things like Keg or Boulder?
This also means that Dinghy now reduces your movement speed more, makes heavy wood sounds and may have slightly different logic than before.

Please let me know what you think.
Alternatively, it is also possible to set Dinghy to "medium weight" but then allow action keys again.

## Steps to Test or Reproduce

1. Go to Sandbox.
2. Build three tunnels, some shops and tent.
3. Pick up a Dinghy into your hand.
4. Notice that you can use the tunnels, the shops and change class at tent without problems.


AsuMagic:
Will put merging this on hold since I am not convinced about that approach:

> You can do one action every 30 ticks, therefore allowing using shops and tunnels, while still disabling slashes/bow.

mugg91:
Yeah, not a good approach.

PR #1261, opened 2022-05-14
[dev] Storage shop icon fixed in Building frame (full of goodies again)

MrPotatobird:
## Status

## Description

makes it look like this again 
image

## Steps to Test or Reproduce

go to build a storage


eps0003 (approved):
Looks better but could potentially mislead new players into thinking it spawns stocked with these items.

AsuMagic:
I agree, IMO it should be left empty looking. I feel like the materials would induce confusion and sort of add to the visual noise (e.g. I feel like such resources or gold should not be shown when not strictly relevant). I'll close this for now.  
That said it might not be too obvious that the current sprite looks like a storage, so other improvements could be worthwhile. Not sure what though...

PR #1262, opened 2022-05-16
[dev] 8x8 icons are centered in inventory HUD

mugg91:
## Status

## Description

This change renders items with 8x8 pixeled icons (e.g. grain, lantern, egg) in the center of a slot rather than in the top left.

`(16-item.inventoryFrameDimension.x)` and `(16-item.inventoryFrameDimension.y)` are used. There is no item whose icon will show up in the HUD that has a greater width or height than 16.

Before: https://i.imgur.com/zDWK792.png
After: https://i.imgur.com/3hqteBK.png

### DIFF
--- a/Entities/Common/GUI/ActorHUDStartPos.as
+++ b/Entities/Common/GUI/ActorHUDStartPos.as
@@ -34,8 +34,8 @@ void DrawInventoryOnHUD(CBlob@ this, Vec2f tl)
 			const int quantity = this.getBlobCount(name);
 			drawn.push_back(name);
 
-			GUI::DrawIcon(item.inventoryIconName, item.inventoryIconFrame, item.inventoryFrameDimension, tl + Vec2f(0 + (drawn.length - 1) * 40, -6), 1.0f);
-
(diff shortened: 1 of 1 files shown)


mugg91:
I was going to add more code for safety. So if there is an icon with more than 16 px width or height, it is not getting displaced more than it should:

`int offsetx = Maths::Max((16-item.inventoryFrameDimension.x), 0);
int offsety = Maths::Max((16-item.inventoryFrameDimension.y), 0);
GUI::DrawIcon(item.inventoryIconName, item.inventoryIconFrame, item.inventoryFrameDimension, tl + Vec2f((drawn.length - 1) * 40 + offsetx, -6 + offsety), 1.0f)`

But then I found that Drill actually does have 32 px width. The current change (without the safety fix) causes it to be pushed to the left, seemingly making it centered, although it is an unintentional side-effect. The safety fix would make it look like in the current public game.

Before: https://i.imgur.com/56oTSl2.png
After: https://i.imgur.com/XTvkVcl.png

Either the side-effect can be accepted or (imo, the better way) Drill's icon needs to be adjusted to be 16x16 pixels.

mugg91:
I figure, the safety check isn't necessary. As it is now, it will center the items in the slot, regardless their width or height. So it is working as I intended after all.
So this PR is ready for review.

AsuMagic:
closed by #1570

Issue #1216, opened 2022-04-03
Cannot select every tunnel if there are too many tunnels


mugg91:
**Description**

When there are a lot of tunnels and trying to use one, the selection window will have a too high width and you won't be able to select the tunnels on the far left.

**Reproduction**

1. Build more than 15 tunnels. (Playing at 1280x720)
2. Use a tunnel.
3. Notice that you cannot click to go to the leftmost tunnels.

**Expected Behavior**

Either the HUD should show all the tunnels (maybe by reducing width of the buttons) or the game should limit the amount of tunnels you can build or spawn.

**Screenshots**

https://i.imgur.com/w07qyPn.png

**Additional context**

Posting this issue at Nananas' request.

mugg91:
Just want to mention, same happens if there are too many ballistas and you are given the choice where to spawn.

PR #1269, opened 2022-05-26
Fix item spam causing lag in Sandbox

mugg91:
## Status

**READY**: this PR is (to the best of your knowledge) ready to be incorporated into the game.

## Short Description

- Can't keep spamming !fishyschool, !chickenflock or other items to lag the server.
- War_base is now blacklisted so you can't spam it to grief in online Sandbox. 
- Buying an item that is limited will not subtract money from you.

## Long Description

This change adds a new rules script ItemLimits.as which is added to Sandbox/gamemode.cfg (scripts section).
Instead of checking for only two items (keg and mine) and imposing a general limit of 20 in Sandbox_Rules.as, all items are counted per team in ItemLimits.as.
So e.g. Blue Team can have 20 kegs and 20 tunnels, Red Team can have their own 20 kegs and 20 tunnels. Etc.
Basicly, this change takes what Sandbox_Rules.as was already doing to the next level.

The maximum allowed per item per team is fetched from ItemLimits.cfg. The default is 200 but I have set most items to lower values, please refer to ItemLimits.cfg to get an understanding.

Materials are not counted as I would like them to disappear based on their decay timer. Gold is currently the only material not decaying so that still needs addressing, otherwise it works as intended.

Builder blobs are not counted because I don't want to prevent someone from filling the whole map with doors, sensors, etc. They are not causing lag in a way that e.g. chicken do, anyway.

This change also adds war_base to the blacklisted items due to griefing (it can destroy the ground beneath you, like the already blacklisted hall could). Tent and TDM_Spawn will also be limited to 20, so the whole map can't be spammed with it.

If adjustments are made, ItemLimits.as could also be added to other game modes, but I have not seen item spam being a problem in CTF/TTH/TDM. This change is just so that Sandbox servers can't be made unplayable by someone spamming !chickenflock, !fishyschool or !war_base.

Shop.as will fetch ItemLimits.as' item counts to check if an item that is about to be bought is exceeding its limit. If yes, the item will not get created and no money is subtracted either. Otherwise, ItemLimits.as will despawn items if they exceed their limit (e.g. when there are 20 kegs and you try !keg). In both cases, a warning in chat lets the player know what happened. 
https://i.imgur.com/Y49xp74.png
https://i.imgur.com/YGg0a0S.png

Removed `this.server_setTeamNum(0); // blue fishy like in sprite sheet;` from Food.as onInit() so that burgers bought in quarters will be properly assigned to the caller's team, so as to make the item counting work properly. I didn't see any side-effects since Fishy sprite in inventory is still blue when you are in a team other than Blue Team.

When a map is loaded, ItemLimits.as will count items already on the map.
/restartmap or /loadmap will properly reset the item counts.

This PR is in a state where I assume all side-effects and crashes are fixed. If you do encounter any issues, please let me know.

## Additional Information

Thanks to all the people on #modding helping me.

### DIFF
--- a/Entities/Common/Factory/Shop.as
+++ b/Entities/Common/Factory/Shop.as
@@ -138,35 +138,6 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
 					}
 				}
 			}
-			
-			/* Method 2:
-			Use this if we want to not rely on dictionary itemLimits in ItemLimits.as
-			and do item counting from scratch:
-			
-			CBlob@[] blobsInMap;
-			getBlobsByName(itemName, @blobsInMap);
-			int blobCounts = 0;
-			
-			int callerTeam = caller.getTeamNum();
-			if (callerTeam < 0 || callerTeam > 7) callerTeam = 8;
-			
-			ConfigFile cfg = ConfigFile();
-			cfg.loadFile("Rules/CommonScripts/ItemLimits.cfg");
-			s32 maximum = cfg.read_s32(itemName, 200);
-			
-			for (uint b = 0; b < blobsInMap.length(); ++b)
-			{
-				int blobTeam = blobsInMap[b].getTeamNum();
-				if (blobTeam < 0 || blobTeam > 7) blobTeam = 8;
-				if (callerTeam == blobTeam) blobCounts++;
-			}
-
-			if (blobCounts >= maximum && maximum > 0) 
-			{	
-				sendChatWarningLimitedItem(maximum, itemName);
-				return;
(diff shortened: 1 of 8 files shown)


mugg91:
This can be a semi-fix to issue #1216 if the tunnel limit is set to 12 or less. An actual fix would still include a GUI change, though.

mugg91:
I have noticed there actually is a side-effect when food isn't set to team 0. The cooked fish will have grey color.
I'm working on a different implementation of item limits so hang tight. This one here will probably get abandoned.

PR #1270, opened 2022-05-27
Crate will change team to unpacker, Crate unpacking cannot be done from afar

mugg91:
## Status

## Description

When a crate is unpacked, it will change its team to the player who started unpacking.

When a player tries to click on "unpack" or "stop unpacking" buttons, he will have to be inside 20.0 radius. You cannot click those buttons from a screen length away anymore.

## Steps to Test or Reproduce

1. Go to Sandbox.
2. Switch to Blue Team.
3. !crate catapult
4. Pick up the crate and place it down. It should change to Blue Team.
5. Switch to Red Team.
6. Show buttons and click on "Unpack".
7. The crate will unpack a Blue Catapult. (After this change, it would instead be a Red Catapult).

---

1. Go to Sandbox.
2. !crate catapult
3. Keep showing "unpack" button and try to walk away.
4. Notice you can still click on the button even when you are a half screen length away. (After this change, you can only click on it if you are next to the crate.)


Entities/Items/Crate/Crate.as
```diff
@@ -328,6 +332,9 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
 			{
 				this.set_u32("unpack time", getGameTime() + this.get_u32("unpack secs") * getTicksASecond());
 				this.getShape().setDrag(10.0f);
+
+				CBlob @caller = getBlobByNetworkID( params.read_u16() );
```
eps0003, line 336: I hate this line so much 😂 
1. `@` goes right after the type, not before the variable name
2. Don't put whitespace around function arguments
```suggestion
				CBlob@ caller = getBlobByNetworkID(params.read_u16());
```

Entities/Items/Crate/Crate.as
```diff
@@ -294,11 +294,15 @@ void GetButtonsFor(CBlob@ this, CBlob@ caller)
...
 	{
-		caller.CreateGenericButton("$DISABLED$", buttonpos, this, this.getCommandID("stop unpack"), getTranslatedString("Stop {ITEM}").replace("{ITEM}", getTranslatedString(this.get_string("packed name"))));
+		CButton@ button = caller.CreateGenericButton("$DISABLED$", buttonpos, this, this.getCommandID("stop unpack"), getTranslatedString("Stop {ITEM}").replace("{ITEM}", getTranslatedString(this.get_string("packed name"))));
+		button.enableRadius = 20.0f;
 	}
 	else if (hasSomethingPacked(this))
 	{
-		caller.CreateGenericButton(12, buttonpos, this, this.getCommandID("unpack"), getTranslatedString("Unpack {ITEM}").replace("{ITEM}", getTranslatedString(this.get_string("packed name"))));
+		CBitStream params;
+		params.write_u16( caller.getNetworkID() );
```
eps0003, line 303:
```suggestion
		params.write_u16(caller.getNetworkID());
```

Entities/Items/Crate/Crate.as
```diff
@@ -328,6 +332,9 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
 			{
 				this.set_u32("unpack time", getGameTime() + this.get_u32("unpack secs") * getTicksASecond());
 				this.getShape().setDrag(10.0f);
+
+				CBlob @caller = getBlobByNetworkID( params.read_u16() );
+				if (caller !is null) this.server_setTeamNum( caller.getTeamNum() );
```
eps0003, line 337: Most of the time you should put if statements over multiple lines like so to avoid lines from getting too long. Exceptions include short if statements, the most common being `if (x is null) return;`.
```suggestion
				if (caller !is null)
				{
						this.server_setTeamNum(caller.getTeamNum());
				}
```

Entities/Items/Crate/Crate.as
```diff
@@ -294,11 +294,15 @@ void GetButtonsFor(CBlob@ this, CBlob@ caller)
 	}
 	else if (isUnpacking(this))
 	{
-		caller.CreateGenericButton("$DISABLED$", buttonpos, this, this.getCommandID("stop unpack"), getTranslatedString("Stop {ITEM}").replace("{ITEM}", getTranslatedString(this.get_string("packed name"))));
+		CButton@ button = caller.CreateGenericButton("$DISABLED$", buttonpos, this, this.getCommandID("stop unpack"), getTranslatedString("Stop {ITEM}").replace("{ITEM}", getTranslatedString(this.get_string("packed name"))));
```
eps0003, line 297: This line is pretty long and hard read. I suggest you move the button text out into a variable since that itself is quite long. Please do the same for the other line you added further down.
```suggestion
		string text = getTranslatedString("Stop {ITEM}").replace("{ITEM}", getTranslatedString(this.get_string("packed name")));
		CButton@ button = caller.CreateGenericButton("$DISABLED$", buttonpos, this, this.getCommandID("stop unpack"), text);
```

eps0003 (changes requested):
Apologies for being nitpicky for this review 🙏

Entities/Items/Crate/Crate.as
mugg91: I have seen white-space being done in `if ( )` statements for readability.
Should it not be done at all, or in if statements only?

Entities/Items/Crate/Crate.as
mugg91: While applying your change, I noticed the item name is not getting translated.
`!crate catapult` will say "Depaqueter catapult" instead of "Depaqueter catapulte" in french.
Isn't getTranslatedString() supposed to translate the string to the local language?

Gingerbeard5773: That's because you spawned it with ``!crate`` and didn't set the name. Translation doesn't work with wrong capitals

AsuMagic (approved):

PR #1273, opened 2022-05-28
Can't change class at enemy Ballista, Can't get resupplies at enemy Tent

mugg91:
## Status

## Description

- You can't change class at an enemy Ballista and in doing so gain resupplies from it, 
as was already the case with enemy Tent and enemy Warboat.
You can still change class at any TDM_Spawn and any shop regardless of Team.

- You can't flip over an enemy Ballista or enemy Catapult.
Doing so would almost always kill you from collision, so it is removed now.

- You can't get resupplies at an enemy Tent.

- You can get resupplies by overlapping with a friendly Warboat or friendly Ballista.
This is because changing class gives materials, so why not just give them automatically when overlapping.

---

Please note, you can still gain resupplies from any shop.
Because if you can buy from and change class at enemy Shops, the shop keeper is probably dumb and will also give you mats.

## Reproduction Steps

1. Be in Blue Team.
2. Go to a Red Ballista and show buttons.
3. Notice there is a "Change Class" button, using it would give you resupplies. **After this change, there is no button.**

---

1. Be in Blue Team.
2. Spawn a Ballista or Catapult and make it fall over.
3. Switch to Red Team and show buttons.
4. Notice there is a "Flip" button, using it would probably kill you. **After this change, there is no button.**

---

1. Be in Blue Team and be a builder.
2. Go to a Red Tent.
3. Notice you can't switch class at the enemy Tent but you will get resupplies. **After this change, you won't get resupplies.**

---

1. Be in Blue Team and be a builder.
2. Go to a Blue Ballista or Blue Warboat.
3. Notice nothing happening. **After this change, you will get automatic resupplies at those entities.**

### DIFF
--- a/Entities/Vehicles/Ballista/Ballista.as
+++ b/Entities/Vehicles/Ballista/Ballista.as
@@ -232,22 +232,24 @@ void onTick(CBlob@ this)
 			}
 		}
 	}
-
 }
 
 void GetButtonsFor(CBlob@ this, CBlob@ caller)
 {
 	if (!canSeeButtons(this, caller)) return;
 
-	if (isOverlapping(this, caller) && !caller.isAttached())
+	if (caller.getTeamNum() == this.getTeamNum())
 	{
-		if (!Vehicle_AddFlipButton(this, caller) && caller.getTeamNum() == this.getTeamNum())
+		if (isOverlapping(this, caller) && !caller.isAttached())
 		{
-			Vehicle_AddLoadAmmoButton(this, caller);
-		}
-		if (/*!isAnotherRespawnClose(this) &&*/ !isFlipped(this))
-		{
-			caller.CreateGenericButton("$change_class$", Vec2f(0, 1), this, buildSpawnMenu, getTranslatedString("Change class"));
+			if (!Vehicle_AddFlipButton(this, caller))
+			{
+				Vehicle_AddLoadAmmoButton(this, caller);
+			}
+			if (/*!isAnotherRespawnClose(this) &&*/ !isFlipped(this))
+			{
+				caller.CreateGenericButton("$change_class$", Vec2f(0, 1), this, buildSpawnMenu, getTranslatedString("Change class"));
+			}
 		}
 	}
 }
--- a/Entities/Vehicles/Ballista/Ballista.as
+++ b/Entities/Vehicles/Ballista/Ballista.as
@@ -238,18 +238,17 @@ void GetButtonsFor(CBlob@ this, CBlob@ caller)
 {
 	if (!canSeeButtons(this, caller)) return;
 
-	if (caller.getTeamNum() == this.getTeamNum())
+	if (caller.getTeamNum() == this.getTeamNum()) return;
+	
+	if (isOverlapping(this, caller) && !caller.isAttached())
 	{
-		if (isOverlapping(this, caller) && !caller.isAttached())
+		if (!Vehicle_AddFlipButton(this, caller))
 		{
-			if (!Vehicle_AddFlipButton(this, caller))
-			{
-				Vehicle_AddLoadAmmoButton(this, caller);
-			}
-			if (/*!isAnotherRespawnClose(this) &&*/ !isFlipped(this))
-			{
-				caller.CreateGenericButton("$change_class$", Vec2f(0, 1), this, buildSpawnMenu, getTranslatedString("Change class"));
-			}
+			Vehicle_AddLoadAmmoButton(this, caller);
+		}
+		if (/*!isAnotherRespawnClose(this) &&*/ !isFlipped(this))
+		{
+			caller.CreateGenericButton("$change_class$", Vec2f(0, 1), this, buildSpawnMenu, getTranslatedString("Change class"));
 		}
 	}
 }
--- a/Entities/Vehicles/Ballista/Ballista.as
+++ b/Entities/Vehicles/Ballista/Ballista.as
@@ -238,7 +238,7 @@ void GetButtonsFor(CBlob@ this, CBlob@ caller)
 {
 	if (!canSeeButtons(this, caller)) return;
 
-	if (caller.getTeamNum() == this.getTeamNum()) return;
+	if (caller.getTeamNum() != this.getTeamNum()) return;
 	
 	if (isOverlapping(this, caller) && !caller.isAttached())
 	{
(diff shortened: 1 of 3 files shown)


Entities/Vehicles/Catapult/Catapult.as
```diff
@@ -143,15 +143,17 @@ void GetButtonsFor(CBlob@ this, CBlob@ caller)
...
-	CBlob@ occupiedBlob = this.getAttachments().getAttachmentPointByName("MAG").getOccupied();
-	if (
-		!Vehicle_AddFlipButton(this, caller) &&
-		this.getTeamNum() == caller.getTeamNum() &&
-		isOverlapping(this, caller) &&
-		!caller.isAttached() &&
-		(occupiedBlob is null || !occupiedBlob.hasTag("player"))
-	) {
-		Vehicle_AddLoadAmmoButton(this, caller);
+	if (caller.getTeamNum() == this.getTeamNum())
```
eps0003, line 146: Change this to an early return to reduce the indentation, similar to the line above this.

Rules/CTF/Scripts/CTF_GiveSpawnItems.as
```diff
@@ -9,7 +9,7 @@ const u32 materials_wait_warmup = 40; //seconds between free mats
 //property
 const string SPAWN_ITEMS_TIMER = "CTF SpawnItems:";
 
-string base_name() { return "tent"; }
+const string[] base_name = {"tent", "warboat", "ballista"};
```
eps0003, line 12: Const variables should be written in uppercase. Don't forget to rename all occurrences of this variable.
```suggestion
const string[] BASE_NAME = {"tent", "warboat", "ballista"};
```

Rules/CTF/Scripts/CTF_GiveSpawnItems.as
```diff
@@ -229,24 +229,32 @@ void onTick(CRules@ this)
...
+				bool isShop = (spotName.find("shop") != -1);
+
 				for (uint o_step = 0; o_step < overlapping.length; ++o_step)
 				{
 					CBlob@ overlapped = overlapping[o_step];
 					if (overlapped !is null && overlapped.hasTag("player"))
 					{
-						if (!isShop || name.find(overlapped.getName()) != -1)
+						if 	((!isShop && (overlapped.getTeamNum() == spotTeam)) // is a friendly base
+							|| spotName.find(overlapped.getName()) != -1) // or is any shop
```
eps0003, line 257: To make this more readable, move the two conditions out into bool variables. Also, avoid using parentheses when they aren't required.
```suggestion
						bool isFriendlyBase = !isShop && overlapped.getTeamNum() == spotTeam;
						bool isAnyShop = spotName.find(overlapped.getName()) != -1;
						if 	(isFriendlyBase || isAnyShop)
```

eps0003 (changes requested):

Rules/CTF/Scripts/CTF_GiveSpawnItems.as
```diff
@@ -222,42 +222,45 @@ void onTick(CRules@ this)
...
-				string name = spot.getName();
-				bool isShop = (name.find("shop") != -1);
-				for (uint o_step = 0; o_step < overlapping.length; ++o_step)
+				string overlappingName = overlapping[i].getName();
+				int overlappingTeam = overlapping[i].getTeamNum();
+				
+				bool isSameTeam = playerBlob.getTeamNum() == overlapping[i].getTeamNum();
+				bool isBase = false;
+				
+				for (uint b = 0; b < BASE_NAME.length; ++b)
```
eps0003, line 245: Although you will be removing this because of the previous comment, I think it's worth pointing out the following. It is a de facto standard for programmers to follow the alphabet from `i` for nested loops. In this case, you should use `j` as the index.

Rules/CTF/Scripts/CTF_GiveSpawnItems.as
```diff
@@ -9,7 +9,7 @@ const u32 materials_wait_warmup = 40; //seconds between free mats
 //property
 const string SPAWN_ITEMS_TIMER = "CTF SpawnItems:";
 
-string base_name() { return "tent"; }
+const string[] BASE_NAME = {"tent", "warboat", "ballista"};
```
eps0003, line 12: All blobs that the player can respawn from are tagged with `"respawn"`. `BASE_NAME` can be replaced with a check to see if overlapping blobs have the tag.

eps0003 (changes requested):

Rules/CTF/Scripts/CTF_GiveSpawnItems.as
eps0003: I would avoid changing anything that isn't broken unless you are sure of the change 😁

Entities/Vehicles/Ballista/Ballista.as
```diff
@@ -232,16 +232,17 @@ void onTick(CBlob@ this)
...
 		}
 	}
-
 }
 
 void GetButtonsFor(CBlob@ this, CBlob@ caller)
 {
 	if (!canSeeButtons(this, caller)) return;
 
+	if (caller.getTeamNum() == this.getTeamNum()) return;
```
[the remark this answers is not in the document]
mugg91: It will be fixed.

Rules/CTF/Scripts/CTF_GiveSpawnItems.as
mugg91: Thank you.

Rules/CTF/Scripts/CTF_GiveSpawnItems.as
```diff
@@ -222,42 +222,45 @@ void onTick(CRules@ this)
...
-					}
+					if (BASE_NAME[b].find(overlapping[i].getName()) != -1) 
+					{ 
+						isBase = true;
+						break;
+					};
 				}
-			}
+		
+				bool isClassShop = overlappingName == "buildershop" || overlappingName == "knightshop" || overlappingName == "archershop";
```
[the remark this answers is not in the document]
mugg91: I can use `.get_string("required class")`. Knightshop will set this to "knight", Archershop to "archer", Buildershop to "builder", so I can also easily compare if the gotten string is identical to my player blob name.
This change will be committed now.
If this is not a good way to do it, please make a new comment.
Thank you.

AsuMagic:
This now conflicts due to recent changes to CTF resupply. As to whether enemy tents should give materials or not, I'm not sure. I think it is not an actual issue in itself and should be fine to keep from a gameplay perspective.

AsuMagic:
> You can't flip over an enemy Ballista or enemy Catapult. Doing so would almost always kill you from collision, so it is removed now.

Superseded by the mentioned PR above for the functionality as there are reasons to close this one.

> You can't get resupplies at an enemy Tent.

Players seem to generally be against the change.

As the two other changes are invalidated due to past gameplay changes, I will close the issue.

Issue #598, opened 2019-05-26
!gold abuse in sandbox


mugg91:
Gold does not despawn, it seems.

It is possible to bring a sandbox server to its knees by !gold spam to spawn 1000+ gold heaps, which causes framerate to become >1 FPS.

Possible solutions:
* Cap the amount of commands you can send
* Limit the amount of gold (and other objects that don't despawn) on the field at the same time (For example, only allow each team to have 20 boats, boulders, etc. at maximum).
*

Harrison-Miller:
That or just make things that don't despawn, despawn in sandbox.

Harrison-Miller:
ok actually apparently there's a fix for this already, closing.

mugg91:
This is not fixed in the current build 3900. Tested in Official Sandbox EU server.

mugg91:
Gold spam was fixed a while ago.

PR #1276, opened 2022-05-30
Gold will decay over time (only in modes other than Challenge, CTF, TTH)

mugg91:
## Status

## Description

This fixes issue #598.

Gold which is spawned into Sandbox would never despawn, since Gold doesn't have a 'decay step' set like the other materials do. Therefore, spamming !gold in Sandbox servers would be a griefing method, since nothing could be done to despawn the Gold.

This change adds a 'decay step' to Gold and DecayQuantity.as to its script section, so it will eventually decay over time.

Similar to Challenge.as, CTF.as and WAR.as will remove DecayQuantity for any gold blobs created so it won't decay in those game modes.

Although ChatCommands.as is getting refactored, I did change the existing ChatCommands.as just to be sure.  
`!gold` will spawn 50 gold instead of 200 and `!allmats` will spawn 500 wood, 500 stone, 50 gold instead of 500 wood, 500 stone, 100 gold. This was done to reduce gold spam and to make decaying gold look less weird (it reduces, then merges, then reduces, etc.)

## Steps to Test or Reproduce

1. Go to Sandbox.
2. !gold !gold !gold
3. Notice the Gold will disappear.

### DIFF
--- a/Rules/CTF/Scripts/CTF.as
+++ b/Rules/CTF/Scripts/CTF.as
@@ -716,6 +716,6 @@ void onBlobCreated(CRules@ this, CBlob@ blob)
 {
 	if (blob.getName() == "mat_gold")
 	{
-		blob.RemoveScript("DecayQuantity");
+		blob.RemoveScript("DecayQuantity.as");
 	}
 }
--- a/Rules/CTF/Scripts/CTF.as
+++ b/Rules/CTF/Scripts/CTF.as
@@ -711,3 +711,11 @@ void onBlobDie(CRules@ this, CBlob@ blob)
 		}
 	}
 }
+
+void onBlobCreated(CRules@ this, CBlob@ blob)
+{
+	if (blob.getName() == "mat_gold")
+	{
+		blob.RemoveScript("DecayQuantity");
+	}
+}
--- a/Rules/CommonScripts/ChatCommands.as
+++ b/Rules/CommonScripts/ChatCommands.as
@@ -202,7 +202,7 @@ bool onServerProcessChat(CRules@ this, const string& in text_in, string& out tex
 				server_CreateBlob('chicken', -1, pos);
 			}
 		}
-		else if (text_in == "!allmats") // 500 wood, 500 stone, 100 gold
+		else if (text_in == "!allmats") // 500 wood, 500 stone, 50 gold
 		{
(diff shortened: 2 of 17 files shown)


Rules/CTF/Scripts/CTF.as
```diff
@@ -711,3 +711,11 @@ void onBlobDie(CRules@ this, CBlob@ blob)
 		}
 	}
 }
+
+void onBlobCreated(CRules@ this, CBlob@ blob)
+{
+	if (blob.getName() == "mat_gold")
+	{
+		blob.RemoveScript("DecayQuantity");
```
eps0003, line 719: It should include the file extension to avoid the chance of the file matcher finding the wrong file.
```suggestion
		blob.RemoveScript("DecayQuantity.as");
```

Rules/CommonScripts/ChatCommands.as
```diff
@@ -212,7 +212,6 @@ bool onServerProcessChat(CRules@ this, const string& in text_in, string& out tex
 			stone.server_SetQuantity(500);
 			//gold
 			CBlob@ gold = server_CreateBlob('mat_gold', -1, pos);
```
eps0003, line 214: The blob doesn't need to be stored to a variable since you have removed the line below it.
```suggestion
			server_CreateBlob('mat_gold', -1, pos);
```

Rules/WAR/Scripts/WAR.as
```diff
@@ -1050,6 +1050,10 @@ void onBlobCreated(CRules@ this, CBlob@ blob)
 	{
 		MakeWarTradeMenu(blob);
 	}
+	else if (blob.getName() == "mat_gold")
+	{
+		blob.RemoveScript("DecayQuantity");
```
eps0003, line 1055: Same as previous comment.
```suggestion
		blob.RemoveScript("DecayQuantity.as");
```

eps0003 (dismissed):

mugg91:
I applied all of your suggested changes.

Rules/CommonScripts/ChatCommands.as
eps0003: You forgot this change.

AsuMagic (approved):

PR #1279, opened 2022-06-06
[fixed] floating Arrows

mugg91:
## Status

## Description

[removed] TeamStructureNear.as from Arrow.as as it seems unused.
[removed] A check in Arrows.as (see lines 168-177) because it seems useless.
[changed] isOpen() is the same between SwingDoor, TrapBlock and Bridge, so moved it to DoorCommon.as.
[fixed] Arrows floating in the air.

This is a simple floating arrows fix.

When an arrow hits a structure blob, then that blob is passed to ArrowHitMap(), 
otherwise if the map is hit then null is passed.

In the former case, a reference to the hit blob is saved to the arrow,
in the latter case, the endpos of the map raycast check is saved to the arrow.

onTick() which is called every tick deals with these two scenarios and checks if the structure blob is still there,
or if the structure blob or map tile is still solid. Otherwise the arrow is set to die.

Please note, this was tested only locally. This has yet to be tested on a dedicated server.

## Steps to Test or Reproduce

1. Shoot an arrow at a door. Walk through it. Notice it dies.
2. Shoot an arrow at a door/platform/bridge/magazine/etc. that is built on a wall. Make the structure collapse. Notice the arrow dies.
3. Shoot an arrow at a bridge, switch team and fall through it. Notice the arrow dies.
4. /loadmap Princess06, shoot an arrow at the trap blocks, fall through. Notice the arrow dies.
5. Build obstructor blocks and activate them via switch. Shoot at them. Then disable the blocks via switch. Notice the arrow dies.

## How to make this Better?

I said this is a simple fix. I think these would be nice to have. Not sure if I will work on it:
- When the door with arrow on it is opened, make the arrow shake and fall only after 1-2 sec. So enemy has to leave himself vulnerable.
- Instead of simply dying, it would be nice if the arrows fell down (kind of like a second shot).
- I made a PR [here](https://github.com/transhumandesign/kag-base/pull/1249) which makes the Dummy strawman hittable by arrow. The arrow will be stuck on it. So just in case both PRs get accepted, the behavior of Arrow sticking to a Dummy should be looked at.
- Instead of dying when hitting an enemy crate or catapult, make it stick, move and rotate along with those blobs.


Entities/Items/Projectiles/Arrow.as
```diff
@@ -165,22 +163,40 @@ void onTick(CBlob@ this)
...
 			{
-				CBlob@ _blob = this.getTouchingByIndex(step);
-				_blob.getShape().checkCollisionsAgain = true;
+				this.server_Die();
+			}
+			else if (gottenBlob.hasTag("fallen")										// structure blob is falling
+					|| (gottenBlob.getName().find("door") != -1 && isOpen(gottenBlob))	// opening door
+					|| (gottenBlob.getName() == "bridge" && isOpen(gottenBlob))			// opening bridge
+					|| (gottenBlob.getName() == "trap_block" && isOpen(gottenBlob)))	// opening trap block
+			{
```
Gingerbeard5773, line 181: Checking if gottenBlob is open three times is redundant, it can be done with just one check. You will have to add another set of parenthesis that group your getName checks. Then just add the isOpen check at the end. Should be something like ``if (b.hasTag("tag") || ((namecheck1 || namecheck2 || namecheck3) && isOpen(b)))``

Entities/Items/Projectiles/Arrow.as
```diff
@@ -165,22 +163,40 @@ void onTick(CBlob@ this)
...
+			
+			CMap@ map = getMap();
+			Vec2f hitpos = this.get_Vec2f("hitWorldPoint");
+			
+			Tile hitTile = map.getTile(hitpos);
+			
+			if (!map.isTileSolid(hitTile))
+				this.server_Die();		
+		}
+
```
Gingerbeard5773, line 197: All of the code you added here (lines 166 to 196) should only be calculated by the server since all the pathways from the "hitBlobExists" check only do server_die() and nothing else. Add a ``isServer()`` check.

Entities/Items/Projectiles/Arrow.as
```diff
@@ -582,6 +598,24 @@ void ArrowHitMap(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, u8 c
...
+	}
+	else // hitBlob (door, platform, etc.)
+	{
+		this.set_bool("hitBlobExists", true);
+		this.set_u32("hitBlob", hitBlob.getNetworkID());
+		
+		this.Sync("hitBlob", true);
+	}
+	
+	this.Sync("hitBlobExists", true);
```
Gingerbeard5773, line 618: Since we're only setting vars to kill this blob (server-side) eventually, put a ``isServer()`` check for lines 601 to 618. And since we are only doing server-side anyways, you don't need the Sync() instances, so remove all these syncs you added since they aren't used for anything good.

Gingerbeard5773 (changes requested):

Entities/Items/Projectiles/Arrow.as
mugg91: Thank you. All of your suggestions have been addressed in [8f1846a](https://github.com/transhumandesign/kag-base/pull/1279/commits/8f1846a6a8c8d30cfbc241c86f9026684e938d0b).

mugg91:
> [removed] A check in Arrows.as (see lines 168-177) because it seems useless.

I noticed in online Arrows can get stuck to you and will keep stuck while you are walking around.
Maybe the check I removed is doing this, so it is not useless.

AsuMagic:
> `[removed] TeamStructureNear.as from Arrow.as as it seems unused.`

This was for grief protection code which ended up being reverted IIRC, so yeah.

AsuMagic:
Fixed conflicts.

AsuMagic (approved):

AsuMagic:
BTW, you have been hit by operator precedence:

`bool isOpened = (isOpen(gottenBlob) && n.find("door") != -1 || n == "bridge" || n == "trap_block");`

binds like

`bool isOpened = ((isOpen(gottenBlob) && n.find("door") != -1) || n == "bridge" || n == "trap_block");`

which I fixed to

`bool isOpened = (isOpen(gottenBlob) && (n.find("door") != -1 || n == "bridge" || n == "trap_block"));`

because arrows could immediately break when touching bridges or trap blocks.

Issue #1284, opened 2022-06-10
getBlobsInBox() will ignore blob's radius and verticesXY


mugg91:
## Description

`getBlobsAtPosition()` and `getBlobAtPosition()` will consider a blob's radius and verticesXY.

However, `getBlobsInBox()` will only consider a blob's sprite_frame_width and sprite_frame_height. 
It will not consider a blob's radius or verticesXY. I'm under the impression that it should.

## Reproduction

In DrawHelp.as, remove lines 512-520 and add:

```
Vec2f mousepos = getControls().getMouseWorldPos();
				
CBlob@[] overlapping;
// getMap().getBlobsInBox(mousepos, mousepos, @overlapping);
// getMap().getBlobsAtPosition(mousepos, @overlapping);
		
for (uint i = 0; i < overlapping.length; i++)
{
	CBlob@ mouseBlob = overlapping[i];
							
	if (mouseBlob !is null)
	{
		string invName = getTranslatedString(mouseBlob.getInventoryName());
		Vec2f dimensions;
		GUI::SetFont("menu");
		GUI::GetTextDimensions(invName, dimensions);
		GUI::DrawText(invName, getDriver().getScreenPosFromWorldPos(mouseBlob.getPosition() - Vec2f(0, -mouseBlob.getHeight() / 2)) - Vec2f(dimensions.x / 2, -8.0f), color_white);
	}

}
```
----
Uncomment one of the two commented lines.
Go to Sandbox.
If `getMap().getBlobsAtPosition(mousepos, @overlapping);` is used, then a blob's name is drawn on screen if your cursor is inside their radius or verticesXY:

https://i.imgur.com/nB7tYSy.png - Not inside radius → no text
https://i.imgur.com/WCil1JJ.png - Inside radius → text

https://i.imgur.com/XLVWfyg.png - Not inside VerticesXY → no text
https://i.imgur.com/3gwh1V3.png - Inside VerticesXY → text

----
If `getMap().getBlobsInBox(mousepos, mousepos, @overlapping);` is used, then a blob's name is drawn on screen if your cursor is inside their sprite_frame_width and sprite_frame_height. It will not take radius or verticesXY into account:

https://i.imgur.com/lzJRdQW.png - Not inside radius, but within sprite frame → text
https://i.imgur.com/2uZv06Y.png - Not inside VerticesXY, but within sprite frame → text

AsuMagic:
[...]

PR #1285, opened 2022-06-10
Placing a block will destroy scenery in the whole block instead of only in the top left of the block

mugg91:
## Status

## Description

```
A B C
D E F
G H I
```

Assume there is a Bush placed in the tile at E.
I noticed the bush only gets destroyed if a block is placed in E, F, H, I.
So it gets destroyed when placing a block to its right, but it stays alive when placing a block to its left.
I looked into it.

If a building is built, `DestroyScenary(tl, br);` in BuilderCommon.as is executed. 
tl is actually the top-left of the building.
br is actually the bottom-right of the building.
Everything is fine here.

If a block is placed, `DestroyScenary(cursorPos, cursorPos)` in BlockPlacement.as is executed.
cursorPos is equivalent to the tl of the block. 
So scenery will only be destroyed in that section.
This is not consistent with the behavior in BuilderCommon.as.

This PR fixes that so that the whole block is considered, to make it consistent with the behavior in BuilderCommon.as.

A [bug](https://github.com/transhumandesign/kag-base/issues/1284) in `getBlobsInBox()` (which is used in `DestroyScenary()`) makes it so a Bush will still be destroyed if the tl of a placed block is within the sprite_frame of the bush, regardless of its radius.

After this PR, assuming `getBlobsInBox()` stays the way it is, a bush will be destroyed when placing a block in any of the 9 tiles A, B, C, D, E, F, G, H, I, making it consistent.
There is no such thing as "it is destroyed when placing it to its right, but not destroyed when placing it to its left".

## Steps to Test or Reproduce

Go to Sandbox and place a wood block or stone block inside it or in the 8 tiles surrounding it.

### DIFF
--- a/Entities/Common/Building/BlockPlacement.as
+++ b/Entities/Common/Building/BlockPlacement.as
@@ -31,9 +31,10 @@ void PlaceBlock(CBlob@ this, u8 index, Vec2f cursorPos)
 
 	if (validTile && hasReqs && passesChecks)
 	{
-		DestroyScenary(cursorPos, Vec2f(cursorPos.x+8, cursorPos.y+8));
+		CMap@ map = getMap();
+		DestroyScenary(cursorPos, Vec2f(cursorPos.x+map.tilesize, cursorPos.y+map.tilesize));
 		server_TakeRequirements(inv, bc.reqs);
-		getMap().server_SetTile(cursorPos, bc.tile);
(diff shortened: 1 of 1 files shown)


soren-null (approved):
Seems to work as described, tested on dedicatedserver

Gingerbeard5773 (approved):
Works as intended

AsuMagic (approved):

PR #1286, opened 2022-06-10
[Added] Shooting at Foods will stick them to the Arrow's tip

mugg91:
## Status

As far as my judgement goes, this is good to go and ready, but please do test and verify this yourselves.
Please consider the "Things to Add" section below.

## Description

Added a "TIP" attachment point to Arrow that attachable blobs can get attached to.
Removed the previous "specialArrowHit" logic that added velocity if an attachable blob was shot at.
I hope this makes the "shooting foods to teammates" mechanic more reliable, fun and increase depth.

Attachable blobs are:
- food
- grain
- bread
- steak
- cooked_steak
- cooked_fish
- heart (this was commented out before. I included it here because why not.)
- sponge (small enough and makes sense)

A check is added in onTick() that is performed for flying arrows.
If the flying arrow is normal-type or fire-type, velocity is sufficient, and an attachable blob is near,
then that blob will get attached to the arrow.

After the arrow collides, the tipped blob will get detached and set to `SetStatic(true)`, so it can be picked up while staying sticked to the wall. The tipped blob will get set to `collidable = false` to avoid unwanted collision with other blobs.

When the arrow dies, the tipped blob will get set to `SetStatic(false)` and the collidable state will be reverted to what it was before, so it can fall down and have collisions as usual. When the player picks up the tipped blob, the collidable state will also be reverted.

As far as I tested, medium-charged arrow will have enough velocity to pass the check and attach an attachable blob to itself. Having foods bounce on a trampoline and spamming arrows on that trampoline, the arrows will not gain enough velocity to pass the check.

## Steps to Test or Reproduce

Shoot at burgers, hearts and sponges.
Notice they will get attached to your arrows.

## Things to Add

Foods that are attached to a flying arrow should heal teammates.
The coin reward for throwing a healing item to a teammate should be applied here.
I can add a "detach when overlapping with teamplayer" logic, but maybe there is a better way, to avoid accidentally healing the enemy that is overlapping with the teamplayer.

### DIFF
--- a/Entities/Items/Projectiles/Arrow.as
+++ b/Entities/Items/Projectiles/Arrow.as
@@ -177,8 +177,7 @@ void onTick(CBlob@ this)
 				{	
 					AttachmentPoint@ tip = this.getAttachments().getAttachmentPointByName("TIP");
 					
-					if (tip !is null 								// attachment point exists
-						&& tip.getOccupied() is null)				// no blob in the attachment point yet
+					if (tip !is null && tip.getOccupied() is null)				// no blob in the attachment point yet
 					{
 						this.server_AttachTo(overlapping[i], "TIP");
 						overlapping[i].getSprite().SetRelativeZ(500.0f);			
--- a/Entities/Items/Projectiles/Arrow.as
+++ b/Entities/Items/Projectiles/Arrow.as
@@ -166,7 +166,7 @@ void onTick(CBlob@ this)
 		// attach certain blobs with the tip of the arrow	
 		if (!this.hasAttached() 												// arrow doesn't have anything attached yet
 			&& (arrowType == ArrowType::normal || arrowType == ArrowType::fire)	// only normal and fire type
-			&& this.getShape().vellen > 10)										// only if enough velocity
+			&& this.getShape().vellen > 12)										// only if enough velocity
 		{
 			CBlob@[] overlapping;
 			getMap().getBlobsInRadius(this.getPosition(), 7.0f, @overlapping);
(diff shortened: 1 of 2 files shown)


mugg91:
I added some fixes.

Z Position of blob in arrow's tip is saved and reverted later.

Added a check to determine if the tipped blob is a food item. If yes,
-  `collideWhenAttached` is set to true so foods will heal teammates.
- "healer" is added to the blob, so there is a coin reward.

When the arrow has collided (also when picked up, for safety):
- `collideWhenAttached` is set to what it was before.

I can't think of more things to fix or change, so this is good for review now.

AsuMagic:
I quickly tested this and observed a few issues:

1. If shooting a burger at a full health enemy, they will get damaged, but the burger will be left floating.
image
2. Sometimes when a burger arrow hits the ground, said burger is stuck below the ground, which looks pretty glitchy visually, and makes it difficult to pick them up.
3. Naturally spawned fish, after being picked up, won't collide with arrows.

General thoughts:
- This is interesting and I wanted to try doing a mechanic like this a while back.
- This makes it so you cannot pick up an in-flight item. Maybe it would be a more interesting mechanic if you could?
- Burger arrows shot at enemies damage them... which looks a bit weird. (It doesn't _always_ happen, e.g. if the burger hitbox touches the enemy but not the arrow itself).
- If it is easy to do, items could follow the arrow rotation for a nicer visual look.
- The attachment stuff should be sanity checked a fair bit for edge cases.
- **More community feedback from a gameplay POV would be good to have.**

mugg91:
Will probably look at @AsuMagic 's points in like 2 weeks. Won't get to it before then.

Issue #1245, opened 2022-04-30
A list of lantern/storage bugs


mugg91:
**Description**

~When storing a turned-off lantern in Storage, it will show up as turned-on in the Storage.~ (Fixed by #1287)

~When the overlaying lantern inside the storage is extinguished by being inside water, and then withdrawing the lantern, it will still be turned on.~ (Fixed by #1287)

~When the overlaying lantern inside the storage is extinguished by being inside water, and then withdrawing all lanterns, then putting turned-on lanterns inside the storage, the overlaying lantern will stay turned off.~ (Fixed by #1287)

~Lantern sprite in your inventory does not indicate if it's turned off or on.~ (Fixed by [#1264](https://github.com/transhumandesign/kag-base/pull/1264))

If storing two lanterns in a storage and then withdrawing both, sometimes it will still show a turned on lantern overlaying on the storage even when there is none inside. This only happens sometimes both online and offline.

~Action 3 will turn on lantern when under water.~ (Fixed by [#1248](https://github.com/transhumandesign/kag-base/pull/1248))

Just mentioning here the fact that Lantern will get extinguished underwater but not by water bucket, water bomb or water arrow like a fireplace does.

**Expected Behavior**

The overlaying lantern on storage should be synced better with what's inside the storage:

When there is no lantern or only a turned off lantern inside, the overlaying lantern on storage should be turned off.

When at least one turned on lantern is inside, it should show an overlaying turned on lantern.

When the storage is in water, it should turn off all lanterns inside it as well as the overlaying lantern. Maybe also do this for other blobs that have inventories e.g. crate, dinghy, players. (This one is debatable)

For consistency reasons, it may be better if lantern gets extinguished by all water sources. (This one is not really a problem though)

PR #1287, opened 2022-06-13
Lantern overlay on Storage will be synced with Lanterns inside Storage

mugg91:
## Status

## Description

This fixes issue #1245.

When a lantern is added to or removed from a Storage and `updateLayers()` is executed, it will check all lanterns in the Storage. If at least one lantern is turned on, the overlay lantern will be turned on, too.
If no lantern is turned on, the overlay lantern will be turned off.

When the overlay lantern is turned off, it can mean one of two things:

Either 1) only turned-off lanterns have been added to the Storage.

Or 2) the overlay lantern has been flooded by water. 

A code in `onTick()` will go through all lanterns inside the Storage and each one that is turned on will be turned off. This is not a problem, since logically all lanterns inside should be turned-off anyway - see 1).
TickFrequency is kept at 60 like before so it is not expensive. (It doesn't lag even if tickFrequency is 1 and running the same code inside a `for (uint a = 0; a < 200000; a++)` loop. So it may be worth considering increasing the tickFrequency.)

## Steps to Test or Reproduce

Go to Sandbox.
Spawn a Lantern. Turn it off and put it in a Storage.
Notice the overlay will not turn on.

Put a turned-on Lantern inside.
Notice the overlay will be turned on and stay that way until the turned-on Lantern is withdrawn.

Put some turned-on Lanterns inside.
Spawn water so the overlay turns off.
When withdrawing all lanterns, notice they will all be turned off.

### DIFF
--- a/Entities/Industry/CTFShops/Storage/Storage.as
+++ b/Entities/Industry/CTFShops/Storage/Storage.as
@@ -1,6 +1,7 @@
 ﻿// Storage.as
 
-#include "GenericButtonCommon.as"
+#include "GenericButtonCommon.as";
+#include "LanternCommon.as";
 
 void onInit(CSprite@ this)
 {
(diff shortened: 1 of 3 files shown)


eps0003:
Resolves #1245

AsuMagic:
Contained by #1581

PR #1290, opened 2022-06-17
[fixed] Boulder not hitting the buyer's team

mugg91:
## Status

## Description

A Boulder bought from Trader/Buildershop or spawned in Sandbox will always be of team 255. 
Therefore it will hit anyone who isn't in your team.
However, for Boulders bought at Trader, line 308 in TradingMenu.as will set the Boulder team to your team:
```
blob.server_setTeamNum(caller.getTeamNum());
if (!item.buyIntoInventory || !caller.server_PutInInventory(blob))
{
	caller.server_Pickup(blob);
}
```

If the enemy picks it up, it will still be considered to be of the buyer's team. So the buyer's team cannot be hit by the boulder when the enemy picks it up. This causes strange situations in TDM: "I stole your Boulder and threw it back at you. Wtf it isn't dealing damage?"

To fix this, line 308 in TradingMenu.as can be removed. But it is important to set the item's team to the buyer for things like Keg, Sponge and others. So I decided to add SetTeamToCarrier.as to Boulder.cfg. Now, the one who gets attached to the Boulder will also set the Boulder's team to theirs. The damage owner will still be set to the throwing player like before.

## Steps to Test or Reproduce

Go to Save the Princess.
!tradingpost
!builder
!coins 999
Buy a Boulder, place it on the ground.
!team 1
Pick up the Boulder. Throw it at the builder who is in Team 0.
Notice the builder will not take damage (because the Boulder is also Team 0).
**After this change, the Boulder will assume the picker's team - in this case, team 1 - and deal damage to the builder. There is no "why isn't the boulder I stole from you not hitting you?" situation.**

Go to Sandbox.
!builder
!boulder
!team 1
Pick up the boulder and throw it at the builder.
The builder will take damage, due to the Boulder being in Team 255. **After this change, the Boulder will assume the picker's team - in this case, team 1 - and deal damage to the builder. It doesn't make a difference here.**

### DIFF
--- a/Entities/Items/Boulder/Boulder.cfg
+++ b/Entities/Items/Boulder/Boulder.cfg
@@ -108,7 +108,8 @@ $name                               = boulder
 										CheapFakeRolling.as;
 										DecayIfSpammed;
 										GenericHit.as;
-										SetDamageToCarrier.as
+										SetDamageToCarrier.as;
+										SetTeamToCarrier.as
 f32 health                          = 1.5
 # looks & behaviour inside inventory
 $inventory_name                     = Boulder


Gingerbeard5773 (approved):
Works as intended

AsuMagic (approved):

PR #1292, opened 2022-06-18
[changed] Team Bridge will not pass light, allowing Spikes/Saw to be hidden underneath

mugg91:
## Status

## Description

I liked that you could hide Saws and Spikes underneath Trap Blocks.
The Team Bridges that have been added in Build 3479 to replace Trap Blocks don't have this feature.
Because it seems this was not a deliberate decision, I decided to make this PR which makes Team Bridges to not pass light to add this feature back.

I used this code:
```
sprite.getConsts().accurateLighting = true;
this.SetLight(true);
this.SetLightRadius(7.0f);
```
This will cause each Bridge tile to emit just enough light to make it look normal without showing what's underneath.
When enclosed in a room, the Bridge tile will be almost dark but still slightly visible. Maybe it's a price worth paying to get "hiding spikes/Saw" back.

Before - https://i.imgur.com/oz6Npkb.png
After - https://i.imgur.com/i0jVafI.png


AsuMagic:
Not sure about the gameplay implications of that. Some discussion/feedback on this would be nice.

At the moment I am wary towards toggling light passing dynamically. At the moment, vanilla code is very bad at lighting updates, so this might result in much more observed visual glitchiness because of lighting. Staging fixes this, though, making light updates instant and rather fast.

mugg91:
I think we should not use this. 
I would be more in favor of bringing back Trap Blocks.
Armored Warfare mod has better Trap Blocks (allows teammates to pass from bottom to top) which I would recommend.

Issue #1295, opened 2022-06-21
Add features based on Daytime/Nighttime


mugg91:
## Describe the feature

I think it would be nice to add some effects and logics based on the map's daytime.
`getMap().getDayTime()` will tell the current daytime, as a float counting from 0 to 1. 
Stuff can be done based on this value.

Roughly speaking, night takes place during 0.8 to 1.0, the rest is sunset/dawn/day.
Having said that, I think it would be good to make night last from 0.75 to 1.0 to give night more time without being overbearing.

1) Make a new ambient music play during sunset/nighttime (e.g. [chirping crickets](https://pixabay.com/sound-effects/ambience-night-field-cricket-01-7015/))
2) Make sure that `ambient_forest.ogg` - which contains birds chirping sounds - isn't playing during nighttime.
3) I think the stars background layer can be improved, but it seems to be done engine-side. 
Stars shouldn't be appearing during dawn time. 
Don't make them to move vertically so much, as it looks odd. 
Maybe make the visuals to be more than just "white pixels".
Add a rare shooting star animation?

4) Make Bison and chicken to sleep at night - They could be startled and behave as normal, but go to sleep if idle for a while. Just adds more life to the game.

5) Butterfly:
This concept was briefly discussed in Discord.
The idea is to make Butterfly particles or blobs appear near Flowers, serving as nice visuals to make the game feel more alive.
It's a different feature entirely, I'm just saying, if Butterflies were to be added, make them appear only at day.
If there are going to be butterflies, might as well make Moths that go near light sources, in caves and otherwise only at night.

6) Flowers closed sprite at night

Please let me know if you have more ideas.

PR #1296, opened 2022-06-22
[added] Night ambient music

mugg91:
## Status

## Description

Fixes 2) of #1295.

This adds a new night world ambient music track to WARMusic.as and ChallengeMusic.as.
It is crickets chirping which is common to occur at late afternoon to midnight.
This music plays when `getMap().getDayTime()` is between 0.75 and 1.0 - before sunset starts until before dawn starts.

The music used is the first 45 seconds of this, amplied by +6.1 and saved as .ogg at quality level 6. It is free-to-use: 
https://pixabay.com/sound-effects/cricket-sound-14867/
If you think there is a better soundtrack, let me know.

## Steps to Test or Reproduce

Go to Sandbox or TTH tutorial.
Wait until sunset and night (or type `getMap().SetDayTime(0.7f);` in console to jump right before a sunset).
Notice the night ambient music.


AsuMagic (changes requested):
The background music really doesn't fit, you can easily hear city sounds and cars running. :P

mugg91:
> The background music really doesn't fit, you can easily hear city sounds and cars running. :P

Please note, only the first 45 seconds are used. The part of the car driving is not included.
Any other sound effects are basicly inaudible in actual gameplay (note, there is the game music running on top of this) and could be assumed to be village sounds or wind.

Entities/Meta/ChallengeMusic.as
```diff
@@ -124,7 +134,10 @@ void GameMusicLogic(CBlob@ this, CMixer@ mixer)
 			if (timer % 24 != 0)
 				return;
 
-			if (mixer.isPlaying(world_ambient) || mixer.isPlaying(world_ambient_underground) || mixer.isPlaying(world_ambient_mountain))
+			if (mixer.isPlaying(world_ambient) 
+				|| mixer.isPlaying(world_ambient_underground) 
+				|| mixer.isPlaying(world_ambient_mountain
```
AsuMagic, line 139: Missing `)`, this causes a syntax error.

Entities/Meta/ChallengeMusic.as
```diff
@@ -95,15 +97,23 @@ void GameMusicLogic(CBlob@ this, CMixer@ mixer)
 		if (timer % 48 != 0)
 			return;
 
+		bool isNight = map.getDayTime() > 0.75f;
```
AsuMagic, line 100: This seems to be a generally poor value. `>0.85 && <0.1` would probably be more fitting with the current sky gradient.

AsuMagic (changes requested):

Entities/Meta/ChallengeMusic.as
AsuMagic:
```suggestion
		bool isNight = map.getDayTime() > 0.85f && map.getDayTime() < 0.1f;
```
I have now realized that fix suggestions are a thing.

Entities/Meta/ChallengeMusic.as
AsuMagic:
```suggestion
				|| mixer.isPlaying(world_ambient_mountain)
```

AsuMagic (dismissed):

AsuMagic (approved):

PR #1298, opened 2022-06-25
[changed] Blobs on fire are emitting light

mugg91:
## Status

## Description

I noticed the fire animation sprites aren't emitting light.
So I looked into it.
It is possible to make a blob which is on fire to emit light, but if the fire animation is on a worldspace, it cannot be done.

Before - https://imgur.com/5kvgQJT.png 
After - https://i.imgur.com/tCHG0IU.png
(On the left, the blob that emits light - on the right, the worldspace which cannot be set to emit light)

## How to make this Better

It would be better if the sprite itself emits light, not the blob. Because it looks odd when the whole tree is bright in the dark just from a small flame at its bottom. 

This would also fix that a worldspace/tile cannot emit light. If the sprite can emit light, there is no need for a worldspace/tile to emit it.

### DIFF
--- a/Entities/Common/FireScripts/FireAnim.as
+++ b/Entities/Common/FireScripts/FireAnim.as
@@ -41,6 +41,7 @@ void onTick(CSprite@ this)
 			this.getCurrentScript().tickFrequency = 12;
 
 			fire.SetVisible(true);
+			if (blob !is null)	{ blob.SetLight(true); }
 
 			//TODO: draw the fire layer with varying sizes based on var - may need sync spam :/
 			//fire.SetAnimation( "bigfire" );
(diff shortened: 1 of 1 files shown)


AsuMagic:
Might be a good visual change but aside of the issue that you brought up: if a light is able to emit light and can be set on fire, this would probably lead to trouble, so I'm not a fan of this approach in the current state.

FWIW: when:tm: staging gets merged, there should be a lot more potential for more advanced light source things like these, including fast light particles which could be the better approach for this.

AsuMagic:
Note: Staging now supports light layer particles, but it is not quite documented yet.

Vam-Jam:
Closing in favour of https://github.com/transhumandesign/kag-base/pull/2319

PR #1301, opened 2022-06-28
[fixed] When invincible, you won't be set on fire / knocked back by bomb / lose air in water / take drown damage

### LINKED ISSUE #1221: Players burn when invulnerable
**Description**

Respawned players will take burn damage and start panic running immediately, even while still invulnerable.

**Reproduction**

Happens sometimes in CTF. You can produce this situation in Sandbox:

1. Join Sandbox server.
2. build wood background at the spawn point.
3. Shoot a fire arrow at it. Wait until it burns.
4. Open menu and suicide. 
5. If you respawned at the burning structure, you will take burn damage and start panic running immediately.

**Expected Behavior**

You shouldn't take burn damage the first 3 seconds after respawning.

mugg91:
## Status

## Description

Fixes #1221.

[changed] Player won't get hurt by their own damage sources when invincible
[fixed] Player won't get set on fire when invincible
[fixed] Player won't get knocked back when invincible
[fixed] Player doesn't make a hurt sound and have blood animation 
when bomb explodes in hand while invincible
[fixed] Player won't lose air or take drown damage when invincible

I made it so your own damage source won't hurt you because I think there is no merit to such a game mechanic.
Whatever you did before dying shouldn't be causing you to insta-get-damaged to die again.

You will not be instantly set on fire when you spawn in a tent that has burning grass on it.
You will not be knocked back by a bomb/keg that explodes right after you spawn in.
You will start to lose air in water only after losing invincibility.

Somehow, getting hit by a bomb doesn't initiate any "knocking state" so the invincibility has to be accounted for in the Explosion.as logic instead.
When the blob has invincible status and is hit by a bomb, the bombforce is set to 0.
This somehow also makes it so there is no hurt sound nor blood animation. (Not sure why, but yay.)

As pointed out by [bunniewormy](https://github.com/bunniewormy), removing the `victim !is attacker` check causes a problem where Suiciding while invincible causes the player to hop in the air instead. By adding an onHit() check in RunnerDefault.as, invincibility is removed when suiciding so the Suicide takes place as usual.

Normally, you wouldn't ever take drown damage while invincible, since invincibility wears off before that happens. But there might be a chance that invincibility periods will be added for other things (like [fixing Saw bug](https://github.com/transhumandesign/kag-base/issues/1252)), so it is a good thing a potential future bug is sorted out here and now.

## Steps to Test or Reproduce

Go to Sandbox.
Shoot a fire arrow at the location where you spawn.
When grass is about to burn, Suicide and then walk in the grass.
Notice you will be set on fire while invincible. **After this change, this won't happen.**

---

!bomb !bomb !bomb
Suicide.
Notice you will be knocked back by the explosion while invincible. **After this change, this won't happen.**

---

!spawnwater
Suicide.
Notice you will lose air in water while invincible. **After this change, this won't happen.**


mugg91:
I added two more commits that will make it so animals cannot bite or tackle invincible players.

AsuMagic (approved):

AsuMagic (approved):

PR #1303, opened 2022-07-02
[fixed?] Player becoming stuck in a glitched state when using Bed at the same time Quarters are collapsing (in online servers only)

mugg91:
## Status

## Description

When using bed in Quarters at the same time it is collapsing, there is a chance that the player will get stuck in a glitched state.
This happens occasionally and only in online servers.
Testing this when connected with one's own dedicated server and [clumsy](https://jagt.github.io/clumsy/) set to Lag = 50~75 ms, it is very common to happen.

This PR aims to fix this.
When the Quarters are destroyed, it will save this information in onDie() via `this.set_bool("quarters is destroyed", true)`.
In onCommnand(), where the attaching is done, it will ask this information before doing the attaching.
If `this.get_bool("quarters is destroyed")` is true, it will not do the attaching.
I tested again with my own server and while running clumsy and it hasn't occured anymore.

A different way of fixing this could have been to check if tiles behind the quarters are collapsing and if they are, don't attach the player to the bed.
But since there are multiple tiles and it would have to check for each one, it would be a messy code.

## Steps to Test or Reproduce

Run your dedicated server with the modified Quarters.as.
Run clumsy and set it to Lag 50ms ~ 75ms.
Connect to your server.
Make sure you are not full health.
Spawn a quarters somehwere via !quarters.
Go somewhere else and jump in the air, while at the peak of your jump open chat, up key, enter to place a quarters in the middle of the air.
Jump up again and quickly use the bed button.

**If you successfully used the bed button, either of these would happen:**
- The player gets attached, the building is destroyed and the player detaches.
- The player doesn't get attached.

**Before this PR:**
- The player gets attached, the building is destroyed and the player detaches.
- The player gets attached while the quarters don't exist anymore - bug scenario happens.

### DIFF
--- a/Entities/Industry/CTFShops/Quarters/Quarters.as
+++ b/Entities/Industry/CTFShops/Quarters/Quarters.as
@@ -116,7 +116,7 @@ void onTick(CBlob@ this)
 {
 	// TODO: Add stage based sleeping, rest(2 * 30) | sleep(heal_amount * (patient.getHealth() - patient.getInitialHealth())) | awaken(1 * 30)
 	// TODO: Add SetScreenFlash(rest_time, 19, 13, 29) to represent the player gradually falling asleep
-	bool isServer = getNet().isServer();
+
 	AttachmentPoint@ bed = this.getAttachments().getAttachmentPointByName("BED");
 	if (bed !is null)
 	{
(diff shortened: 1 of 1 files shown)


mugg91:
It is also possible to check if the Quarters's health is 0 or less.
However, is is possible that the Quarters have health 0 or less but are not destroyed yet.

Although the bug scenario happens only if the quarters are destroyed, by checking for health we wouldn't have to add an "onDie()" instance and save extra information via `set_bool()`.

Which is better? 
Checking if the building is destroyed by saving and asking information? 
Or checking if health is 0 or less?

mugg91:
Just wanted to mention the bug happened to me in normal gameplay.
https://imgur.com/kRiTr4C.png
I clicked the sleep button at the same time the enemy destroyed the Quarters.

PR #1307, opened 2022-07-04
[fixed?] Knight bomb cycling bugs

### LINKED ISSUE #1263: Bomb type should switch automatically when running out of a type of bombs

Assume a Knight has both types of bombs, water bombs and normal bombs.
Press Activate key and throw the bombs. After you ran out of a type of bomb, the next Activate key press will make the game toggle to the other type of bomb (indicated by the icon at the bottom).
Also, if you have a Lantern in your inventory, then the "toggle" press will make you take out the Lantern into your hand.

Note: If you run out of Fire arrows, Water arrows or Bomb arrows, and you have normal arrows in inventory, it will switch to Normal Arrows automatically without key press. If you don't, it will stay on the current arrow type.

**Proposed Change**

Either switch to other type of bomb automatically after throwing the last one of a type (the icon switches automatically).
Or make the toggle key press activate the next bomb (the icon switches when pressing the key again).
Make it so the Lantern doesn't go into the hand when bombs are still in inventory.

mugg91:
## Status

## Description

Fixes #1263.

[fixed] picking up a mat_waterbombs with no bombs in inventory will correctly set bomb type to waterbombs.
[fixed] Throwing the last normal bomb while having waterbombs in inventory will set bomb type to waterbombs.
[changed] Attaching to a blob won't set the bomb type.
[removed] "cycle" and "switch" commands from KnightLogic.as

This was difficult to find a fix for. 

In onCommand() this code below was removed because it would execute in online only and I think it was also causing some symptoms of the "bomb type not cycling correctly in online" problem.
```
if (this.getCarriedBlob() !is null)
{
	return;
}
```

By adding `Sync()` instances, all symptoms of "bomb type not cycling correctly in online" have been fixed.

Removed "cycle" and "switch" commands since they appear to be unused and don't serve any purpose.

I briefly tested in online/offline vanilla and online/offline PR fix version. This is the outcome:

Previous behavior:
- Activating your last mat_bombs while having mat_waterbombs in inventory will set bomb type to waterbombs, but only in Offline.
- Having no bombs in inventory and picking up a mat_waterbombs will not set the bomb type to waterbombs.
(Offline and Online)
- Placing a bomb in your hand sets the bomb type to that type of bomb.
(Offline and Online)

New behavior: 
- Activating your last mat_bombs while having mat_waterbombs in inventory will set bomb type to waterbombs **now works in Online, too**.
- Having no bombs in inventory and picking up a mat_waterbombs **will correctly set** the bomb type to waterbombs.
(Offline and Online)
- Placing a bomb in your hand **will not set** the bomb type to that type of bomb. 
(Offline and Online)

## Steps to Test or Reproduce

Go to your dedicated server.
!knightshop !999 coins
Be a knight, buy 1 bomb and 1 waterbomb.
Press "action 3" to activate your bomb and throw it away. 

Notice the bomb type will still be regular bomb even though you don't have any more regular bombs. 
Notice if you try to press "action 3" again, it will set bomb type to water bombs without activating your waterbomb.
**After this change, after throwing away your bomb, the bomb type will change to waterbombs instantly and a further press of "action 3" will directly activate your waterbomb.**

---

Have no bombs in your inventory.
Walk across a waterbomb to auto-pick-up it.
Notice the bomb type will still be "nothing".
**After this change, the bomb type will be correctly set to waterbombs.**

---

Place a bomb or waterbomb in your inventory.
Notice the bomb type will change to that type of bomb.
E.g. If you picked up a waterbomb, the bomb type will switch to waterbombs.
**After this change, the bomb type will not change based on what bombs you are picking up.**

## Additional Information

[This music](https://www.youtube.com/watch?v=i6qoIdUyijg) helped me concentrate.

### DIFF
--- a/Entities/Characters/Knight/KnightLogic.as
+++ b/Entities/Characters/Knight/KnightLogic.as
@@ -112,7 +112,6 @@ void onSetPlayer(CBlob@ this, CPlayer@ player)
 	}
 }
 
-
 void RunStateMachine(CBlob@ this, KnightInfo@ knight, RunnerMoveVars@ moveVars)
 {
 	KnightState@[]@ states;
@@ -123,7 +122,7 @@ void RunStateMachine(CBlob@ this, KnightInfo@ knight, RunnerMoveVars@ moveVars)
 
 	s32 currentStateIndex = this.get_s32("currentKnightState");
 
-	if (getNet().isClient())
+	if (isClient())
 	{
 		if (this.exists("serverKnightState"))
 		{
@@ -165,8 +164,6 @@ void RunStateMachine(CBlob@ this, KnightInfo@ knight, RunnerMoveVars@ moveVars)
 		}
 	}
 
-
-
 	u8 state = knight.state;
 	KnightState@ currentState = states[currentStateIndex];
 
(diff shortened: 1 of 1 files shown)


mugg91:
Command "switch" should not be removed because as I found out, it works in conjunction with `Build Modifier` + `1/2`.
Command "cycle" I don't quite understand yet. It seems to have something to do with "tapping" items in your inventory. It probably shouldn't be removed either.
This PR needs some rewriting since some bomb stuff changed.

PR #1308, opened 2022-07-06
[fixed] Withdrawn items making continuous noise

### LINKED ISSUE #1234: Withdrawing wooden items causes sound bug
**Description**

If you withdraw certain items that make a wood sound effect (bucket, water bucket, drill, lantern, mine, log...) from storage/dinghy/etc. directly into your backpack, the game will erroneously keep playing a wooden sound effect when you touch the ground.

The bug does not happen if your backpack is full and you withdraw the item into your hand.

It also doesn't seem to happen if you withdraw from a crate, for some reason.

**Reproduction**

https://www.youtube.com/watch?v=qHh5JjoX7Z0

1. Join sandbox.
2. Deposit items mentioned above.
3. Withdraw one or more of those items into your backpack.
4. Wooden sound effect keeps playing.

**Expected Behavior**

Play no wood sound effect.

**Additional Info**

The bug happens also if you activate a bomb or water bomb, ~~put it in a dinghy quickly and then withdraw it~~ and put that directly in your backpack.
Bomb will play a metallic and water bomb a watery sound effect.

mugg91:
## Status

## Description

Fixes #1234.

After this change, in online withdrawing a wooden item (bucket, log, lantern) from a Storage or Dinghy into your own inventory will not play a continuous wooden sound effect.
In online, placing a bomb or water bomb into your inventory will not play a continuous metallic or watery sound.
In online, a boulder landing on the ground will play a sound effect instead of being silent.

Since `onCollision()` was almost the same between Fabric/Wooden.as and Fabric/Stone.as, I copy-pasted `onCollision()` from the latter into the former. It is possible to place `onCollision()` and `onGib()` of both into a separate common file, but I opted not to in case one is getting its logic and sounds changed later on. Let me know if a common file should still be created anyway.

## Steps to Test or Reproduce

Go to online Sandbox.
Place a bucket in a Dinghy, then withdraw the bucket from the Dinghy into your inventory.
Notice there is a wooden sound effect continuously playing. **After this change, not anymore.**

---

Place a bomb into your inventory.
Notice there is a continuous metallic sound effect playing. **After this change, not anymore.**

---

Throw a Boulder on the ground.
Notice no sound effect playing. **After this change, there will be a sound effect playing.**

### DIFF
--- a/Entities/Common/Fabric/Stone.as
+++ b/Entities/Common/Fabric/Stone.as
@@ -36,12 +36,7 @@ void onGib(CSprite@ this)
 
 void onCollision(CBlob@ this, CBlob@ blob, bool solid, Vec2f normal, Vec2f point1)
 {
-	if (!solid)
-	{
-		return;
-	}
-
-	if (!getNet().isServer())
+	if (!solid || this.isInInventory())
 	{
 		return;
 	}
@@ -76,9 +71,8 @@ void onCollision(CBlob@ this, CBlob@ blob, bool solid, Vec2f normal, Vec2f point
 	const f32 base = heavy ? 5.0f : 7.0f;
 	const f32 ramp = 1.2f;
 
-	//print("stone vel " + vellen + " base " + base );
 	// damage
-	if (getNet().isServer() && vellen > base && !this.hasTag("ignore fall"))
+	if (isServer() && vellen > base && !this.hasTag("ignore fall"))
 	{
 		if (vellen > base * ramp)
 		{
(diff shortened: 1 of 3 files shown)


Vam-Jam:
To be honest, the better solution may be to disable onCollision when in the inventory engine side. In TC, [we do something similar](https://github.com/TFlippy/kag_territorycontrol/blob/master/TFlippy_TerritoryControl_Core_Dev/Gamemode/DisableInventoryCollisions.as), as it fixes some fps issues with huge inventories & it's just a catch all.

I don't see a use case when you need collision when inside another object.

Vam-Jam (approved):

AsuMagic:
Should be fixed engine side by next build, see issue for details

PR #1310, opened 2022-07-06
[fixed?] Archer bot shooting two arrows in one charge in online

mugg91:
## Status

## Description

This change makes it so Archer bots won't shoot two arrows at once in online servers.

ArcherBrain.as causes the bot character to press keys. 
The actions are managed in ArcherLogic.as.
Apparently both the client and the server go through `ManageBow()` and therefore will execute `ClientFire()` twice.
I added this check at the beginning of `ManageBow()` so only the client executes the code:
```
	if (!isClient())
	{
		return;
	}
```

Archer bots in offline will not be affected and will still shoot one arrow as usual.
I have not seen any side effects.
Not sure why it happens only with bots and not players, though...

## Steps to Test or Reproduce

Go to your own dedicated server.
Spawn an archer bot by using !bot
Change team and wait for the archer bot to shoot at you. 
Notice the bot will shoot two arrows. **After this change, will only shoot one arrow.**

---

Go to Save the Princess in offline.
Play until the first archer bot.
Wait for the archer bot to shoot at you.
Notice the bot will shoot one arrow. **After this change, this behavior is not broken.**

### DIFF
--- a/Entities/Characters/Archer/ArcherLogic.as
+++ b/Entities/Characters/Archer/ArcherLogic.as
@@ -305,10 +305,15 @@ void ManageGrapple(CBlob@ this, ArcherInfo@ archer)
 
 void ManageBow(CBlob@ this, ArcherInfo@ archer, RunnerMoveVars@ moveVars)
 {
+	if (!isClient())
+	{
+		return;	
+	}
+	
 	//are we responsible for this actor?
 	bool ismyplayer = this.isMyPlayer();
 	bool responsible = ismyplayer;
-	if (isServer() && !ismyplayer)
+	if (!ismyplayer)
 	{
 		CPlayer@ p = this.getPlayer();
 		if (p !is null)


Vam-Jam (approved):
Works good both locally and on a dedicated server.

AsuMagic (changes requested):
After some investigation, it turns out you kind of got the fix backwards... The reason is in fact rather weird. *There is no both client and server ticking of the bot happening server-side.* The actual issue is that clients (as in actual players), for some reason, have the authority to make the bot shoot arrows, and do just that.

It probably doesn't happen to non-bot archers because they don't have the authority to as actual players "own" their blob from a netcode standpoint.

The result is that N+1 arrow shooting can occur, where N is the amount of players that may tick the bot (i.e. the players in proximity).

The function's logic for identifying whether whatever is executing the script is responsible for the blob appears to be OK:

```angelscript
	bool ismyplayer = this.isMyPlayer();

	bool responsible = ismyplayer;
	if (isServer() && !ismyplayer)
	{
		CPlayer@ p = this.getPlayer();
		if (p !is null)
		{
			responsible = p.isBot();
		}
	}
```

So what must be missing is checking for `responsible` in a couple of places inside of `ManageBow`... Note that it might not only be about firing arrows per se, but also some cooldown updates, so one would need to review that whole function just in case.

EDIT: For clarity: It _is_ okay for the client and the server to be ticking that function. What's unintended is that some actions that should only be done by the responsible party are run by everyone here.

If you want to deal with this then do it, otherwise let's close the PR and leave the issue opened.

mugg91:
Going to fulfill @AsuMagic's request and closing this for now.
Can try to find a better fix later maybe.

mugg91:
I tested the repro steps again. When spawning an archer bot in online, he will shoot only 1 arrow instead of the expected 2.
So this issue might have been fixed at some point.

PR #1318, opened 2022-07-15
Better bomb throwing with ping

bunniewormy:
## Status

## Description

Alternative to #1038 , which made it impossible to spam bombs if you had high ping (which caused many complaints)
Implements this comment:
"Instead of the client checking if it's holding something or needs to light a new bomb, the server should do that. Pretty much move all the checks the client is doing onto the server instead. The client just needs to blindly tell the server that they want to light/throw a bomb and the server acts on in."
Moved some code from ActivateHeldObject.as to ThrowCommon.as as well
Video - comparison between PR version and pre-Epsilon's PR bomb throwing.

https://www.youtube.com/watch?v=-4n0ouJ2m80

### DIFF
--- a/Entities/Characters/Knight/KnightLogic.as
+++ b/Entities/Characters/Knight/KnightLogic.as
@@ -1113,7 +1113,21 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
 
 			if (carried !is null)
 			{
-				DoThrow(this, carried, pos, vector, vel);
+				bool holding_bomb = false;
+				// are we actually holding a bomb or something else?
+				for (uint i = 0; i < bombNames.length; i++)
+				{
+					if(carried.getName() == bombNames[i])
+					{
+						holding_bomb = true;
+						DoThrow(this, carried, pos, vector, vel);
+					}
+				}
+
+				if (!holding_bomb)
+				{
+					ActivateBlob(this, carried, pos, vector, vel);
+				}
 			}
 			else
 			{
--- a/Entities/Characters/Knight/KnightLogic.as
+++ b/Entities/Characters/Knight/KnightLogic.as
@@ -83,6 +83,7 @@ void onInit(CBlob@ this)
 	this.Tag("flesh");
 
 	this.addCommandID("get bomb");
+	this.addCommandID("activate/throw bomb");
 
 	this.push("names to activate", "keg");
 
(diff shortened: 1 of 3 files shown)


Entities/Characters/Knight/KnightLogic.as
```diff
@@ -1154,6 +1150,30 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
 	{
 		SetFirstAvailableBomb(this);
 	}
+	else if (cmd == this.getCommandID("activate/throw bomb"))
```
AsuMagic, line 1153: I'm a bit confused: `"get bomb"` cannot be sent straight from the client anymore, correct? If `"activate/throw bomb"` is run on the serverside, why do we need to fire a `"get bomb"` command as opposed to just calling the relevant code directly?

This is not just for style FWIW, because I think the fired command will only be effective after a tick, but I'm not sure.

bunniewormy: fixed

AsuMagic:
With heavy network latency emulation I have found two bugs with the current approach. I don't know if they're new at all, but they're related to this code anyway, so might as well fix it properly once and for all.
- **If** you have at least one bomb material in your inventory, and pick up a keg, and use space, the keg will be dropped immediately but it won't be lit.
- If you have both bomb and waterbomb materials in your inventory, select regular bombs, then switch to waterbomb and press space immediately, the switch will not be registered and you will be throwing a regular bomb instead.

bunniewormy:
> If you have both bomb and waterbomb materials in your inventory, select regular bombs, then switch to waterbomb and press space immediately, the switch will not be registered and you will be throwing a regular bomb instead.

Doesn't seem to be fixable script-side - CGridButton clicks (which send the command to select the bomb type) seem to be hardcoded to wait for the button click to arrive on server before the command is sent to client. 
(Unless you want me to write an entire script-side button class for this, which I can do :^) )

> If you have at least one bomb material in your inventory, and pick up a keg, and use space, the keg will be dropped immediately but it won't be lit.

Fixed

AsuMagic:
> Doesn't seem to be fixable script-side - CGridButton clicks (which send the command to select the bomb type) seem to be hardcoded to wait for the button click to arrive on server before the command is sent to client.

oh, right, cringe

AsuMagic (approved):

PR #1321, opened 2022-07-19
[fixed] Drill activating when player is attached in online server

mugg91:
## Status

## Description

When the player is attached (sitting on a vehicle or animal) and is pressing Action1, drill would activate, destroying nearby blobs, scenery or walls. This happens only in online servers.
This PR will add an early return, so the drilling doesn't happen if the player is attached.

## Reproduction Steps

Go to online Sandbox EU.
Spawn a catapult and a drill.
Pick up the drill and sit in the catapult.
Hold Action1 while driving around.
Notice the drill will damage nearby platforms, trees, chickens, etc.
**After this change, not anymore.**

### DIFF
--- a/Entities/Industry/Drill/Drill.as
+++ b/Entities/Industry/Drill/Drill.as
@@ -186,7 +186,7 @@ void onTick(CBlob@ this)
 		AttachmentPoint@ point = this.getAttachments().getAttachmentPointByName("PICKUP");
 		CBlob@ holder = point.getOccupied();
 
-		if (holder is null) return;
+		if (holder is null || holder.isAttached()) return;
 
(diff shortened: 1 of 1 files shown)


AsuMagic:
> This happens only in online servers.

I was curious why: turns out `AnimalRiding.as` calls `ap.SetKeysToTake(key_left | key_right | key_up | key_down | key_action1 | key_action2 | key_action3);`. `SetKeysToTake`, apparently, has never been properly implemented on net.

AsuMagic (approved):

AsuMagic:
Let's merge this as a workaround, for now.

PR #1325, opened 2022-07-19
[fixed] CTF Flag return timer bugs

### LINKED ISSUE #1319: CTF flag timer bugs
## Description

1) When picking up an enemy flag and throwing it on the first possible tick, the return timer will not reset.
2) Picking up an enemy flag on the last possible tick before it would return, if you throw it it will return.

## Reproduction

Reproduction Steps for 1)
- Go to Sandbox and spawn a flag_base, switch team.
- Pick up the flag and throw it somewhere, wait a bit for the return timer to deplete.
- Walk into it and press the Throw key at the same time.
- If done correctly, notice the return timer will not have reset.

Reproduction Steps for 2)
- Go to Sandbox and spawn a flag_base, switch team.
- Pick up the flag and throw it somewhere, wait for the return timer to deplete to zero.
- Walk into it at the last possible moment and then throw it.
- If done corrrectly, notice the flag will instantly return.

mugg91:
## Status

## Description

Fixed #1319.

If you pick up an enemy CTF Flag and throw it on the same tick, its return timer would not reset.
If you pick up an enemy CTF Flag on the last possible moment, it is possible that it could have the "return" flag, so if you throw it, it would instantly return to its base.
This PR aims to fix these bugs by adding these lines to the beginning of `onAttach()`.
```
	this.Untag("return");
	this.set_u16(return_prop, 0);
	this.Sync("return", true);
	this.Sync(return_prop, true);
```
Initially, I wanted to have this code execute on client-side only. But the server could lag and the bugs could still happen, so I now have both client and server enter these lines and run `Sync()`. I have not seen the bugs happen after that.

## Steps to Test or Reproduce

To reproduce "pick up and throw on the same tick", walk into an enemy flag and press the throw key at the same time. It will take a few tries. If done correctly, you will notice the return timer will not have reset.

To reproduce "picking up in the last moment and throwing will return the flag instantly", you can set `return_time` in line 7 to a low number such as 30 and keep trying. You can add a line to `onAttach()` such as this for checking:
`if (this.hasTag("return")) print("If you let go of me now I will return to my base.")`

### DIFF
--- a/Entities/Special/CTF/CTF_Flag.as
+++ b/Entities/Special/CTF/CTF_Flag.as
@@ -283,8 +283,12 @@ bool doesCollideWithBlob(CBlob@ this, CBlob@ blob)
 
 void onAttach( CBlob@ this, CBlob@ attached, AttachmentPoint @attachedPoint )
 {
-    if (!getNet().isServer())
-        return;
+    this.Untag("return");
+    this.set_u16(return_prop, 0);
+    this.Sync("return", true);
+    this.Sync(return_prop, true);
+	
+    if (isServer())	return;
 
     CRules@ rules = getRules();
 


AsuMagic:
Superseded by #1580 which was merged

Issue #204, opened 2018-08-01
[Feature Pitch] Gamemode: War [Massive]


Radinyn:
**New gamemode: War**
**Description:** Maps are really big, each team starts in an war base (!war_base command). On the map you can find empty halls, houses (trader shops) and tents. The goal is to destroy (not capture, that would be too easy) enemy team's war base.

**Halls** - in this gamemode both normal workshop and factory workshop are allowed, basically you can you factory workshops only near halls, out there you can produce items the same way as in TTH. You can buy more workers with gold.

**NPCs and houses** - when you find a house you can take ownership of it for your team, houses basically increase the population, and they can be upgraded with gold to the certain lvl to give more (not to much, we don't want servers to blow). When you have enough population you can buy NPCs in your war base. Of course there are 2 types of them - knights and archers - both to protect your land from those scums in other team. You can move them where you want to make the protection as good as you can. If you lose houses NPCs may turn rebelious and join team -1.
*Edit: NPCs respawn over time, unless they would be worthless with their potato ai.

**Tents** - tents are locations on the map filled with bandits (knights and archers) from team -1, they do not respawn. Under their base you can find chests (from tdm) with coins, bombs and arrows and of course - gold for your upgrades.

This gamemode is probably really hard to be made, however if it will become a thing one day it will most likely be a lot of fun and should gather quite a few players. Maps aren't too much of a problem I can make some with help from my friends, however if someone would be brave enough to take the task of coding it, I would be really grateful. Cheers! :)

Special thanks to Homek for helping me to design this gamemode.

Mazey:
I have work done on a "fixed" war base, could probably use that.

Radinyn:
Yeah that would be nice

Uristthedorf:
Is the kag user base big enough for this?

Enderzilla747:
Maybe if it goes F2P lol

Radinyn:
For playing it? Definitely that's why there are bots! That's literally the whole point.

For creating it? Maybe.

1bardesign:
Sorry, we reviewed this and it's just too big for a realistic feature pitch at this point. You may be able to talk with modders and work out a "non-official" way to get it developed, but I'm closing this out of the feature pitch system. Thanks for your time! 👍

PR #1327, opened 2022-07-23
Improving prickup: prioritize hovered blobs 

Nananas:
## Status

## Description

When hovering the a blob, prioritize pickup of this blob. Checks which blob is drawn at top if stacked.
Related video: https://discord.com/channels/337250580303052801/337266863841214465/756102703280553994

This PR currently uses the CShape of an item to check mouse overlap, together with the draw order. 
However, some items such as bomb-arrows, gold and the flag have CShapes that differ from the actual sprite.
Modifying the CShape to match the sprite does not feel like the correct solution.

After testing, I found some issues with the a previous engine change, which should be fixed in Engine PR `#204`.

Commit 23271f9 adds the z-ordering (SetZ) and "hover-polygons" of some "important" items to draw them on top of other stuff. This is done by adding the ImportantPickup.as script, also see the comments in that file. 

I already added "hover-polygons" to the flag and the special arrows, as they differ a lot from their CShape.

(This PR is a duplicate of Gitea PR 7)


Entities/Characters/Scripts/RunnerDefault.as
```diff
@@ -58,12 +58,6 @@ void onAttach(CBlob@ this, CBlob@ attached, AttachmentPoint @attachedPoint)
 	this.server_DetachFrom( attached ); CRASHES*/
 }
 
-// set the Z back
```
AsuMagic, line 61: Why is this removed?

This appears to cause an issue when leaving a catapult, for instance. The player Z appears to be modified when entering a catapult to visually appear behind it, but with this change, the player will still appear behind the catapult.

Entities/Materials/Ammunition/MaterialWaterArrow.as
```diff
@@ -9,4 +9,17 @@ void onInit(CBlob@ this)
   this.maxQuantity = 2;
 
   this.getCurrentScript().runFlags |= Script::remove_after_this;
+
+  // Pickup: custom hover area
+  Vec2f[] hoverRect =
```
AsuMagic, line 14: This is repeated across all arrow material scripts, maybe introduce a `genericArrowHoverRect` constant in a `ArrowCommon.as` script? Or some kind of constants script (e.g. `ArrowConstants.as`) though nothing else in KAG seems to follow such a convention (yet)...

Entities/Common/Controls/ImportantPickup.as
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  this.getSprite().SetZ(baseZ);
+  this.set_f32("important-pickup", baseZ);
+}
+
+void onAttach(CBlob@ this, CBlob@ attached, AttachmentPoint @attachedPoint)
+{
+  f32 baseZ = this.getSprite().getZ();
+  this.set_f32("important-pickup", baseZ);
```
AsuMagic, line 54: Why do we need to check the sprite Z on attach? Looking at the logic my assumption would be that the sprite Z is not expected to change, so that `"important-pickup"` should always be correct.

This is not *really* important but is this something you've seen happen?

Entities/Common/Controls/ImportantPickup.as
```diff
@@ -0,0 +1,102 @@
...
+
+void onDetach(CBlob@ this, CBlob@ detached, AttachmentPoint@ attachedPoint)
+{
+  f32 baseZ = this.get_f32("important-pickup");
+  this.getSprite().SetZ(baseZ);
+}
+
+// Add ImportantPickup.as to sprite_scripts in the .ctf file to enable rendering of 
+// the hover square (for debug purposes)
+void onRender(CSprite@ this)
```
AsuMagic, line 66: The debug HUD is nice for testing... Though, TBH, the script may as well add itself to the sprite and have the function early exit if `g_debug == 0`. Because right now, if one wants to check/rework the hover shapes they would need to find all blobs that use the script to edit their `.cfg` individually.

Entities/Common/Controls/StandardControlsCommon.as
```diff
@@ -40,3 +40,61 @@ bool isTapPickup(CBlob@ this, int ticks = 15)
...
+        Vec2f q = Vec2f(-polyVerts[i].x, polyVerts[i].y);
+        mirrored.push_back(q);
+      }
+
+      polygon = mirrored;
+    } 
+    else 
+    {
+      polygon = polyVerts;
+    }
```
AsuMagic, line 65: Scrap the local variable. `Vec2f[] polyVerts` is already passed by value, so the original will be untouched anyway.

```suggestion
bool PointInsidePolygon( Vec2f Point, Vec2f[] polygon, Vec2f polyPos, bool facingLeft )
{
    // Mirror the polygon when the blob is facing left
    if (facingLeft) 
    {
      for ( int i = 0 ; i < polygon.length ; i++ )
      {
        polygon[i].x = -polygon[i].x;
      }
    }
```

Entities/Common/Controls/StandardControlsCommon.as
```diff
@@ -40,3 +40,61 @@ bool isTapPickup(CBlob@ this, int ticks = 15)
...
+    {
+        return false;
+    }
+
+    bool inside = false;
+    for ( int i = 0, j = polygon.length - 1 ; i < polygon.length ; j = i++ )
+    {
+    	Vec2f pvi = polyPos + polygon[ i ];
+    	Vec2f pvj = polyPos + polygon[ j ];
+        if ( ( pvi.y > Point.y ) != ( pvj.y > Point.y ) &&
```
AsuMagic, line 91: I'll trust that the math is correct as it appears to work :P

AsuMagic (changes requested):
Does appear to work. Generally speaking the approach seems sound, but it will need a good amount of real world testing to find issues (bugs or finding other shapes that may need hover shapes). On top of my head, some things that should be good to check in testing:
- This changes a few Z values, most notably the keg one. It should be checked whether this causes unintended effects.
- We might as well look over some more defaults, such as the bomb Z which has not been mentioned in the PR.

Entities/Special/CTF/CTF_Flag.cfg
```diff
@@ -3,7 +3,7 @@
 # @ array
 
 $sprite_factory                                   = generic_sprite
-@$sprite_scripts                                  = CTF_Flag.as;
+@$sprite_scripts                                  = CTF_Flag.as; ImportantPickup.as
```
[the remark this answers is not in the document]
Nananas: No idea how I missed this, woops

Entities/Materials/Ammunition/MaterialBombArrow.as
```diff
@@ -6,4 +6,19 @@ void onInit(CBlob@ this)
   this.set_u16('decay time', 300);
 
   this.getCurrentScript().runFlags |= Script::remove_after_this;
+
+  // Pickup: custom hover area
```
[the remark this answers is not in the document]
AsuMagic: None that really matters, no. It only sets `"decay time"` which is only matters to the server.

Entities/Special/CTF/CTF_Flag.cfg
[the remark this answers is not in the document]
Nananas: Removed

Entities/Materials/Ammunition/MaterialBombArrow.as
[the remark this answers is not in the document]
Nananas: I removed the `SERVER_ONLY` define. I hope that's fine.

Entities/Characters/Scripts/RunnerDefault.as
Nananas: I reverted the remove. I always forget to test with catapults.

Entities/Materials/Ammunition/MaterialWaterArrow.as
Nananas: I moved the shape into `ArrowCommon.as`

Entities/Common/Controls/ImportantPickup.as
Nananas: Added a `g_debug` check

Entities/Common/Controls/StandardControlsCommon.as
Nananas: I got that from somewhere on the ~~ internet ~~

Entities/Common/Controls/ImportantPickup.as
Nananas: I removed `onAttach` for now. It does not make sense at the moment to modify the baseZ. Maybe just old code

AsuMagic (approved):

AsuMagic:
This will need some real world testing and feedback, also keeping an eye out for anything weird with Z levels that may arise (notably related to attachments such as pickup, catapults, etc.).

PR #1160, opened 2021-11-10
[CTF Rebuild] Fire arrows now light team kegs as well

bunniewormy:
## Status

(pick one)

- **READY**: this PR is (to the best of your knowledge) ready to be incorporated into the game.

## Description

[changed] fire arrows can now light team kegs

PR #1329, opened 2022-07-23
Set the "fire source" tag as early as onInit for arrows

AsuMagic:
## Status

## Description

This fixes an issue with #1160 where the keg would often not lit up if the fire arrow was shot from a close enough distance. For some reason, the tag assignment was guarded by a `gametime % 6` condition on the arrow ticking, which was used to play some effects. Hence, the late tagging looks unintentional, and I have not found issues with this fix during my testing.

## Steps to Test or Reproduce

1. Go to sandbox (online or offline)
2. Switch to archer
3. Spawn a keg
4. Shoot a fire arrow at the keg. It should work even if you're basically overlapping the keg, and at short distances.

Without this workaround the arrow would miss most of the time which is largely undesirable.

### DIFF
--- a/Entities/Items/Projectiles/Arrow.as
+++ b/Entities/Items/Projectiles/Arrow.as
@@ -46,15 +46,18 @@ void onInit(CBlob@ this)
 
 	const u8 arrowType = this.get_u8("arrow type");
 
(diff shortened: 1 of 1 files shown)


Vam-Jam (approved):
3:

PR #1331, opened 2022-07-24
[fixed] getting stuck in a glitched state if getting in and out a crate at the same time

### LINKED ISSUE #1330: Get stuck after popping out of your crate
## Description

When you hold your crate and get in while also pressing "pick up" to get out, you may instantly pop out and become stuck in a glitched state.

This happens in offline sandbox.

I think the bug happens if you do it correctly on the first try. If you pop out normally, then the bug will not happen unless you suicide and try again.

## Reproduction

To make testing easy, you can edit Crate.as to copypaste the code from the "get out" command over to the end of the "get in" command such as this:

```
else if (cmd == this.getCommandID("getin"))
	{
		if (this.getHealth() <= 0)
		{
			return;
		}
		CBlob @caller = getBlobByNetworkID( params.read_u16() );
		print(caller.isAttached() + " " + caller.isAttachedTo(this) + " " + caller.hasAttached());

		if (caller !is null && this.getInventory() !is null) 
		{
			CInventory@ inv = this.getInventory();
			u8 itemcount = inv.getItemsCount();

			// Boobytrap if crate has enemy mine
			CBlob@ mine = null;
			for (int i = 0; i < inv.getItemsCount(); i++)
			{
				CBlob@ item = inv.getItem(i);
				if (item.getName() == "mine" && item.getTeamNum() != caller.getTeamNum())
				{
					CBitStream params;
					params.write_u16(caller.getNetworkID());
					params.write_u16(item.getNetworkID());
					this.SendCommand(this.getCommandID("boobytrap"), params);
					return;
				}
			}

			// We might have to make room
			while (!inv.canPutItem(caller) && itemcount > 0)
			{
				// pop out last items until we can put in player or there's nothing left
				CBlob@ item = inv.getItem(itemcount - 1);
				this.server_PutOutInventory(item);
				float magnitude = (1 - XORRandom(3) * 0.25) * 5.0f;
				item.setVelocity(caller.getVelocity() + getRandomVelocity(90, magnitude, 45));
				itemcount--;
			}

			Vec2f velocity = caller.getVelocity();
			this.server_PutInInventory( caller );
			this.setVelocity(velocity);
		}
		
		CBlob@ sneaky_player = getPlayerInside(this);
		if (caller !is null && sneaky_player !is null) {
			if (caller.getTeamNum() != sneaky_player.getTeamNum())
			{
				if (isKnockable(caller))
				{
					setKnocked(caller, 30);
				}
			}
			this.Tag("crate escaped");
			this.server_PutOutInventory(sneaky_player);
		}
		// Attack self to pop out items
		this.server_Hit(this, this.getPosition(), Vec2f(), 100.0f, Hitters::crush, true);
		this.server_Die();
	}
```

Go to your offline Sandbox and try to get in a crate. Notice you will get stuck.

mugg91:
## Status

## Description

Fixed #1330.

In offline, you can get stuck if getting in and out a crate at the same time.
The cause of the problem is `server_PutInInventory()` and `server_PutOutInventory()` happening on the same tick.
This PR will make it so the two lines don't happen on the same tick by saving a timestamp when getting in and checking for that timestamp when getting out.

This being a workaround fix, it should be looked into what's happening in `server_PutInInventory()` and `server_PutOutInventory()`  to make this bug happen to find the actual root cause. The bug doesn't seem to happen if you got out a crate successfully before.

## Steps to Test or Reproduce

You can edit Crate.as to copypaste the code from the "get out" command to the end of the "get in" command to simulate getting in and out at the same time.
After this change, the "get out" part is ignored if it happens on the same tick you get inside.


mugg91:
Nvm, the problem isn't `this.server_PutInInventory()` and `this.server_PutOutInventory()` on the same tick.

It is `this.server_PutInInventory()` and `this.server_Die()` on the same tick.

mugg91:
The bug scenario can theoretically happen with any other `server_PutInInventory()` instance when it happens at the same time the crate (or other inventory blob) is destroyed.

However, when I tested putting a burger in a crate at the same time the crate is destroyed, it will simply fall down or it will vanish for 2 seconds and then reappear and fall down. The floating in midair doesn't seem to happen. But there are a lot more possible cases that need checking. Such as when materials are put in a crate automatically at the same time it's destroyed.

mugg91:
The bug still happens so I fixed this PR's file conflict. 
It's ready.

Video of the bug:

https://github.com/user-attachments/assets/96aa2125-5ed2-4176-91e1-1c4179003b57

PR #1332, opened 2022-07-24
[fixed] picking up items that have already been attached / picking up items after getting in a crate

### LINKED ISSUE #1320: Picking up stuff while going in a crate (online bug)
## Description

Works best at high ping. 
Pick up an empty crate and keep showing buttons.
Then throw the crate aside.
You should still see the "enter crate" button.
If you press it and quickly pick up another item, you will be inside the crate, and the item you wanted to pick up will be stuck and considered to be in your hand.
While inside the crate, you can put it in the same crate you are in (unless it does not fit or does not go in inventories).
While holding something like this while inside a crate, the "get out" button will be greyed out.

If you do this with an enemy ctf flag, the return timer will deplete but will not return when it reaches 0.
Once you get out of the crate, you will instantly grab the flag.

It is possible to pick up the crate that you got into.
If you do this while in water, you will sink to the bottom and the crate will not decay.

Video: https://www.youtube.com/watch?v=igcOM0DnvFc

### LINKED ISSUE #1323: Picking up stuff from one's hand when laggy
**Description**

Laggy players can pick up items that other just picked up a few ticks ago because of lag.  
This generally is frustrating and surprising, and gives an advantage to laggy players for item pickup.

**Reproduction**

Player A has 33ms ping. Player B has 133ms ping.

Player A picks up an item. The server registers the pick up.  
Player B picks up the item, because they have not registered the pick up on their client yet.  
Player B effectively steals the item player A has picked up.

**Expected Behavior**

If the item is already picked up server-side by player A, the server should not allow player B to pick it up.

### LINKED ISSUE #1519: You can pick up a Dinghy at the same time someone gets in
## Description

It is possible that one player presses down to get in a Dinghy while another player picks it up.
Then the player who picked up the Dinghy can move around with it as normal while the player that sits in the Dinghy just keeps attached until they get out.

## Proposed Fix

Only go ahead with the pickup if nothing is attached.

mugg91:
## Status

## Description

Fixes #1320:

At high ping in online, you could pick up items after sending the "get in" command but before you got put in the crate.
This PR makes it so `onCommand()` in StandardControls.as checks if the picker is in inventory. If yes, the picking up will not happen.

Fixes #1323:

At high ping in online, it is possible that a player picks up an item that has already been picked up or attached.
This PR makes it so the pickup does not happen if the picked item is attached.

Fixes that you could go to Bed or attach yourself and then pick up an item in quick succession.

Fixes #1519:

It was possible to pick up a Dinghy that someone got into just a few ticks ago.
Commit [c73b3b1](https://github.com/transhumandesign/kag-base/pull/1332/commits/c73b3b1331a7f35bb0e10cc25c675b573992eab4) fixes that so it will not be possible.

## Steps to Test or Reproduce

Go to a high ping server.
Have a crate in hand and keep showing buttons.
While showing buttons, throw the crate aside.
Be near the crate or any item and click on the "get in" button and then press "pick up" in quick succession.
The crate (or other item) will be picked up while you are inside the crate.

**After this change, you will get inside the crate and the pickup will not happen.**

---

To test "picking up item from another player's hand", I used this code in BuilderLogic.as:

```
int z=0;
void onTick(CBlob@ this)
{
	if (isServer())
	{	
		if (this.getTeamNum() == 1)
		{
			z++;
		
			CBlob@[] crates;
			getBlobsByName("crate", crates);
			for (uint i = 0; i < crates.length; i++)
			{
				if (z%80==0)
				{
					if (crates[i].isAttached())
					{
						crates[i].server_DetachFromAll();
						break;
					}
				
					else
					{
						server_Pickup(this, this, crates[i]);
						break;
					}
				}

			}
		}
	}
```

Go to your own dedicated server. Use [clumsy](https://jagt.github.io/clumsy/) to simulate high ping.
Spawn a red builder, then go back to being blue.
Spawn a crate. The code above will cause the crate to repeatedly get picked up by and then detached from the red builder.
If you pick up the crate before the red builder, the "bug scenario" can happen.
**After this PR , not anymore.**

---

In online, go heal in a Bed and pick up an item in quick succession.
Notice you will hold your item while you are sleeping.
**After this PR, not possible anymore.**

### DIFF
--- a/Entities/Common/Controls/StandardControls.as
+++ b/Entities/Common/Controls/StandardControls.as
@@ -60,7 +60,8 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
 		    && !owner.isInInventory()
 		    && !owner.isAttached()
 		    && pick !is null 
-		    && !pick.isAttached())
+		    && !pick.isAttached()
+		    && !pick.hasAttached())
 		{
 			owner.server_Pickup(pick);
 		}
--- a/Entities/Common/Controls/StandardControls.as
+++ b/Entities/Common/Controls/StandardControls.as
@@ -56,7 +56,10 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
 		CBlob@ owner = getBlobByNetworkID(params.read_netid());
 		CBlob@ pick = getBlobByNetworkID(params.read_netid());
 
-		if (owner !is null && pick !is null && !owner.isInInventory())
+		if (owner !is null 
+		    && !owner.isInInventory()
+		    && pick !is null 
+		    && !pick.isAttached())
 		{
 			owner.server_Pickup(pick);
 		}
--- a/Entities/Common/Controls/StandardControls.as
+++ b/Entities/Common/Controls/StandardControls.as
@@ -58,6 +58,7 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
 
 		if (owner !is null 
 		    && !owner.isInInventory()
+		    && !owner.isAttached()
 		    && pick !is null 
 		    && !pick.isAttached())
 		{
--- a/Entities/Common/Controls/StandardControls.as
+++ b/Entities/Common/Controls/StandardControls.as
@@ -61,7 +61,7 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
 		    && !owner.isAttached()
 		    && pick !is null 
 		    && !pick.isAttached()
-		    && !pick.hasAttached())
+		    && pick.canBePickedUp(owner))
 		{
 			owner.server_Pickup(pick);
 		}
--- a/Entities/Common/Controls/StandardControls.as
+++ b/Entities/Common/Controls/StandardControls.as
@@ -56,7 +56,7 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
 		CBlob@ owner = getBlobByNetworkID(params.read_netid());
 		CBlob@ pick = getBlobByNetworkID(params.read_netid());
 
-		if (owner !is null && pick !is null)
+		if (owner !is null && pick !is null && !owner.isInInventory())
 		{
 			owner.server_Pickup(pick);
 		}


mugg91:
We should consider adding another check `!pick.hasAttached())` because it could be possible someone sits in a Dinghy that is about to be picked up.

mugg91:
The issue I brought up in my previous comment actually existed ( #1519 ) so I made a commit that will fix it.

Reproduction Steps:

1) Add this hack to `Lantern.as onTick()`:
```
	CBlob@ b = getBlobByName("archer");
	CBlob@ d = getBlobByName("dinghy");
	if (b !is null && d !is null && isServer())
	{
		if (getGameTime() % 140 == 0)
		{
			if(!d.isAttached())
				d.server_AttachTo(b, "ROWER");
			
		}
		else if (getGameTime() % 140 == 100)
		{
			b.server_DetachFromAll();
		}
	}
```
2) Remove `this.getCurrentScript().runFlags |= Script::tick_inwater;` from `Lantern.as`.
3) Run [clumsy](http://jagt.github.io/clumsy/) to simulate 300 ping.
4) Run `dedicatedserver.bat` and join it.
5) Spawn `!dinghy`, `!archer` and `!lantern`.
6) The archer will now repeatedly get in and out of Dinghy. Try to pick up the Dinghy at the same time the archer gets in.

You will notice, before commit [c73b3b1](https://github.com/transhumandesign/kag-base/pull/1332/commits/c73b3b1331a7f35bb0e10cc25c675b573992eab4) it's possible to do it easily. But it will never happen after [c73b3b1](https://github.com/transhumandesign/kag-base/pull/1332/commits/c73b3b1331a7f35bb0e10cc25c675b573992eab4).

Entities/Common/Controls/StandardControls.as
```diff
@@ -56,7 +56,12 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
 		CBlob@ owner = getBlobByNetworkID(params.read_netid());
 		CBlob@ pick = getBlobByNetworkID(params.read_netid());
 
-		if (owner !is null && pick !is null)
+		if (owner !is null 
+		    && !owner.isInInventory()
+		    && !owner.isAttached()
+		    && pick !is null 
+		    && !pick.isAttached()
+		    && !pick.hasAttached())
```
Gingerbeard5773, line 64: !pick.hasAttached() could potentially mess with some mods, so instead use pick.canBePickedUp(). This still works the same with the dinghy.
```suggestion
		    && pick.canBePickedUp(owner))
```

Gingerbeard5773 (changes requested):

AsuMagic (approved):

PR #1335, opened 2022-07-25
[added] New item to Boatshop, Fishing Rod

mugg91:
## Status

## Description

This is an item I developed. 
It is still in development as many things are yet to do. 
It currently only works in offline.

You can go to Sandbox and either spawn !fishingrod or !boatshop and buy it there.
It consists of a rod (the fishingrod), which has an attachment point TIP.
There are separate fishingline blobs that are basicly nodes that are added to the fishingrod, the first one being attached to the rod's TIP, and all other ones are just floating around and correct their position based on the next segment at a time.
The last fishingline segment is the hook which is given its own attachment point HOOK.

Using !fishingline won't work as it immediately despawns if it isn't part of the rod.

You can put the fishing rod in your inventory. Upon doing so, or getting close to a wall, or attaching the rod to something else (catapult mag)*, the fishingline segments are removed. They are re-added when getting away from a wall or when picking up the rod.

(EDIT: * Nevermind it doesn't, but it's supposed to.)

Use "Action1" to extend the line and use "Use" to pull the line in.

There is a list of blobs to catch - they will attach to the hook's attachment point HOOK (e.g. bucket, sponge, fishy etc.).
And there is a second list of blobs to catch - they will attach the hook to themselves, by adding a MOUTH attachment point to them (e.g. shark, bison). It is not yet implemented but planned to make them pull away from the rod user and pull them along with them using `AddForce()`.
Using "Use", there will be a Detach button to detach the caught blob from the hook.

At the end of FishingRod.as, there is a todo list.
I tried my best and I'm aware I worked on this project without speaking with the devs, I did it for fun and I'm not expecting anything back. Maybe it will be worth in some way.
Let me know if you have ideas or any feedback.

I will add a gameplay video later.


mugg91:
I haven't done more work on this and probably won't in the near future. 
New content would be nice but a fishing rod may be too situational or unfitting. 
It was a good learning project back then for sure.
If anyone wants to use my code, feel free.

PR #1166, opened 2021-11-14
Add banners for game start, build phase start

bunniewormy:
## Status

(pick one)

- **READY**: this PR is (to the best of your knowledge) ready to be incorporated into the game.

## Description

[added] "Capture the flag!" banner at game start, currently for CTF gamemodes only
[added] "Build the base!" banner at build base start, currently for CTF gamemodes only
[added] added the team wins banner to small CTF (previously it didn't use banners at all)

teamwinbanner.as is now empty with an #include of GameStateBanners.as to not break mods 

There's some text in the build the base banner which will be correct only if CTF Rebuild's faster build time building will be merged in (https://github.com/transhumandesign/kag-base/pull/1151)

PR #1346, opened 2022-07-30
[fixed] Minimap on top of map voting window and exception in CTF when match ends via vote pass or chat command

### LINKED ISSUE #1343: Minimap is rendered on top of post game map vote
## Description

This assumes you have minimap turned on in your settings.

When winning a game of CTF normally, the post game map voting will start and the minimap will be disabled.
When the next map loads, the minimap will be enabled again.

But if you use chat command `!endgame` while being mod on your server or if a "Next map" vote passes, then the post game map voting will occur without the minimap being disabled, so the minimap will be rendered on top of the vote window.
There is no instance of `this.minimap = false;`.

These observations are based on Build 3900.

## Additional

I noticed `GameStateBanner.as` which was added recently, will handle turning the minimap on and off, but I don't know if anything has been fixed by this:
```
bool minimap = true;

// ...

void onRestart(CRules@ this)
{
	// ...
	this.minimap = minimap;
	// ...
}

// ...

void SetBanner(CRules@ this)
{
// ...

this.set_u8("Animate Banner", BannerType::GAME_END);
this.minimap = false;
```

## Reproduction

Do one of these:
- Finish a game of CTF normally - Notice no problems.
- Start a "next map" vote and have it pass - Notice the minimap will be on top of the next map voting.
- Use chat command `!endgame` - Notice the minimap will be on top of the next map voting.

## Suggestions

Set `rules.minimap = false;` when game state banner or map voting is shown, set to true otherwise.

mugg91:
## Status

## Description

This PR fixes an exception that could happen when using chat command `!endgame` or when "Next map" vote passes, in a CTF server, with the newly introduced game state banners (See PR #1166 which introduced GameStateBanners.as and CTF_Banners.as).

Variable `banner_type` could become equal to the length of the `banners` array which causes an out of bounds exception.
I made it so it will return from the function if that's the case and when `banners[banner_type]` is about to be used.

Fixes #1343:

Previously, in a nutshell, `GameStateBanners.as` would set `rules.minimap = true` when a new map loads.
When the game finishes due to a team winning, the winning banner is shown and `rules.minimap = false` executes, so the minimap isn't shown until the next map loads.
When a "Next map" vote passes or when using chat command `!endgame`, having applied the above fix, the minimap would still be visible above the map voting window.
Also, gamemodes other than CTF will use post game map voting but will not use `GameStateBanners.as`, so the minimap would be visible on top of the map voting when a TTH match ends.

So I edited `PostGameMapVotes.as`.
When the map voting window is rendered, then `rules.minimap = false` will execute.
When "voteEndTag" is passed in onCommand() in `PostGameMapVotes.as` and therefore a new map is going to load, then `rules.minimap = true` executes. So even though `GameStateBanners.as` is absent or not executed, the minimap will still be set to true when the map voting ends.

## Steps to Test or Reproduce

Have your up to date CTF or TTH server set up with lots of files changed that allow you to instantly start and pass votes and send chat commands that only mods are able to use normally.

Use chat command `!endgame` or have a "Next map" vote pass. Notice an exception in the console and game banners not showing up anymore. **After this PR , this will not happen. Also notice minimap being disabled when the map voting starts and being enabled again when the map voting ends.**

### DIFF
--- a/Rules/CTF/Scripts/CTF_Banners.as
+++ b/Rules/CTF/Scripts/CTF_Banners.as
@@ -69,6 +69,10 @@ void onTick(CRules@ this)
 	if (this.get_bool("Draw Banner"))
 	{
 		u8 banner_type = this.get_u8("Animate Banner");
+		
+		if (banner_type >= banners.length)	
+			return;
+		
 		Banner@ banner = banners[banner_type];
 
 		if (banner_type == BannerType::GAME_START)
@@ -109,6 +113,9 @@ void onRender(CRules@ this)
 
 				bannerPos = Vec2f_lerp(bannerStart, bannerDest, frameTime);
 			}
+
+			if (banner_type >= banners.length)
+				return;
 
 			Banner@ banner = banners[banner_type];
 
--- a/Rules/CommonScripts/PostGameMapVotes.as
+++ b/Rules/CommonScripts/PostGameMapVotes.as
@@ -233,6 +233,7 @@ void onCommand(CRules@ this, u8 cmd, CBitStream@ params)
 			case 3:	LoadMap(mvm.button3.filename); break;
 			default: LoadNextMap(); break;
 		}
+		this.minimap = true;
 	}
 }
 
@@ -252,4 +253,6 @@ void RenderRaw(int id)
 	Render::SetBackfaceCull(true);
 	Render::SetZBuffer(false, false);
 	mvm.Render();
+	
+	getRules().minimap = false;
 }


mugg91:
This PR has only been checked for normal CTF.

Because it seems Small CTF will show minimap on top of map voting after a match ends normally, this PR should be tested with Small CTF.

AsuMagic:
> This PR fixes an exception that could happen when using chat command `!endgame` or when "Next map" vote passes, in a CTF server, with the newly introduced game state banners (See PR #1166 which introduced GameStateBanners.as and CTF_Banners.as).
> 
> Variable `banner_type` could become equal to the length of the `banners` array which causes an out of bounds exception. I made it so it will return from the function if that's the case and when `banners[banner_type]` is about to be used.

OK, good catch. But the fix to add such guards is a bit peculiar to me.  
Upon a quick glance at what actually sets the `"Draw Banner"` property, it seems that the logic for defining the `"Animate Banner"` property (which defines the kind of banner to use) boils down to:

```angelscript
		if (state == GAME_OVER && this.getTeamWon() >= 0) // <--- sussy
		{
			this.set_u8("Animate Banner", BannerType::GAME_END);
			this.minimap = false;
		}
		if (state == WARMUP || state == INTERMISSION)
		{
			this.set_u8("Animate Banner", BannerType::WARMUP_START);
		}
		if (state == GAME)
		{
			this.set_u8("Animate Banner", BannerType::GAME_START);
		}
```

`"Draw Banner"` is set unconditionally, but `state == GAME_OVER && this.getTeamWon() == 0` means that `"Animate Banner"` will never be set. Hence this error. So I see two proper fixes to this:

- Either we don't draw a banner at all -- which is what your PR does -- but instead not set `"Draw Banner"` in that corner case.
- Or we make a proper variant of the banner when "the map was skipped" (either by vote or `!endgame`).

AsuMagic (approved):
This solves the issue for now though so "let's just fix it properly later"

PR #1348, opened 2022-08-03
[fixed] Saw killing you when changing class or when using tunnel

### LINKED ISSUE #1252: Saw kills you when changing class

### LINKED ISSUE #1540: Christmas Presents can't spawn at Saws
## Current in-game behaviour.

During the Christmas Event, Presents will spawn at Pine Trees and float from the sky at a rare chance, when the "ho ho ho" sound plays. But when there is a Saw at the Pine Tree, the Present will get Sawed.
I found this behavior odd, so I'm raising this issue.

## Proposed Change

I guess the best solution is to prevent Presents from getting Sawed right after spawning by adding a check to Saw in `canSaw()`:
```
if (blob.getName() == "present" && blob.getTickSinceCreated() < 30) 
return;
```

I already did changes to `canSaw()` in PR #1348. So I could add the above check, too.
Just asking for opinions for now, though.

Alternative possible solutions:
- Don't spawn present if a Saw is near.
- Spawn presents in different places, not at Pine Trees.

mugg91:
## Status

## Description

Fixes #1540 

Christmas Presents spawning at Pine Tree during the Christmas Event will not get sawed immediately.
Only blobs that have tag "can be sawed immediately" will get sawed right away - only Log has this Tag for now.

-- Older description follows: --

Fixes #1252.

When travelling via tunnel, apply 
`caller.setPosition(Vec2f(tunnel.getPosition().x, tunnel.getPosition().y + getMap().tilesize));` 
instead of `caller.setPosition(tunnel.getPosition());`.
This causes you to be placed slightly below Saw level, so a Saw that stands at a tunnel will not kill you.
In [this image](https://imgur.com/jnw8Ink), a Saw in the third picture will not kill you but a Saw like in the fourth and fifth picture, will kill you.
This is done for logical reasons.
A character who travels to a tunnel like in the 3rd picture should not be falling and a Saw that stands at ground level should not kill any traveller.
The traveller has to actually fall in a Saw to get killed.

Saw will only kill blobs that are created not less than 30 ticks ago.
`blob.getTickSinceCreated() < 30` is applied.
Therefore, if you change class while standing inside a Saw or while holding a Saw, you will not get killed by it after having changed class.

I have tried fixing this issue before.
After adding a check to Saw so only blobs that have a certain downward velocity are killed, some success has been achieved but devs were not satisfied by this. See #1257 
I have also tried applying an immunity span after travelling via tunnel or changing class, but this posed another bag of problems.
A player could spam respawning to become invincible in a TDM_Spawn, a tunnel traveller could not be hurt by attacks when he actually should, and it was not garantueed that the character doesn't get killed by a Saw due to high ping.

## Steps to Test or Reproduce

See the image above and test picture 1, 3, 4 and 5.

## Additional 

Please note that the "tunnel travel" change and the "change class" change, that are both included in this PR, are separate from each other.
If people think that placing a Saw at an enemy tunnel as a tactic should be kept in the game, then that change can be left out without a problem.
I just think it has to take a bit more effort to block an enemy tunnel and therefore added this change.
Please let me know if you want me to change it back.

### DIFF
--- a/Entities/Industry/Saw/Saw.as
+++ b/Entities/Industry/Saw/Saw.as
@@ -152,8 +152,11 @@ void Blend(CBlob@ this, CBlob@ tobeblended)
 
 bool canSaw(CBlob@ this, CBlob@ blob)
 {
-	if (blob.getRadius() >= this.getRadius() * 0.99f || blob.getShape().isStatic() ||
-	        blob.hasTag("sawed") || blob.hasTag("invincible"))
+	if (blob.getRadius() >= this.getRadius() * 0.99f 
+		|| blob.getShape().isStatic() 
+		|| blob.hasTag("sawed") 
+		|| blob.hasTag("invincible")
+		|| blob.getTickSinceCreated() < 30)
 	{
 		return false;
 	}
--- a/Entities/Industry/Saw/Saw.as
+++ b/Entities/Industry/Saw/Saw.as
@@ -156,7 +156,7 @@ bool canSaw(CBlob@ this, CBlob@ blob)
 		|| blob.getShape().isStatic() 
 		|| blob.hasTag("sawed") 
 		|| blob.hasTag("invincible")
-		|| blob.getTickSinceCreated() < 30)
+		|| (blob.hasTag("player") && blob.getTickSinceCreated() < 30)) // fix for getting insta-killed after changing class
 	{
 		return false;
 	}
--- a/Entities/Industry/Saw/Saw.as
+++ b/Entities/Industry/Saw/Saw.as
@@ -156,7 +156,7 @@ bool canSaw(CBlob@ this, CBlob@ blob)
 		|| blob.getShape().isStatic() 
 		|| blob.hasTag("sawed") 
 		|| blob.hasTag("invincible")
-		|| (blob.hasTag("player") && blob.getTickSinceCreated() < 30)) // fix for getting insta-killed after changing class
+		|| (!blob.hasTag("can be sawed immediately") && blob.getTickSinceCreated() < 30)) // fix for getting insta-killed after changing class
 	{
 		return false;
 	}
--- a/Entities/Industry/Tunnel/TunnelTravel.as
+++ b/Entities/Industry/Tunnel/TunnelTravel.as
@@ -149,11 +149,21 @@ void Travel(CBlob@ this, CBlob@ caller, CBlob@ tunnel)
 		//dont travel if caller is attached to something (e.g. siege)
 		if (caller.isAttached())
 			return;
+		
+		// assume destination is center bottom
+		Vec2f position = tunnel.getPosition();
+		position = Vec2f(position.x, position.y + tunnel.getHeight() / 2 - caller.getHeight() / 2);
+		
+		// apply offset (if it exists)
+		if (tunnel.exists("travel offset"))
+		{
+			Vec2f offset = tunnel.get_Vec2f("travel offset");
+			position += tunnel.isFacingLeft() ? -offset : offset;
+		}
 
 		// move caller
-		caller.setPosition(Vec2f(tunnel.getPosition().x, tunnel.getPosition().y + getMap().tilesize));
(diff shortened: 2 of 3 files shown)


mugg91:
In retrospective, the solution I came up with is probably not good because wood logs that come from trees won't get sawed right away.
`blob.getTickSinceCreated() < 30` has to be changed to `(hasTag("player") && blob.getTickSinceCreated() < 30)`.

mugg91:
I applied the fix that I described in my previous post.
The PR is now truly ready.

mugg91:
Presents spawning at Pine Tree "Tree Farms" during Christmas Event will get Sawed.
If they didn't get Sawed and you open the Present while there is a Saw near, the contents (e.g. Drill) could get Sawed.

We only want logs to get Sawed right away, correct?
So perhaps we should change the `(blob.(hasTag("player") && blob.getTickSinceCreated() < 30)` check that was implemented in this PR to
`(blob.getName() != "log" && blob.getTickSinceCreated() < 30)`.
So nothing gets Sawed right after spawning except wood logs.

mugg91:
I pushed a commit that applies these changes:
- Only saw blobs that are older than 30 ticks, except if they have the Tag "can be sawed immediately". Currently this Tag is added to Log, only. Prevents Players from getting sawed after changing class at tent and prevents Presents that spawn at Pine Trees during the Christmas Event to get sawed. Prevents Drill coming out of Presents to get sawed.
- Make tunnel traveler get positioned at the tunnel's bottom, works now for outposts, too. Resolves that files were conflicting in this PR. Instead of blindly placing the traveler a tilesize deeper, it will fetch the tunnel's position, add 1/2 of tunnel's height, subtract 1/2 of traveler's height. The outpost's x-offset will still be applied.

PR is ready once more.

Entities/Industry/Saw/Saw.as
```diff
@@ -152,8 +152,11 @@ void Blend(CBlob@ this, CBlob@ tobeblended)
 
 bool canSaw(CBlob@ this, CBlob@ blob)
 {
-	if (blob.getRadius() >= this.getRadius() * 0.99f || blob.getShape().isStatic() ||
-	        blob.hasTag("sawed") || blob.hasTag("invincible"))
+	if (blob.getRadius() >= this.getRadius() * 0.99f 
+		|| blob.getShape().isStatic() 
+		|| blob.hasTag("sawed") 
+		|| blob.hasTag("invincible")
+		|| (!blob.hasTag("can be sawed immediately") && blob.getTickSinceCreated() < 30)) // fix for getting insta-killed after changing class
```
AsuMagic, line 159: 30 ticks seems too generous as it would be possible to (accidentally) abuse this by coming into contact with a saw shortly after switching classes (could reasonably happen even after TPing into a saw). I'm not sure if 5 fully solves the problem the PR aims to resolve (e.g. due to lag), but let's try. FWIW, the fact that you can get sawed from class swap is more of an engine bug than anything.

```suggestion
		|| (!blob.hasTag("can be sawed immediately") && blob.getTickSinceCreated() < 5)) // fix for getting insta-killed after changing class
```

PR #1349, opened 2022-08-03
[changed] Invincibility span code clean-up and changes

mugg91:
## Status

## Description

I tried finding a fix for #1252, but abandoned the idea of using invincibility spans.
I still did some cleaning up and other changes that may be interesting, hence this PR.

As a result of rewriting code, I renamed 
`IMMUNITY_SECS` to `IMMUNITY_SEC`,
`"spawn immunity time"` to `"immunity ticks"`
and `immune` to `isImmune`.

**UnSpawnImmunity.as:**

- Instead of saving game time stamps, I save the immunity ticks directly to the blob and deplete it every tick.

- Instead of using a modifier that determines a maximum invincibility span, I make it deplete by 2 if action1 is pressed.

- These checks below have been removed because it doesn't make sense they are in onInit() and appear to be unused.
A newly created blob can't have any tags or set values.
```
if (this.hasTag("invincibility done"))
	{
		return;
	}
```
```
if (!this.exists("spawn immunity time"))
```
As a consequence, removed Tag "invincibility done".
It still works the same way as before when testing offline and online.

Added a check to onTick() instead:
```
if (!this.hasTag("invincible"))
	{
		return;
	}
```

- Still kept `this.getCurrentScript().runFlags |= Script::remove_after_this;` for now. 
But if needed in the future, you could add invincibility span to the character on the go.
Then this line would have to be removed and the script keeps checking if any invincibility has to be handled.

- Updated commenting that explains the code.

## Steps to Test or Reproduce

Go to Sandbox.
Suicide or respawn at a tent.
Notice everything is as before.
If you hold action1, notice the invincibility getting lost faster. (Not the same way as before. Now every keypress matters.)

### DIFF
--- a/Entities/Characters/Scripts/UnSpawnImmunity.as
+++ b/Entities/Characters/Scripts/UnSpawnImmunity.as
@@ -1,25 +1,27 @@
 void onInit(CBlob@ this)
 {
-	if (this.hasTag("invincibility done"))
-	{
-		return;
-	}
 	this.Tag("invincible");
-
-	if (!this.exists("spawn immunity time"))
-		this.set_u32("spawn immunity time", getGameTime());
+	this.set_s32("immunity ticks", getRules().get_f32("immunity sec") * getTicksASecond());
 }
 
 void onTick(CBlob@ this)
 {
-	bool immunity = false;
-
-	float time_modifier = (this.isKeyPressed(key_action1) ? 0.75f : 1.0f);
-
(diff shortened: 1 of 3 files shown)


mugg91:
Ok, it has appeared to me right after opening this PR that it should be depleting the span based on ticks per second, not a hard value of 1 or 2. Will change later.

PR #1367, opened 2022-08-31
[changed] Egg breaks upon impact, makes a crack sound and can be used as a projectile

mugg91:
## Status

## Description

[changed] Egg breaks on impact with walls and enemy players
[changed] Egg thrown against enemy players will stun them for 10 ticks
[changed] Egg gib count from 6 to 4
[added] Egg crack sound

This PR makes it so Egg will break if hitting the ground or a wall too hard. 
`if (this.getShape().vellen <= 6.8f) return;` in `onCollision()` is used which is equivalent to the check that causes the water splashing when a full Bucket hits the ground or a wall.

Egg can be thrown against an enemy to mini-stun them for 10 ticks.
This works if the Egg's velocity is higher than `6.8f`.
The Egg will break and not heal the enemy.
The enemy is set to dazzled so the screen flashes briefly which fits the effect of getting an Egg thrown at.
Although `this.server_Hit(blob, pos, Vec2f(), 0.0f, Hitters::water_stun_force, true);` could be used to create dazzle animation, screen flash, stun and ArgShort sound, it was not used because I cannot set the stun to 10 ticks and Sponge stun reduction might take effect unwantedly.

A crack sound has been added which plays upon hatching, hitting the ground or hitting an enemy.
It does not play when healing oneself or when Egg collides with Fireplace to create a Cake.

This has been tested offline and online and seems to work as intended.
The `isServer()` checks have been placed carefully to make sure the behavior is the same offline and online.

## Steps to Test or Reproduce

Go to offline or online Sandbox.
Throw an Egg down on the ground or run forward and throw it against a wall.
Notice Egg will break if hitting too hard.
Notice the Egg's crack sound.

Spawn an enemy builder, hit him once.
Run forward and throw an Egg against an enemy player.
Notice Egg will break and mini-stun the enemy player without healing him.
Notice the Egg's crack sound.

Place an Egg and wait for a chicken to hatch.
Notice the Egg's crack sound.

Place an Egg somewhere and hurt yourself, then pick up the Egg.
Notice the behavior is the same as before (unchanged).

Throw an Egg against a Fireplace.
Notice the behavior is the same as before (unchanged).

Place a Trampoline on a field with 2 empty spaces.
Have your Eggs bounce on it.
Change team to something else.
Go where the bouncing Eggs are.
Notice you will get mini-stunned.


Entities/Natural/Animals/Chicken/Egg.as
```diff
@@ -38,22 +39,54 @@ void onTick(CBlob@ this)
...
+	{
+		BreakEgg(this);
+	}
+	
+	// throw egg at enemies to mini-stun them
+	if 	(blob !is null 
+		&& blob.hasTag("player")
+		&& this.getTeamNum() != blob.getTeamNum())
+	{
+		this.Tag("egg projectile");
```
Gingerbeard5773, line 71:
```suggestion
		this.Tag("healed");
```
Another tag already exists for the same purpose. Lets use that one instead.

Entities/Items/Food/Eatable.as
```diff
@@ -74,7 +74,9 @@ void onCollision(CBlob@ this, CBlob@ blob, bool solid)
 		return;
 	}
 
-	if (getNet().isServer() && !blob.hasTag("dead"))
+	if (getNet().isServer() 
+	&& !blob.hasTag("dead")
+	&& !this.hasTag("egg projectile"))
```
Gingerbeard5773, line 79:
```suggestion
	if (isServer() && !blob.hasTag("dead"))
```
adding the egg projectile check here is unneeded if other suggestion is applied.

Gingerbeard5773 (changes requested):
https://user-images.githubusercontent.com/68350259/232258094-62e8d732-6dc7-4c86-b556-408f893e65a8.mp4

The way this PR has been implemented makes it susceptible to desync problems like the one shown in this video (this is because collisions in kag aren't synced). it may be rare but in gameplay it will be noticed.

mugg91:
Yeah, not sure where I was going with the Egg projectile thing.
If at least the crack noise could make it into the game, I'm happy.
Lost interest in coding for now, so I'm asking if someone else can finish it.

PR #1369, opened 2022-09-04
Client side block selection

laws65:
## Status

## Description
When selecting blocks, it takes 1 round trip for the client's newly selected block to be updated. In high ping, this sort of thing can happen:
-Player has wood block selected
-Player switches to stone block
-Player tries to place block
-Server receives command to switch to stone
-Server receives command to place block
-Client receives command to switch to stone block
-Client receives command to build wood block

Because the game determines what to build from the outdated client info, you end up building the wrong block. This is fixed by reading the current server values of what block a player has selected to build.

To do this I've added a new script, Placement.as, which sends out a build command when the player left clicks. Then the two placement scripts, BlockPlacement.as and BlobPlacement.as, react to the command by seeing if the player has a block/blob selected and if so then builds it. This PR also adds and fixes lots of the server side checks to see if a player is building in an allowed location. 

I'm still working on it because the code is still awful and there's a bug related to showing missing requirements. 

## Steps to Test or Reproduce

- Add 300ms ping using clumsy (http://jagt.github.io/clumsy/)
- Start a dedicated server and join it
- Place blocks while quickly switching between them


Entities/Common/Building/BlockPlacement.as
```diff
@@ -282,12 +171,85 @@ void onRender(CSprite@ this)
...
+bool serverTileCheck(CBlob@ blob, u8 tileIndex, Vec2f cursorPos)
+{
+	CBlob @carryBlob = blob.getCarriedBlob();
+	if (carryBlob !is null)
+	{
+		return false;
+	}
+
+	// Make sure we actually have support at our cursor pos
+    CMap@ map = getMap();
```
Vam-Jam, line 229: double check formatting on github, theres multiple cases of wrong indentation

Vam-Jam (changes requested):
Might be worth calling genericPlaceCheck client side, just so client will get feedback on the off chance they somehow cause it to return false

PR #1370, opened 2022-09-05
GoldBuilding.as refactoring (updated)

Pirate-Rob:
## Status

- **READY**: (unless I've done this PR wrong somehow, but it should all be good)

## Description

Fixes Mazey's GoldBuilding.as refactor/removal by making it compatible with the inclusion of outpost and the removal of ballista's gold cost.

## Steps to Test

Creating an outpost then destroys it gives gold.
Destroying a ballista does not give gold.
Destroying all other gold cost buildings still gives gold as expected.
Destroying outpost crate gives gold.
Destroying other crates does not give gold.


Vam-Jam:
IMO:
- Move GoldBuilding.as to a new folder called old/modcompatability instead of removing it.
- Rework `InitCosts();`, it reads a cfg thats already been initalized on every blob creation, it'll suck for people who have bad hard drives and just adds more boilerplate

Mazey:
@Vam-Jam obtaining the old files is so easy for KAG I'm not a huge fan of old unused files lingering around in the base game.

I think InitCosts would be a separate optimisation. This PR is mostly about removing legacy code, making sure all gold blobs behave the same & reduce gamemode-specific code in blobs. Would you approve it, as you did with the original PR?

Vam-Jam:
There's plenty of old unused files lingering around the base game. Gold building is used by any mod which has any sort of custom shop cfg, which is a lot of them. I normally would not bat an eye if it was a smaller script that's used for one specific thing, but this will break a lot mods. 

Pushing the file into base/old folder and keeping it about for a while isn't partially hard for us to do. If we had some sort of deprecation warning, it would work well.

The optimization can be a separate pr, I was more just pointing out the fact it's an issue that will cause issues for a few players (seen how a lot of kag players are running on potatoes). It's not a blocking the pr from being accepted.

PR #1371, opened 2022-09-10
[added] new component Flamer

mugg91:
## Status

- **READY**
Developed and tested in offline.
Tested briefly in online, no observed regressions or bugs.

## Description

Please note: PR #1374 has to be applied so the player can take damage even when fire hits are applied repeatedly in succession.

This PR adds a new interesting component tile "Flamer" that can be placed by a Builder in Sandbox.

- Requires 50 Stone to build.
- Has a stone background so as to not cause burning to itself indirectly.
- Will continuously spawn flame particles.
- Will burn wood blocks, grass or other flammable tiles.
- If the fire is emitting, will also emit a sound effect and a small light.
- Will ignite unlit fireplaces.
- To apply fire to other blobs, a radius check has been added. 
This seemed like the best approach to simulate a hurtbox.
A different approach that added a function to particles was neglected.
Within radius 7.5f of the tile it is facing, will ignite arrows.
Within radius 3.7f of the tile it is facing, will ignite unlit fireplaces, cook food and apply fire damage to blobs.
The applied fire damage only lasts 75% of the duration that the fire state normally does so touching Flamer fire is not too annoying.
- Will not emit fire for 30 ticks if hit by water splash.
- Will not emit fire if the exit is blocked or if the exit is underwater.

`ArrowCommon.as` has been added so Flamer can ignite Arrows.
`FireplaceCommon.as` has been added so Flamer can ignite Fireplaces.

Initially, a tick frequency of 3 was used but was changed to 1 to increase consistency of igniting Arrows.

I opted against the name being "Flamethrower" because Flamethrower sounds more like an item and Flamer in google images looks a lot like what I wanted the component to be like.
The sprite is a mockup by me. 
I don't remember where I got the fire sound, but it was most likely from here somewhere and edited by me: 
https://pixabay.com/sound-effects/search/fire/
If requested by the development team, I can change the sprite, sound or name.

## Steps to Test or Reproduce
Go to Sandbox and go to the devices tab, select Flamer and place it.
Place a source such as a lever, activate it.
Notice flames coming out.
Play around with it :-)

## Inspiration
There is not a direct inspiration but I kept thinking of these.
https://www.youtube.com/watch?v=mafCrNI6P70&t=25888s
https://www.youtube.com/watch?v=YAwXye_HlE4&t=365s


mugg91:
Showcase video - https://www.youtube.com/watch?v=VFT1vhhkjP0

mugg91:
I changed the "is something inside the flame" logic so that getBlobsInRadius is called only once instead of twice.
First, getBlobsInRadius is used to fetch blobs that are within 7.5f of the center of the tile that the Flamer tile is facing.
It checks if the blob is a normal arrow and if yes, ignites it.
If not, it checks if the blob is within 3.7f. ( https://i.imgur.com/Lt8SWd5.png If the distance is below zero, it applies stuff to the blob: ignite fireplace, cook food, apply damage).
This should be smoother code-wise and performance-wise.

Also I added a tag "activated" so Flamer only ticks if it has that tag.
In case the Flamer is underwater but is powered, it will still be ticking, though.

mugg91:
I made a different PR #1374 that fixes the issue of player not getting damaged when fire hits keep happening repeatedly.
That PR must be applied in order for this PR to work.
Setting the status of this PR to **Ready**.

mugg91:
It may be more sane and less cluttery to make use of `map.getHitInfosFromArc` or `map.getHitInfosFromRay` to burn tiles and blobs. The underwater and "is exit blocked" check could still be the same as is.
Waiting for dev to give thoughts.

mugg91:
I will make a new PR of this

PR #1373, opened 2022-09-12
[added] new building "Bank Deposit" to CTF

mugg91:
## Status

Tested offline and online, works as intended.
Except for these remaining issues:
- Withdrawing money so that you can't withdraw anymore will sometimes play sound "NoAmmo" even when it shouldn't (In Online only).

## Description

This PR adds a new building Bank which is basicly a storage for your coins.

- Holds up to 1000 coins. You can deposit and withdraw as much as you want in steps of 20, 100 or all coins.
- If destroyed, drops all deposited coins.
- Has a sprite layer depending how much money is stored.
- Can place one Bank per team.

Showcase video https://www.youtube.com/watch?v=oaIli086Pko

This allows your team to withdraw from or deposit to a shared pool of coins, i.e. allowing your team to share coins with one another. So if two players have 60 coins, they can both deposit their coins so that a player can then withdraw 120 coins to buy a Keg. Or if you have many coins on hand but want to make a risky play that will get you killed, you can deposit your coins so you won't lose all your money.

To make it not too strong, it is limited to 1000 coins.

The potential risk is that the enemy team can destroy your Bank and get your cash.
So there is an incentive as well as a risk for building it, and an incentive for destroying the enemy Bank.

Sprites are mockups by me. If requested by the dev team, changes can be made.

## Ideas

- Ability to send your coins to a specific player directly.
- Everyone having their own bank account.


mugg91:
Fixed some of the remaining issues:
- Menu icons and coins spritelayer in the Bank now aren't always blue-colored but team-colored instead.
- Use interaction icon 25 (small coin) instead of 8 (tunnel travel).
- Fixed that the menu didn't open if there was more than one Bank.
- You can withdraw until you have 32765 (the game's coin limit) instead of 1000.
- Use description "Withdraw all coins" instead of "Withdraw all your coins".
- Changed the Bank's sprite so you know what team it is from.

PR #1386, opened 2022-10-03
[fixed] Fireplace now properly plays SparkleShort.ogg when cooking, while playing online

### LINKED ISSUE #1383: cooking food on fireplace doesn't make a sound (online only)
## Description 

When egg, grain, fishy, etc. touches fireplace, it will turn into its cooked counterpart and a sound effect should play.
But in online, it doesn't play a sound.

`food.getSprite().PlaySound("SparkleShort.ogg");` in MakeFood.as isn't executed.

mugg91:
## Status

## Description

Fixes #1383.

Throwing grain, egg, steak, fishy etc. at fireplace will now play a `SparkleShort.ogg` sound when playing online.

The blob can only be created server-side, so trying to play a sound based on that blob will not play the sound on the client-side when playing online.
So the sound is now played from the ingredient blob instead.

## Steps to Test or Reproduce

Go to online Sandbox, throw egg at a fireplace.
Notice it makes a sound now.

### DIFF
--- a/Entities/Items/Food/MakeFood.as
+++ b/Entities/Items/Food/MakeFood.as
@@ -1,6 +1,6 @@
 CBlob@ server_MakeFood(Vec2f atpos, const string &in name, const u8 spriteIndex)
 {
-	if (!getNet().isServer()) { return null; }
+	if (!isServer())  return null; 
 
 	CBlob@ blob = server_CreateBlobNoInit("food");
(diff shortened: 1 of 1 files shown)


Gingerbeard5773 (approved):
Works as intended

AsuMagic (approved):

Issue #1387, opened 2022-10-04
Arrow doesn't apply any push force when hitting a blob that had SetStatic(true) and then SetStatic(false) applied


mugg91:
## Description

A blob that had `blob.getShape().SetStatic(true)` applied in one tick, and then `blob.getShape().SetStatic(false)` in a subsequent tick cannot be pushed by an arrow.

## Reproduction

Go to offline Sandbox.
Type `!food` in chat.
Type `getBlobByName("food").getShape().SetStatic(true);` in console.
Type `getBlobByName("food").getShape().SetStatic(false);` in console.
Shoot a charged arrow at the food.
Notice it doesn't react to the arrow at all.

mugg91:
If you spawn a grain and apply 
`getBlobByName("grain").getShape().SetStatic(true);`
and then
`getBlobByName("grain").getShape().SetStatic(false);`
and then
`getBlobByName("grain").AddForce(Vec2f(0, -50));`, 
the grain will not fly up in the air.
But if you pick up the grain and place it back down OR if you never did the SetStatic stuff, then it will fly up in the air.

mugg91:
Adding on to what I described above.
I know that picking up and throwing the item will fix the static bug so AddForce() will take effect.
So I'm finding out what in the pickup and throw code is causing the fix.
In this post I list the code lines that don't fix it, see c).

a) `getBlobByName("bucket").getShape().SetStatic(true);`
b) `getBlobByName("bucket").getShape().SetStatic(false);`
c) I tried here:
`getBlobByName("bucket").getShape().UpdateBody();`
`getBlobByName("bucket").setVelocity(Vec2f(10,10));`
`getBlobByName("bucket").server_DetachAll();`
`getBlobByName("bucket").server_DetachFromAll();`
but it doesn't fix it.
d)  `getBlobByName("bucket").AddForce(Vec2f(20,20)); // no effect`

Issue #1380, opened 2022-09-25
Make flora grow faster when hit by water splashes


mugg91:
## Describe the feature

Splashes by bucket, water bomb or water arrow should speed up growth of grass, flowers and trees.

ewetcr:
I would remove it from trees, it'd be a slight buff to treefarms, and they're already giving too much wood in current balance. In other words, builders carrying 1k wood more often is not a good idea.

Issue #719, opened 2019-10-29
please add more things for map making


Shadowblitz16:
**Describe the feature**
I was just wondering what people think about adding new content to the game for more diverse maps

for example..
- more liquids
  - lava (like water but sets fire to entities and tiles and produces light)
  - quicksand (like water but you can't swim in it only sink and bomb jump out)
  - tar (just quicksand recolor)

- more tiles
  - black background tiles (for cave maps, basically a dark version of air)
  - sand tiles (for dessert maps)
  - snow tiles (for snow maps)
  - ice (easy to break, slippery, for snow maps)

- more fauna
  - coyote (for sand and snow maps)

- more flora
  - vines (for decoration, easy to cut)
  - dead bushes (for snow and dessert maps)
  - dead trees (for snow and dessert maps)
  - dead vines (for snow and dessert maps)

- weather
   - rain (puts out fires)
   - snowstorm  (reduces visability, creates snow tiles)
  -  sandstorm (reduces visability, blows things towards the edge of map)
  - windstrom (blows things towards the edge of map)

**Additional context**
this all can be added with a mod and texturepack but why? kag maps all have the same style and I think bringing more diversity in would not only benefit map makers but also make kag a little more enjoyable too.

Vam-Jam:
Adding more tiles are fairly easy to add, and doesnt require much work

Along with fauna and flora, they would be blobs, you just need to get somebody to sprite it and it can be done.

Liquids would require an engine change, currently kag is only really made for water, and afaik we cant easily add new liquids in right now. However moving water to script side could be possible, and it would allow anybody to make more liquids easily.

https://github.com/1bardesign/kag-flowing-water

This was done by geti and goldenguy, somebody could finish it off when they have the time and see if its worth adding into the base game

Shadowblitz16:
thanks for the reply.
ya the only problem with the black background tiles is detecting what tiles break into, it, or air 

also I knew liquids were engine side although what would be better then have liquids be script side is  do it half way. basically liquids would be handled engine side but they would be expandable and definable script side.

Panky123:
Different sprites for flowers fauna ect is cool. Liquid additions and dark tiles are alot more game breaking imo.

Shadowblitz16:
@Panky123 the liquids would add diversity to game play.
they can be balanced later as seen fit

Harrison-Miller:
Reopening because I'm on the fence about some of this being to mod like and some of this maybe something that could be in kag.

[...]

Issue #452, opened 2019-03-11
Trees burn players


MatheusGodoi:
Right now, when trees catch fire, only the ground of the tree burns, and it don't spread to the rest of the tree until it falls down, so if a archer is standing on the tree while it burns he is completely safe of the fire damage, even if he stays on the tree spot when it dies, the fire will be over so he will take no damage at all.

The proposed change is to make the entire tree catch fire, making players on it burn, so you can't safely sit on top of a burning tree.

I'm not sure about making the tree spread fire to other blocks/players (wood blocks/backwalls, workshops, grass, etc) while its burning/when it dies and fall on the ground, as it was already discussed on this [feature pitch](https://github.com/transhumandesign/kag-base/issues/195).

PR #1389, opened 2022-10-05
[added] Apple Tree

mugg91:
## Status

See "things left to do" below.

## Description

[added] Apple item; an eatable item that heals 1 heart.
[added] Apple tree; a bushy tree recolor with reduced height; spawns apples over time.
[changed] Trees subfolder is better organized. Each tree has its own image file.
[changed] chat command: !tree -> !seed_pine
[changed] chat command: !btree -> !seed_bushy
[added] chat command: !seed_apple
[changed] onTick() of TreeSync.as moved to each tree's own script, changed tick frequency from 1 to 60; might improve performance.
[changed] Seed changes sprite to index 0 on pick up immediately, not delayed.
[fixed] Trees spawned in via chat command will not get stuck below the ground.
[fixed] Pine tree only having side leaves on the right side. Now it has side leaves on both sides.
[fixed] Make tree not have black line between early segments.
[fixed] Fireplace not making a sound when cooking food in online.

Note: For convenience, PR #1386 is included in this PR.

Addresses issue #719 by adding new flora.

Please note: There is a PR #1393 that improves upon this PR and adds leaf wiggling.

----

This PR adds a new kind of tree, the apple tree.
For now it is a recolor of bushy tree, but it can be changed if needed.
It can grow to medium size and will spawn up to 5 apples over time if enough space is available.
To prevent abuse in Sandbox mode, no more than 40 apples can spawn per map.

Apples will spawn as apple flowers first and grow to full size after some time.
If apples are spawned in via chat command, they are full sized already.
Full size apples will behave as other foods, healing the user if low on health. 
It plays an "apple bite" sound and heals for 1 heart.
If a full size apple touches fireplace, it turns into a cake.
If an apple hits the ground hard, it makes a thud sound.

When apples have grown to full size on the tree, they can be picked if you overlap with them, otherwise can be picked from afar.
I opted against apples getting affected by Sword, Arrow or Saw, so it is in line with how Grain behaves, also being invincible to those damage sources. Apples on the tree should not prevent combat and apples should not fall from the tree that easily.
However, when an apple is not hanging on the tree anymore, it does get affected by Arrow shots to be in line with how other foods can be shot. 

(For the time being, apples that fell down from the tree cannot be shot by Arrow. 
I tried finding out why this happens for 3 days but ended up concluding this being an engine-side bug with SetStatic(), see #1387.
Otherwise, tested online and offline with no more issues observed.)

When the apple tree falls down, apples that have grown to full size will fall down with a slight AddForce() effect.
I opted against apples being spritelayers or attached to the tree via attachment points. 
It made more sense that apples are blobs to begin with.
Attachment points cannot be removed later and shouldn't be spammed.
Spritelayers cannot be grabbed or cooked (if fireplace touches a hanging apple).

For now, I didn't make any more adjustments to the map loading, so the game will randomly choose if a tree "pixel" will become a pine tree, bushy tree or apple tree. To prevent unfairness (e.g. one team having the apple tree, the other doesn't), the logic should be updated so only 1 or 2 apple trees are spawned somewhere in the center of the map.
Alternatively, apple tree could have its own tree "pixel" but then some map image files would need to be updated.

Sounds taken from here:
- https://pixabay.com/sound-effects/eating-apple-6928/ - bite
- https://www.fesliyanstudios.com/royalty-free-sound-effects-download/falling-down-121 - thud1
- https://www.fesliyanstudios.com/royalty-free-sound-effects-download/book-238 - thud2
- https://pixabay.com/sound-effects/body-falling-to-ground-100474/ - thud3

Minimap icons for trees could fit with 8x16 icons instead of 8x32, but I can't change the render logic, as it seems to be engine-side.

----

Lastly, I would like to encourage people to be open to new changes.
I have tried my best to implement a fun and interesting new feature that adds depth to the game and I made considerations to make it not abusable or game-breaking.

In my opinion, apples fit well with medieval topics and play a bigger role in some fields such as mythology and religions.
A new tree shouldn't be "just another plain old tree", but instead bring something new to the table.

## Steps to Test or Reproduce

Go to Sandbox. Experiment with these chat commands:
!apple
!tree_apple
!seed_apple

## Things left to do

- Make apple trees only spawn in middle

I might make a duplicate PR, but with the below changes added:

- ~~Grass, bush and tree leaves are animated and play a sound when you touch them.~~ (Done, see: [#1393](https://github.com/transhumandesign/kag-base/pull/1393))
- Fire climbs up on a tree, player catches fire if inside the tree fire too long. - [#452](https://github.com/transhumandesign/kag-base/issues/452)
- Water splashes grow flora faster. - [#1380](https://github.com/transhumandesign/kag-base/issues/1380)


mugg91:
The Apple Tree can be seen in [this video](https://www.youtube.com/watch?v=25sks1HfPyY) at 0:48.

AsuMagic:
Closing this as it is superseded by #1393

PR #1390, opened 2022-10-06
[added] Despawn Flickering to some blobs

### LINKED ISSUE #1385: Add an indicator for when items are about to despawn
## Current in-game behaviour.

- Heart despawns after 40 seconds.
- Dead fishy despawns after 40 seconds.
- Dead trader gibs after 20 seconds. (despawns after 45 seconds if PR [#1283](https://github.com/transhumandesign/kag-base/pull/1283) is applied)
- Log despawns after `240 + XORRandom(60)` seconds.
- Material Waterbomb despawns after 240 seconds.
- Material Bomb despawns after 300 seconds.
- Material Arrow despawns after 45 seconds.
- Material Water Arrow despawns after 180 seconds.
- Material Fire Arrow despawns after 180 seconds.
- Material Bomb Arrow despawns after 300 seconds.
- ...

Items disappear without prior warning.

## Proposed Change

This change request would like items to show an indicator when they are about to disappear.
Probably the best way is to make it start blinking at 5 seconds left and make it quicker when the remaining time gets shorter.

A different way to do it would be to add a seconds counter above the item that counts down the time (5...4...3...2...1...poof).
However, this may add too much clutter on the screen.

It should be considered if collided arrows should blink, too.

## Optional

As was previously suggested by someone, don't make items despawn when in an inventory blob.
This should not be a problem, because if players actually try to hoard hearts and materials, such play should be rewarded.
There is no problem in hoarding corpses.

mugg91:
## Status

## Description

This PR makes it so some blobs that are about to despawn will start to flicker.
The flickering becomes faster the shorter the remaining time is.
If you pick up the blob, the decay time is reset, so there is no "flickering item in hands" and the player is rewarded for picking up an item that is about to expire.

Fixes #1385.

----

```
[added] DespawnFlicker.as - this script causes blobs to flicker visually in the last seconds before despawn.
		This script has been added to:
		Heart, Grain, Steak, Log, Fishy, mat_arrows, mat_ballistabolts, mat_ballistabombbolts, mat_bombs,
		mat_waterbombs, mat_bombarrows, mat_firearrows, mat_waterarrows.
[added] DecayUnused.as to Heart, Log and Fishy - this means those blobs do not despawn if kept in inventory or when attached to player or catapult.
[changed] Grain uses DecayUnused.as instead of TimeOut.as, with a decay time of 60 seconds set to it (same time as was set in TimeOut.as).
[changed] Steak uses DecayUnused.as instead of TimeOut.as, with a decay time of 60 seconds set to it (same time as was set in TimeOut.as).
[changed] Steak will gib if it is destroyed via damage, instead of nothing happening
[changed] Steak will not gib if it despawns via server_SetTimeToDie().
[changed] Fishy will gib if it is destroyed via damage and if its age is greater than 0, instead of nothing happening
[changed] Fishy will not gib if it despawns via server_SetTimeToDie().
[changed] Log will not gib if it despawns via server_SetTimeToDie().
[removed] TimeOut.as - as it did the same thing as DecayUnused.as.
[fixed] Dead Fishy will not gain size.
```
----

These blobs will not flicker as I decided against it for now:
- Corpses
- Flying or collided arrow
- Flying ballista bolt

Materials Wood, Stone and Gold don't have `DecayUnused.as` and `DespawnFlickering.as` added to them since they are never set to despawn at once, but in steps in quantity.

Tested offline and online and no bugs observed.

## Steps to Test or Reproduce

Go to offline or online Sandbox.
`!log`, `!heart`, `!grain`, `!steak`, `!mat_arrows`, `!mat_bombs` etc.
or `!fishy` and have the fishy die.
Then wait until the blob despawns.
Notice it will be flickering in the last 4 seconds.

`!heart`
Pick up the heart, keep it in your hand or place it in an inventory and come again later.
Notice the heart will not have disappeared.

`!steak`
Keep hitting the steak by pickaxe or sword jab.
Notice it will gib now.

`!fishy`
Kill the fishy and keep hitting it.
Notice it will gib now.

`!fishy`
Kill the fishy and wait for a while.
Notice it will not grow to larger size while dead now.


mugg91:
I shall try to make it tick less in an upcoming commit, for performance reasons.

mugg91:
Despawn flickering can be seen in [this video](https://www.youtube.com/watch?v=25sks1HfPyY) at 0:18 and 1:35.

PR #1393, opened 2022-10-08
[added] Apple Tree + Leaf Wiggle

mugg91:
## Status

- **In Development**: this PR is still being worked on for now.

The apple tree is ready.
Leaf wiggling is ready.

In development until these things are done. 
If a dev wants to merge this, contact me and I will work on these faster :-)
- Make apple tree spawn equally on both sides in CTF/TTH.
- Add a check that disables leaf wiggling if too many trees are close to each other or if there are too many trees on the map.
- Apple Tree leaf wiggle should produce olive green plant particles, not grass green.
- No leaf wiggling underwater. (underwater plants would be nice btw.)
- Remove `RenderCheckboxes.as`, not needed.
- Don't apply changes in `BasePNGLoader.as` lines 483-485.

## Description

This PR includes #1389 and applies the following changes:

[added] Leaf spritelayers on trees will shake, make a rustle sound and produce leaf particles, when a player touches them.
[added] Bush will shake and make a rustle sound, when a player touches them.
[changed] Moved common tree sounds and tree scripts to respective subfolders.
[added] Apple tree to test.kaggen (Sandbox). When a bushy tree is spawned, there is a 12.5% chance it will be an apple tree.
[added] When apple tree falls, Apple flowers will gib into plant particles.

At first, when the tree grows via `GrowSprite()`, references to certain spritelayers are saved to the tree blob via `SaveWigglyLeaf()`.

The tick frequency is changed from 60 back to 1, but certain functions are only entered every n-th tick which is done by checking against blob-saved values `"grow check tick frequency"` and `"leaf proximity check tick frequency"` which are initialized in `TreeSync.as`. Therefore `GrowCheck()` is still only entered every 60th tick (every tick if tree is falling) and the new `LeafProximityCheck()` is entered every 10th tick.

`LeafProximityCheck()` will check if a player is near the tree via `getBlobsInBox()` and if yes, changes the blob-saved tick frequency to 1. If not, changes the blob-saved tick frequency to 10. This way, if a blob is inside the tree, the tree keeps track where the player is more often. 
This function checks if the player is inside any previously saved wiggly leaf spritelayer, and if yes, initiates the wiggling by setting `this.set_u8("wiggly leaf duration " + j, 6);`. It will play one of three possible leaf rustle sounds randomly and produce particles. Depending on the `"particle type"` that was saved to the blob in onInit(), the particles will be plant-type (bushy, apple) or grass-type (pine).

Lastly, `ProcessLeafWiggle()` will animate the leaf spritelayer that was touched by changing the y offset, so it looks like it bounces down and up for a bit.

A similar effect was added to Bush. I changed the sprite sheet so there are now 5 possible bushes with 3 frames each.
When the sprite is initiated, an animation is created and the 3 corresponding frames are added.
Still depending on its own Net Id to make sure it looks the same for everyone.
I made sure the animation frame is the same before and after touching a Bush.
Touching a Bush also plays a random "LeafRustle" sound.

I made it so both tree leaf wiggling and bush wiggling only occurs if the player has set "Faster Graphics" off.

The performance on wiggling Bush should be a no-brainer, since there are no checks going unless the player is colliding and `onTick()` only runs as long as the Bush is animated.
The performance on wiggling tree leaves is a bit draining, but due to the proximity check, it is still very good.
You need to stack like 200 trees on top of each other before lag becomes noticeable which is not happening unless you actually spawn them on top of each other in Sandbox. There is also the possibility to improve performance in the future by making the tree blobs disable their own leaf wiggling if too many trees are on the map or close to each other.

Apple flower will gib into plant particles when the apple tree that it hangs on starts falling.

Tested in offline and online with no bugs observed.

The three Leaf rustle sounds are all taken from this one: 
https://pixabay.com/sound-effects/leaves-rustling-14633/

## Steps to Test or Reproduce

Like in the previous PR, experiment with `!apple`, `!tree_apple` and `!seed_apple`.
Plus:

Go to Sandbox offline or online.
Notice there will be an apple tree or two at a rare chance.
If you fell it while it has an apple flower on it, notice the apple flower gibbing.
Jump around inside trees, notice the leaf spritelayers will slightly bounce up and down, play a sound and produce particles.
Walk through a bush, notice it will wiggle and play a sound.

## Additional

Some inspiration from [Smash Bros.](https://www.youtube.com/watch?v=EVa9Hu5aWZM&t=75) that has tree branches that shake up and down.


mugg91:
Leaf wiggling can be seen in [this video](https://www.youtube.com/watch?v=25sks1HfPyY).

PR #1396, opened 2022-10-10
[fixed] corpse making a grunt when falling

### LINKED ISSUE #1382: Falling corpse makes a grunt sound
## Description

Player corpse will make a grunt sound when hitting the ground from a high place.
This commonly happens in online during gameplay.
Does not seem to happen with spawned in runner blobs, but it happens with bots.

## Proposed Change

In `FallDamage.as`, make `Sound::Play("/BreakBone", this.getPosition());` only execute if `this.hasTag("dead")` is false.

## Reproduction

Go to Sandbox, build a high ladder.
Type `AddBot("abc");` in console, place two background tiles and kill it by falling spikes.
Grab the corpse, climb up the ladder and throw it away.
If done properly, notice the corpse will make a grunt sound.

mugg91:
## Status

- **Fixed**
- No known issues.

## Description

Fixes #1382.

A dead player could make a grunt when fallen from a high place.
If you keep throwing the same corpse, instead of the grunt it could make a broken bone sfx.

This PR checks if the player is actually dead by checking for `hasTag("dead")` before playing 
`Sound::Play("/BreakBone", this.getPosition());` or `Sound::Play("/FallDeath.ogg", this.getPosition());`.

## Steps to Test or Reproduce

Go to Sandbox.
Place one ladder on the ground, then have spikes fall on you to kill yourself.
Pick up your own corpse, climb to a high place and throw your corpse to the ground.
If done correctly, should make a grunt sound or a break bone sfx. 
--> **After this PR, it doesn't make any sound.**

### DIFF
--- a/Entities/Common/Movement/FallDamage.as
+++ b/Entities/Common/Movement/FallDamage.as
@@ -57,7 +57,7 @@ void onCollision(CBlob@ this, CBlob@ blob, bool solid, Vec2f normal, Vec2f point
 		// stun on fall
 		const u8 knockdown_time = 12;
 
-		if (doknockdown && setKnocked(this, knockdown_time))
+		if (!this.hasTag("dead") && doknockdown && setKnocked(this, knockdown_time))
 		{
 			if (damage < this.getHealth()) //not dead
 				Sound::Play("/BreakBone", this.getPosition());
--- a/Entities/Common/Movement/FallDamage.as
+++ b/Entities/Common/Movement/FallDamage.as
@@ -5,6 +5,8 @@
 #include "KnockedCommon.as";
 #include "FallDamageCommon.as";
 
+const u8 knockdown_time = 12;
+
 void onCollision(CBlob@ this, CBlob@ blob, bool solid, Vec2f normal, Vec2f point1)
 {
 	if (!solid || this.isInInventory())
@@ -54,10 +56,7 @@ void onCollision(CBlob@ this, CBlob@ blob, bool solid, Vec2f normal, Vec2f point
 			}
 		}
 
-		// stun on fall
-		const u8 knockdown_time = 12;
-
-		if (!this.hasTag("dead") && doknockdown && setKnocked(this, knockdown_time))
+		if (getGameTime() == this.get_u32("death sound time stamp") && doknockdown && setKnocked(this, knockdown_time))
 		{
 			if (damage < this.getHealth()) //not dead
 				Sound::Play("/BreakBone", this.getPosition());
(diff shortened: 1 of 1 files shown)


Entities/Common/Movement/FallDamage.as
```diff
@@ -57,7 +57,7 @@ void onCollision(CBlob@ this, CBlob@ blob, bool solid, Vec2f normal, Vec2f point
 		// stun on fall
 		const u8 knockdown_time = 12;
 
-		if (doknockdown && setKnocked(this, knockdown_time))
+		if (!this.hasTag("dead") && doknockdown && setKnocked(this, knockdown_time))
```
AsuMagic, line 60: This causes `FallDeath.ogg` to not be played when falling to your death (at least in singleplayer).

AsuMagic (changes requested):

Entities/Common/Movement/FallDamage.as
mugg91: I committed a different solution.

Added `onHealthChange()` to `FallDamage.as`, which saves a bool to the runner blob.
If the old health was above zero, the bool is set true, otherwise false.
After that, `onCollision()` is executed next. 
Here, the sound is played only if the bool is true.

Because 12 is the only value 'knockdown_time' will be set to, I placed that variable outside the function.

mugg91: Nvm, it's not working in online. So the solution is not good to go yet.

mugg91:
I tried a lot of things now with `onHealthChange()` and `onHit()`. But it ain't working.

So I made a commit that uses `tickIfTag = "dead"` and added `onTick()` which adds another tag `"should be silent"` for one tick.
BreakBone.ogg and FallDeath.ogg are now only played if the blob doesn't have `"should be silent"`.
I tested it in offline and online and I'm not experiencing problems.
Let me know if there are any more issues.

Entities/Common/Movement/FallDamage.as
Gingerbeard5773: I found that this original solution (commit 1) works in both localhost and dedicated server. Why the requested changes? The first commit works as intended with no issues found during my testing. Here is a video showing this first change on a dedicated server, turn on sound:

https://user-images.githubusercontent.com/68350259/202875039-a7229379-24e3-4ac1-908f-096629c1cd80.mp4

As you can hear, the dead bodies do not make any sounds while the alive do.
Unless I am missing something, return to your first solution.

mugg91: I tested and this is the result:

Prior to this PR:
- Dead body makes a grunt sound when throwing from 13 blocks height
- Alive player falling from a very high place makes a death sound (grunt or bone break)

Commit 1:
- Dead body **doesn't** make a grunt sound when throwing from 13 blocks height (Good)
- Alive player falling from very high **doesn't** make a death sound (Bad)

Commit 4:
- Dead body **doesn't** make a grunt sound when throwing from 13 blocks height (Good) 
- Alive player falling from a very high place makes a death sound (grunt or bone break) (Good)

This also applies to `!builder` spawned in.
Maybe you didn't load the patched file correctly? I can make a video if needed.

AsuMagic (approved):
Workaroundy but will do for now

PR #1406, opened 2022-10-22
rebuild poll popup at server join

bunniewormy:
## Status

- **HOLD**: need to talk to verra to setup the poll so that the poll opens at set day && put in the proper link once its done

## Description

image


AsuMagic (approved):

mugg91:
Maybe remove `printf("yo");` at line 118.

PR #1407, opened 2022-10-22
[fixed] FakeRolling will properly work in online

### LINKED ISSUE #1384: FakeRolling doesn't properly spin items in online
## Description

Egg and Keg have `FakeRolling.as`.

In offline, a thrown Egg will spin around in the air and on the ground. 
But in online, it will not spin at all until it has stopped moving. 
After it stops moving, it assumes a new rotation.

In offline, a thrown Keg will spin around in the air and on the ground.
In online, it will not spin around in the air.
On the ground, it will spin around until it stops moving.
Sometimes it assumes a different rotation after it stops moving.

## Reproduction

Go to offline Sandbox and throw Egg and Keg.
Go to online Sandbox and throw Egg and Keg.
Notice the differences.

mugg91:
## Status

- **READY**: No known issues.

## Description

Fixes #1384.

`FakeRolling.as` is used by Egg and Keg.
When Egg/Keg is thrown in offline, it spins around but in online it didn't.
This PR changes the `FakeRolling.as` logic so the two items will spin around in the air when thrown in online.

Two values are saved to the blob, `angle0` and `angle1`.
Angle is fetched from one and then saved to the other every tick, alternatedly.
The value calculation and saving is done server-side and then synced.
[This causes almost 100% consistent values between client and server.](https://i.imgur.com/KLZZ6hU.png)
In very rare cases, a value may be different but will correct itself on the next tick or so.
I have not seen any desyncs/disconnections with this particular code.
The value that was fetched is applied via `setAngleDegrees()` both server and client side.
This console output in the above image is done by this test code:
```
if (this.getName() == "keg" && getGameTime() % 30 < 1)
print("I'm server? " + isServer() + " I'm client? " + isClient() + " time: " + getGameTime() + " angle: " + this.getAngleDegrees());
```

Because I spotted inconsistency between client and server values, I removed these lines, therefore angle is set every tick with no pause:
```
this.getCurrentScript().runFlags |= Script::tick_not_attached;
this.getCurrentScript().runFlags |= Script::tick_moving;
this.getCurrentScript().runFlags |= Script::tick_not_sleeping;
```

Although values are almost 100% consistent between client and server and no issues were spotted in my testing, it should still be tested if desync disconnections can occur in a large match setting.

I made it so angle is reset to 0 and no angle calculation, saving and applying is done when the blob is attached, in order to make Keg not look weird in hand.

Since Egg doesn't have spritelayer and has a circle hitbox, you could exchange `FakeRolling.as` with `CheapFakeRolling.as`.
I didn't do it in this PR, but I'm just pointing that out..

## Steps to Test or Reproduce

Go to online, throw a Keg through the air. Notice it spinning!

### DIFF
--- a/Entities/Common/Movement/FakeRolling.as
+++ b/Entities/Common/Movement/FakeRolling.as
@@ -3,35 +3,51 @@ const float MAX_ROTATION_SPEED = 60.0f;
 void onInit(CBlob@ this)
 {
 	f32 angle = (this.getNetworkID() * 977) % 360;
-	this.set_f32("angle0", angle);
-	this.set_f32("angle1", angle);
+	this.set_f32("angle 0", angle);
+	this.set_f32("angle 1", angle);
+	this.set_bool("should rotate 0", true);
+	this.set_bool("should rotate 1", true);
 }
 
 void onTick(CBlob@ this)
 {	
 	if (this.isAttached())	
 	{
-		this.set_f32("angle0", 0.0f);
-		this.set_f32("angle1", 0.0f);		
+		this.set_f32("angle 0", 0.0f);
+		this.set_f32("angle 1", 0.0f);		
 		return;
 	}
 	
-	f32 old_angle = this.get_f32("angle" + getGameTime() % 2); 
-	Vec2f vel = this.getVelocity();
+	f32 old_angle 			= this.get_f32("angle " + getGameTime() % 2); 
+	bool old_should_rotate 	= this.get_bool("should rotate " + getGameTime() % 2);
 	
-	if (Maths::Abs(vel.x) > 0 && isServer())
+	if (isServer())
 	{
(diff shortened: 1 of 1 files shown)


mugg91:
Egg was shaking in standstill position, so I made another commit that fixed that, by introducing two bools `"should rotate"`.
If velocity is greater than 0, it is set to true (otherwise stays false), then is saved to the blob and synced.

The "should rotate" value of the previous tick is fetched, called `old_should_rotate`.

`this.setAngleDegrees(old_angle);` is now only applied if `old_should_rotate` is true.

Looking at prints, `old_should_rotate` is the same between client and server, just like `old_angle`.
It turns from true to false or from false to true on the same tick between client and server.
So I think this is good.
Please test in a large match just in case, though.

Gingerbeard5773 (commented):
Really, all rolling blobs should be using CheapFakeRolling.as rather than FakeRolling.as . Setting the angle here from server isn't very efficient for a task that could be done client only.
I understand that fakerolling is only used because the keg has a spritelayer, and CheapFakeRolling doesn't support spritelayers (but it could!).

PR #1417, opened 2022-10-29
fix doors not being slashed with raycast slashes

bunniewormy:
## Status

## Description

theyre a temp blob xD


AsuMagic:
OK, so the "can't slash held items" commit (from the raycast PR) seems to be responsible for it.  
The actual issue is more that doors are tagged as a "temp blob". IIRC, this tag is supposed to be used for held blobs, and builder code should untag it when the blob is placed. Not sure what's going on.

AsuMagic (approved):

PR #1438, opened 2022-11-01
Bring back a (fixed) cinematic camera

AsuMagic:
## Status

## Description

The cinematic camera was introduced a few years ago but was removed early due to a client crash issue. It is rather likely that this issue was a dangling `CBlob@` reference, which was fixed in this branch.

Changes:
- Fixed(?) the likely crash with cinematic cameras
- Possibly optimized the importance routine

Issues to sort out:
- We really, really need less awful GUI that allows us to capture mouse clicks properly.


AsuMagic:
I will merge this since the original code was already reviewed and tested somewhat outside of the dangling blob bug. This will need a bunch of testing to make sure it feels right and doesn't have issues.

eps0003:
Thanks for reviving this, Asu!

PR #1465, opened 2022-11-08
[changed] chicken and fishy will not spawn inside Mines; RegrowPlants improvements; Enabled moss growth again

### LINKED ISSUE #1354: RegrowPlants.as can cause high tick times
**Description**

`RegrowPlants.as` is `SERVER_ONLY`, but how much it affects gameplay in practice is undetermined. Clients will not observe stutters in those cases, but it may contribute to perceived server lag/hitching.

During high map activity (e.g. during explosions) tick time spikes can be observed because of `RegrowPlant.as:onSetTile`. The culprit is likely to be `findTileByCoords` which is a linear lookup over all the recorded tiles:

```angelscript
u32 findTileByCoords(TileInfo@[] tiles, Vec2f coords)
{
	for (u32 i = 1; i < tiles.size(); i++)
	{
		TileInfo tile = tiles[i];
		if (tile.coords == coords)
		{
			return i;
		}
	}

	return 0;
}
```

There is some code involved in that hook that also imply linear complexity but I suspect it is much less of an issue because it all happens into native code, whereas this likely is strongly limited by script engine call performance.

The effect is likely to scale linearly with the amount of tiles in the map, so larger maps may be even more affected.

On staging in localhost mode, I have observed it to be >10ms over a single tick in certain cases. On weaker server CPUs, the effect is likely much more dramatic, especially on vanilla servers.

**Reproduction**

1. Use staging
2. Tap F5 -> Windows -> Profiler
3. Tick "Enable reliable profiling"
4. Tap F5
5. Spawn a few kegs and make them explode
6. Within <5 seconds, make sure to tap F5
7. Untick "Enable reliable profiling"
8. Identify an outlier you are likely to see
9. Hold middle click and move around to select the spike, then release
10. Observe `RegrowPlants.as:onSetTile` take the most CPU time during the tick

**Expected Behavior**

The CPU usage of `RegrowPlants.as` should be very small.

**Screenshots**

profiler output

**Your Setup:**

- OS: Linux
- CPU: 3900X

### LINKED ISSUE #1463: Chicken shouldn't spawn inside mines
## Current in-game behaviour.

Mine lies on the ground.
Chicken spawns inside the Mine.
Boom!

## Proposed Change

Chickens and fishies should set off mines, but not at the moment they spawn in.
Make chicken and fishy spawn only if Mine isn't inside a certain radius of their spawn position.

### LINKED ISSUE #1529: Moss collapse glitch
## Description

Enable moss in `RegrowPlants`
- Set `moss_stone` to true - enables moss growth.
- Increase `moss_stone_chance` to 0.05 - more moss growth for testing purposes.
- Set `moss_time` to 10 - don't need to wait 10 minutes.
- Set `min_random_time` and `max_random_inc` to 10 - very frequent checks. 

Play CTF and build a castle. Wait for moss to grow.
https://imgur.com/eaqk2P7.png

Collapse the castle.
https://imgur.com/2zYvgYD.png

Build a test wall.
https://imgur.com/3cLzy82.png

Then wait. Sometimes walls or backgrounds will appear (in the image, a block appeared from nowhere)
https://imgur.com/wF3B0u0.png

It seems, `castle_tiles` in `RegrowPlants.as` isn't properly kept up to date.
Maybe `onSetTile()` in `RegrowPlants.as` doesn't register (all of the) collapsing tiles.

mugg91:
# Status

- **READY**.
- No bugs observed.

# Description

[changed] Chicken spawns only on land, Fishy spawns only in water, both don't spawn near Mines (Fixes #1463)
[changed] Plants will not grow underwater.
[changed] Plants and grass will not grow near fireplace.
[changed] `RegrowPlants` performance improvements. (Partially fixes #1354)
[changed] Enabled moss growth again.
[changed] Moss will grow faster if water is adjacent.
[fixed] Moss collapse glitch. (Fixes #1529)

### Better performance

Separate dirt checks from castle checks.
Loop only through one or the other in one tick.
Check up to 80 dirt tiles or 80 castle tiles in one tick.
For dirt and for castle tiles, determines a factor by which the "next check time" should be reduced and to determine how many tiles should be checked.

To ensure that it still looks like growth happens in intervals and not continuously and thereby not be draining performance with checks being too frequent, each grow chance has been increased by factor 2.
So `grass_grow_chance = 0.015f;` is now `grass_grow_chance = 0.03f;` etc.
In exchange, the base wait time has been increased from 200 to 400 and the random time increase from 60 to 120.
For castle tiles, I even increased the base wait time to 600 and the random time increase to 150 to make moss growth more subtle.

So if you load a CTF map with 1800 ground tiles, this would be the scenario before this PR:
```
- Checks every 200 to 260 ticks.
- Checks all 1800 ground tiles in one tick, causing lag spikes.
```

And after this PR:
```
- Determines a factor of 23.
- 1800 tiles / 23 factor = 78 tiles checked per tick
- 400 base wait time / 23 factor = 17 ticks + (increased wait time of 0 to 59) 
- Checks every 17 to 76 ticks.
- Checks tiles 0 to 78 in one tick, then 79 to 157 on the next, etc.
```

To ensure some level of randomness and to make sure tiles aren't checked from left to right, each tile inserted into the array will receive a random index.

This might not fix all but at least some performance problems raised in [#1354](https://github.com/transhumandesign/kag-base/issues/1354).

Video before: https://www.youtube.com/watch?v=IBAil6zX8ig
Video after: https://www.youtube.com/watch?v=ERek2bTeTVU (Was recording during testing but also represents the end result anyway.)

### Additional conditions

It looks weird that flowers and grain can grow underwater.
It also looks weird when plants and grass can grow at a fireplace. 
So I made it so plants cannot grow underwater and plants+grass cannot grow when a burning fireplace is in a radius of 8. 

Chicken and Fish could spawn inside Mines and instantly set them off.
They could also spawn some distance away and still move into the Mines.
This adds an unwanted amount of randomness and luck to the game;
Players usually prefer execution and skill over randomness and luck.
Therefore, I set it so chicken and fish can't spawn when a Mine is in a radius of 32.

It looks odd that chicken could spawn underwater.
It was also possible for Fish to breed when on land.
So I added checks to prevent such.

### Moss growth

It was suggested that moss should grow on castle tiles if water is adjacent.
I noticed that it has been disabled by CatWiz in October 2020 in their PR [#928](https://github.com/transhumandesign/kag-base/pull/928) because moss castle tiles have slightly less health than their normal castle counterparts and mossification makes a "block placed" sound and creates particles.

These issues have not been addressed yet. My opinion is that moss growth isn't such a huge issue since walls are destroyed by bomb arrows, kegs and drill fast anyway. Normal castle tiles don't add that much more protection.
The "block placed" sound can only be fixed engine-side.
It seems, it makes the sound not because the block is placed but because the previous castle tile is removed.

Previously, moss growth had this behavior:
- Can grow on a castle tile if the castle tile existed for 10 minutes and is adjacent to other moss tile, bedrock or dirt.
- Pre-existing moss tiles don't count as "having moss adjacent".

I made the following additions:
- In `mossLuck()`, add `luck += 0.01f;` for each adjacent water.
- Change `hasMossAdjacent()` to `hasMossOrWaterAdjacent()`. 
This allows that moss can grow without having to wait 10 minutes if water is adjacent.

### Moss collapse glitch

See issue [#1529](https://github.com/transhumandesign/kag-base/issues/1529).
Castle tiles didn't get removed from the array after collapsing so the moss growth logic could cause tiles to appear in the same places that the collapsed tiles existed before.
I added this hook and I verified in testing that the bug doesn't happen anymore.
```
bool onMapTileCollapse(CMap@ this, u32 offset)
{
	Vec2f pos = this.getTileWorldPosition(offset);
	u32 tindex_castle = findTileByCoords(castle_tiles, pos);
	
	// castle tile collapsed, remove from array
	if (tindex_castle != 0 && castle_tiles.size() > 0)
	{
		castle_tiles.removeAt(tindex_castle);
	}
	return true;
}
``` 

# Reproduction Steps

### Animals spawning in Mines

Set `chicken_grow_chance` in `RegrowPlants.as` to be `0.1f` so chicken spawns very commonly.
Go to a small map like `Mazey_damnsmallflat`.
Fill the ground with Mines. 
Notice that chicken will spawn inside Mine. **After this PR , they only spawn next to Mines, if there is space.**

----

Use `!spawnwater`.
Kill all existing chickens.
Notice that new ones will spawn underwater.
**After this PR, they will not spawn inside water. If you dry a small area with sponges, chicken will spawn in that area only.**

----

Use `!spawnwater`.
Fill water with a bunch of mines.
Notice fishy will spawn inside Mines.
**After this PR, they will spawn only next to Mines, if there is space, instead.**

----

Build a small area where fishy can't escape.
`!fishy` twice.
Let the two fishies hop around until they die.
If you repeat a few times, sometimes they will breed a third fishy.
**After this PR, breeding only happens inside water.**

### Plants not growing 

Edit `RegrowPlants` to increase growth chance and decrease check times.

Go to a CTF map and spawn a bunch of fireplaces on normal dirt ground that doesn't have grass on it.
Wait until lots of plants and grass has grown.
Notice the plants and grass will have grown and covered the fireplaces.
**After this PR, there wouldn't be any plants or grass grown at the fireplaces.**

----

Go to a CTF map with water.
Wait until lots of plants and grass has grown.
Notice grain, flowers and bushes will have grown also underwater.
**After this PR, they wouldn't grow underwater.**

### Moss growth

Go to a CTF map and place castle tiles on dirt, bedrock or other pre-existing mossy tiles.
Wait for 10 minutes. 
Moss will start spreading to the castle tiles that existed for 10 minutes.

Place castle tiles in water.
Notice nothing happening for 10 minutes.
**After this PR, moss will grow without having to wait 10 minutes if water is adjacent.**

### Plants and grass growth in general

Play CTF.
Go to a custom map with lots of dirt strips in the air.
Notice lag spikes happening every few seconds and plants/grass growing after the lag spike happens.
**After this PR, lag spikes wouldn't happen and plants/grass grow in shorter intervals.**

### Moss collapse glitch

Play CTF.
Try to reproduce the glitch as seen in the video: https://www.youtube.com/watch?v=iDfqfLmywdk
**After this PR, it will not happen.**


mugg91:
I pushed a commit that makes these changes:
- Use `break;` instead of `return;` in `RegrowPlants.as`. We don't want to return from the whole function yet.
- Changed `SpawnFish.as` to account for the `while (i ++ < 3)` loop.

Rules/CommonScripts/RegrowPlants.as
```diff
@@ -292,7 +292,15 @@ void onTick(CRules@ this)
 
 				if (random_grow <= chicken_grow_chance && chicken_count < chicken_limit)
 				{
-					server_CreateBlob("chicken", -1, tinfo.coords - Vec2f(0,tilesize));
+					Vec2f spawn_position = tinfo.coords - Vec2f(0,tilesize);
+				
+					if (map.isBlobInRadius("mine", spawn_position, 16.0f) // don't spawn chicken inside mines
+						|| map.isInWater(spawn_position)) // don't spawn chicken inside water	
+					{
+						break;
```
AsuMagic, line 300: This is going to break out from the entire tile loop, and we don't want that.

asumagic (changes requested):

Rules/CommonScripts/RegrowPlants.as
mugg91: It should be better now.

mugg91:
Setting this to "in development" because I would like to add more changes to `RegrowPlants.as`.

- Improving performance and fixing [#1354](https://github.com/transhumandesign/kag-base/issues/1354).
Separate dirt checks from castle checks, loop only through one or the other in one tick.
Check up to 80 dirt tiles or 80 castle tiles in one tick, have the "next check time" reduced according to the number of dirt tiles on the map.
If there are more than 80 dirt tiles, then check the first 80 first, then the next 80 in the next check, etc.
- Plants shouldn't grow underwater.
- Plants and grass shouldn't grow when fireplace is near.
- Increase area chicken and fish can't spawn when Mine is near. (I have witnessed enough gameplay moments where chicken ruined my carefully placed Mines...)
- Enable moss growth on castle tiles again, but with some modifications:
Make mossification more subtle by changing added chance per adjacent dirt/bedrock/other mossy tile from `luck += 0.002f;` to `luck += 0.0015f;`.
Also give added chance when water is adjacent.
Make mossification not make a "tile placed" sound (This may only be done by dev on the engine-side).

mugg91:
I pushed two commits that addressed the things I raised in my previous comment.
Please refer to the opening post. I updated it with a lot of information.

mugg91:
Bush that spawns in Saw will get sawed immediately.
But I refrain from fixing it in `RegrowPlants.as` because it is fixed by #1348.

mugg91:
Let me know if I should split this PR up.

mugg91:
Converting to draft as I separate the changes into single PRs

PR #1467, opened 2022-11-08
[fixed] Greg not detaching players before attaching them

### LINKED ISSUE #1459: Greg can grab player that is already attached
## Description

Greg can grab player that is currently sleeping in quarters or sitting in a catapult.

The player will then be attached to both things at the same time.

Player should either detach or Greg should ignore those players.

mugg91:
## Status

## Description

Fixes #1459.

When Greg grabs a player who is already attached, then that player would be attached to that thing and Greg at the same time, causing glitchy behavior.

This PR makes it so the player detaches from catapult/bed/dinghy etc. before getting attached to Greg.
The player still holds onto their item in their hand when getting grabbed.
Tested in offline and online.

## Steps to Test or Reproduce

Go to Sandbox, spawn a Greg, get grabbed when sleeping in bed or sitting in a Catapult.
Notice you are still inside the bed or in the catapult but also fly away with Greg at the same time as indicated by the orange arrow pointer and camera movement. 
**After this PR, no such glitchy behavior happens, as the player properly detaches before getting attached to Greg.**

### DIFF
--- a/Entities/Zombies/Greg/Greg.as
+++ b/Entities/Zombies/Greg/Greg.as
@@ -527,8 +527,7 @@ void onCollision( CBlob@ this, CBlob@ blob, bool solid )
         this.set_s32("statue time", 0);
         this.set_u16("struggle count", 0);
 
-		if (blob.isAttached())
-			blob.server_DetachFromAll();
+        if (blob.isAttached()) { blob.server_DetachFromAll(); }
 		
         this.server_AttachTo(blob, this.getAttachmentPoint(0));
         this.setVelocity(Vec2f_zero);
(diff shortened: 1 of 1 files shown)


Entities/Zombies/Greg/Greg.as
```diff
@@ -526,6 +526,10 @@ void onCollision( CBlob@ this, CBlob@ blob, bool solid )
 
         this.set_s32("statue time", 0);
         this.set_u16("struggle count", 0);
+
```
AsuMagic, line 529: Should fix indent
```suggestion
    if (blob.isAttached()) { blob.server_DetachFromAll(); }
```

AsuMagic (approved):

PR #1475, opened 2022-11-13
Warboat crate drops gold when destroyed

### LINKED ISSUE #1449: Destroying a crate with Warboat doesn't give gold
## Current in-game behavior

Crate with outpost, purchased at vehicle shop for 50 gold, gives gold when it is destroyed.
Crate with warboat, purchased at boat shop for 50 gold, does **not** give gold when it is destroyed.

## Proposed change

Make crate with warboat give gold when it is destroyed.

Or consider removing the gold requirement, maybe also reduce the coin requirement. 
Because who wants to spend 50 gold and 250 coins on a shitty boat that enemies can farm coins from hitting? 
Consider applying changes I suggested in my comment [here](https://github.com/transhumandesign/kag-base/issues/1404).

eps0003:
## Status

## Description

Resolves #1449

## Steps to Test or Reproduce

1. Buy a warboat crate.
2. Destroy the crate.
3. Verify 50 gold is dropped.
4. Verify no gold is dropped for other crates that don't require gold to buy.

### DIFF
--- a/Entities/Items/Crate/Crate.as
+++ b/Entities/Items/Crate/Crate.as
@@ -28,7 +28,12 @@ void onInit(CBlob@ this)
 	string packed = this.get_string("packed");
 
 	this.set_u32("boobytrap_cooldown_time", 0);
-	this.set_s32("gold building amount", packed == "warboat" ? 50 : 0);
+	int goldAmount = 0;
+	if (packed == "warboat")
+	{
+		goldAmount = 50;
+	}
+	this.set_s32("gold building amount", goldAmount);
 
 	u8 frame = 0;
 	if (this.exists("frame"))
--- a/Entities/Items/Crate/Crate.as
+++ b/Entities/Items/Crate/Crate.as
@@ -25,14 +25,15 @@ void onInit(CBlob@ this)
 	this.addCommandID("stop unpack");
 	this.addCommandID("boobytrap");
 
+	string packed = this.get_string("packed");
+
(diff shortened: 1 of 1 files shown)


Entities/Items/Crate/Crate.as
```diff
@@ -25,14 +25,15 @@ void onInit(CBlob@ this)
 	this.addCommandID("stop unpack");
 	this.addCommandID("boobytrap");
 
+	string packed = this.get_string("packed");
+
 	this.set_u32("boobytrap_cooldown_time", 0);
-	this.set_s32("gold building amount", 0);
+	this.set_s32("gold building amount", packed == "warboat" ? 50 : 0);
```
AsuMagic, line 31: I'd rather move from a ternary to an if branch because adding cases would be super uglier with the ternary.
```suggestion
	int goldAmount = 0;
	if (packed == "warboat")
	{
		goldAmount = 50;
	}
	this.set_s32("gold building amount", goldAmount);
```

AsuMagic (approved):

mugg91:
The gold drops for Outpost and Warboat are declared in two different places.

Gold drop for Outpost crate is added via `VehicleShop.as` `onCommand()`
whereas this PR adds Gold drop for Warboat crate in `Crate.as` `onInit()`.

To be consistent, the Warboat crate gold drop should be declared in `BoatShop.as` `onCommand()` 
or the Outpost crate gold drop should be declared in `Crate.as` ``onInit()`.

PR #1478, opened 2022-11-13
Fix admin tab not showing on esc menu

### LINKED ISSUE #156: Admin Tab doesn't appear on esc menu.
this seems to be a long standing bug from kag where if the user is a admin or mod of a server and doesn't login to rcon and reload security the admin tab doesn't appear in the escape menu.

https://forum.thd.vg/threads/1591-has-admin-powers-but-has-no-admin-tab.24242/
this is fixed however when the user logs in and reloads security.

a simple fix would be to reload security when the server starts up automatically so people do not have to login to recon and reload the security every time the server is started.

eps0003:
## Status

## Description

Fixes #156 🎉 

Seems too simple of a fix to actually be the true fix to this longstanding issue, but it seems to work fine and is the solution players have manually done to resolve the issue (`/rcon /reloadseclevs`).

## Steps to Test or Reproduce

1. Be admin
2. Start and join a dedicated server
3. Verify the admin menu is visible

### DIFF
--- a/Rules/CommonScripts/KAG.as
+++ b/Rules/CommonScripts/KAG.as
@@ -8,6 +8,8 @@ void onInit(CRules@ this)
 	LoadDefaultMapLoaders();
 	LoadDefaultGUI();
 
+	getSecurity().reloadSecurity();
+
 	sv_gravity = 9.81f;
 	particles_gravity.y = 0.25f;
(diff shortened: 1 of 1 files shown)


AsuMagic (approved):

AsuMagic:
Should test things to see if that can cause dumb issues...

PR #1480, opened 2022-11-13
Prevent fast travel if tunnel/outpost is destroyed or team has changed

### LINKED ISSUE #1434: issues related to travelling via tunnels/outposts
## Description

Assume blue team has outposts A, B, C.
Blue Player P stands at outdoor C and has the tunnel select open.
Outpost A is converted to red team.
Blue Player P can still click on the menu button to travel to outpost A.

Pickup/throw will not close the tunnel select window, but activating items does.
So Player P cannot activate a Keg and then travel to the enemy outpost, but he can still bring a Keg with him and activate it after travelling. Theoretically, player P could camp at outpost C and wait for a good time (no enemy players waiting for him) to travel in with his Keg.

As an aside, assuming there is tunnels A, B, C and player P is standing at tunnel C and keeps the tunnel select opened.
If C is destroyed, the tunnel select will stay open for a short while and player P can still travel to A and B.

If there are only two tunnels A and B, it is also possible to pickaxe the last hit on your own tunnel B and tap "travel" to take your gold with you and travel to A.

## Suggested change

- Update the tunnel select window if any tunnels are converted or destroyed.
- Make it impossible to travel to a converted outpost.
- Make it impossible to travel at all if the tunnel you travel from is destroyed.

eps0003:
## Status

## Description

Resolves #1434 except the issue of the menu not updating immediately which affects more than just tunnels.

## Steps to Test or Reproduce

Outpost team changed:
1. Have three or more outposts
2. Open the travel menu
3. Have an enemy capture one of the outposts
4. Attempt to travel to the captured outpost

Tunnel destroyed:
1. Have three or more tunnels or outposts
2. Open the travel menu
3. Have someone destroy the tunnel/outpost you're travelling from
4. Attempt to travel

### DIFF
--- a/Entities/Industry/Tunnel/TunnelTravel.as
+++ b/Entities/Industry/Tunnel/TunnelTravel.as
@@ -146,6 +146,10 @@ void Travel(CBlob@ this, CBlob@ caller, CBlob@ tunnel)
 		//if (isKnockable(caller) && caller.get_u8("knocked") > 0)
 		//	return;
 
+		//dont travel if tunnel team has changed while tunnel menu was open
+		if (tunnel.getTeamNum() != caller.getTeamNum())
+			return;
+
 		//dont travel if caller is attached to something (e.g. siege)
 		if (caller.isAttached())
 			return;
--- a/Entities/Industry/Tunnel/TunnelTravel.as
+++ b/Entities/Industry/Tunnel/TunnelTravel.as
@@ -148,7 +148,7 @@ void Travel(CBlob@ this, CBlob@ caller, CBlob@ tunnel)
 		//	return;
 
 		//dont travel if tunnel team has changed while tunnel menu was open
-		if (tunnel.getTeamNum() != caller.getTeamNum())
+		if (this.getTeamNum() != tunnel.getTeamNum())
 			return;
 
 		//dont travel if caller is attached to something (e.g. siege)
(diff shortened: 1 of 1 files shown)


Entities/Industry/Tunnel/TunnelTravel.as
```diff
@@ -139,13 +139,18 @@ bool doesFitAtTunnel(CBlob@ this, CBlob@ caller, CBlob@ tunnel)
 
 void Travel(CBlob@ this, CBlob@ caller, CBlob@ tunnel)
 {
-	if (caller !is null && tunnel !is null)
+	CBlob@ thisTunnel = getBlobByNetworkID(this.getNetworkID());
```
AsuMagic, line 142: er, wtf? Surely we should prevent this possibility from happening prior to the `Travel` call. If `this` has a possibility of being a dead blob, then unless we're sure it happens within the same tick, there is a possibility of crashing if the tunnel blob was deleted.

AsuMagic (changes requested):

Entities/Industry/Tunnel/TunnelTravel.as
eps0003: This appears to be a problem with how generic buttons work. I haven't checked, but it appears as though it doesn't check if the blob exists when it sends the command when a button is clicked.

Considering no crashes have occurred due to this as far as we know, I think it's fine to just keep this fix.

AsuMagic (approved):

PR #1482, opened 2022-11-13
Hide blood on spikes when kidssafe is enabled

### LINKED ISSUE #1352: Spikes should not show blood when "kids safe" is enabled
## Current in-game behaviour.

When "kids safe" is disabled and you jump in a Saw, it shows a bloody sprite.
When "kids safe" is enabled, it will not show a bloody sprite.

This difference is not present with spikes. 
Spikes will always be bloody after jumping in them.

Same with Spiker.

## Proposed Change

To be consistent with Saw, make Spikes and Spiker only bloody when "kids safe" is disabled.

eps0003:
## Status

## Description

Resolves #1352

Spikes still keep track of whether they are bloody but don't show that to players with `g_kidssafe` enabled. This means spikes will immediately turn bloody when the setting is disabled if someone has jumped into the spikes.

## Steps to Test or Reproduce

1. Enable `g_kidssafe`
2. Jump into spikes
3. Verify blood isn't visible on spikes
4. Disable `g_kidssafe`
5. Notice that blood becomes visible

### DIFF
--- a/Entities/Structures/Components/Load/Spiker/GenericSpike.as
+++ b/Entities/Structures/Components/Load/Spiker/GenericSpike.as
@@ -64,6 +64,8 @@ void onHitBlob(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@
 
 	this.Tag("bloody");
 
+	if (g_kidssafe) return;
+
 	CSpriteLayer@ layer = this.getSprite().getSpriteLayer("blood");
 	if (layer is null) return;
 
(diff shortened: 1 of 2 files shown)


ewetcr:
Saw could enter here too, it gets blood on the metal piece.

eps0003:
Saws already don't show blood when `g_kidssafe` is enabled.

AsuMagic (approved):

mugg91:
I tested it.

For spikes, it updates from blood to no blood and vice versa after changing the kidssafe setting.
But only if you are within the proximity of `124.0f`and moving around.
If you are further away or just standing still forever, it will not update.
This is fine though.

For spiker, it doesn't update. 
If you get hit while the kidssafe setting is off, it will become bloody but turning kidssafe off doesn't remove the bloodiness.
If you get hit while the kidssafe setting is on, it will never ever turn bloody. Not even when you turn kidssafe off again. 
This looks improvable.

If one wants to go high effort on this, you could use `void OnCloseMenu(CRules@ this)` and write a separate function that fetches all Saws, spikes and spikers and updates the blood layer all at once after changing the setting.

Issue #1314, opened 2022-07-10
Can't see background of Tunnel, Storage and Quarry


mugg91:
## Current in-game behaviour.

When building a Tunnel, Storage or Quarry, like with all buildings they will have a window added to them (center tile's background is removed). For those buildings you cannot see the background, though. So if building these in an underground place, the missing background will cause your underground stuff to be visible to the enemy. 

This seems to not be obvious to many players. In CTF I see my team has placed background everywhere but our tunnel is bright due to the missing background.

## Proposed Change

- When building one of those three, don't remove the center tile's background.
- ~~Make the background to last forever and not take any damage until the building is destroyed. Being stone buildings, they should not have their background removed by fire anyway.~~

## Additional Info

There may be ways to remove backgrounds that one should be aware of. The main problem is with the buildings starting out with missing background to begin with, though.

- Please be aware a bug exists which allows to remove backgrounds. A fix exists but has yet to be merged. #592 

- I'm not quite sure but catapulted boulders may have special property that allows destruction of tiles even in places that are supposed to be invincible, such as below a tent. So if making backgrounds invincible, they may still be destroyed by boulders, maybe.

PR #1483, opened 2022-11-13
Buildings no longer remove center backwall when built on hand-placed backwall 

eps0003:
## Status

## Description

Currently, the backwall of a building is generated and then 10 ticks later the center tile is set to air to make a window. This PR changes this so it generates the backwall with the window on the same tick, and the window doesn't remove any existing hand-placed backwall.

Resolves the first proposed change in #1314

## Steps to Test or Reproduce

1. Build a building in the open
2. The window is generated as normal
3. Build a wall of backwall
4. Build a building in front of that backwall
5. Notice that the window is not generated because the backwall exists

### DIFF
--- a/Entities/Structures/Common/AddTilesBySector.as
+++ b/Entities/Structures/Common/AddTilesBySector.as
@@ -20,3 +20,28 @@ void AddTilesBySector(Vec2f ul, Vec2f lr, string sectorName, TileType tile, Tile
 		}
 	}
 }
+
+void AddTilesBySectorWithWindow(Vec2f ul, Vec2f lr, string sectorName, TileType tile, TileType omitTile = 255)
+{
+	if (getNet().isServer())
+	{
+		CMap@ map = getMap();
+		const f32 tilesize = map.tilesize;
+		Vec2f tpos = ul;
+		Vec2f tsize(map.tilesize, map.tilesize);
+		Vec2f center = ul + (lr - ul - tsize) * 0.5f;
+		while (tpos.x < lr.x)
+		{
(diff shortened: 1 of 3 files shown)


Entities/Structures/Common/AddTilesBySector.as
```diff
@@ -20,3 +20,28 @@ void AddTilesBySector(Vec2f ul, Vec2f lr, string sectorName, TileType tile, Tile
...
+		CMap@ map = getMap();
+		const f32 tilesize = map.tilesize;
+		Vec2f tpos = ul;
+		Vec2f tsize(map.tilesize, map.tilesize);
+		Vec2f center = ul + (lr - ul - tsize) * 0.5f;
+		while (tpos.x < lr.x)
+		{
+			while (tpos.y < lr.y)
+			{
+				if (map.getSectorAtPosition(tpos, sectorName) !is null && map.getTile(tpos).type != omitTile && tpos != center)
```
AsuMagic, line 37: Note: exact floating-point comparisons can cause issues due to precision.  
I have checked in engine, though, and `Vec2f::opCmp` luckily uses an epsilon value, so it is not an exact comparison. It seems to be OK in this usecase.

AsuMagic (approved):

mugg91:
This PR doesn't resolve #1314 in contrary to what is stated in the opening message.

#1314 suggests that Tunnel, Outpost, Storage and Quarry have stone background everywhere without any window.

This PR only adds a mechanic that doesn't remove a pre-built background tile in the window's tile position.

PR #1486, opened 2022-11-13
[modified] amount of materials in build time (balance change)

bunniewormy:
## Status

## Description

The amount of materials added to the game because of the build time changes was underestimated, so it's being modified.

REBUILD (CURRENT):
30s timer (resupply at 180/150/120/90/60/30) - 6 resupplies 
21600 wood total (12 players, 300 per player)
7200 stone total (12 players, 100 per player)

THIS PR:
40s timer (resupply at 180/140/100/60/40) - 5 resupplies 
12000 wood total (12 players, 200 per player)
3600 stone total (12 players, 60 per player)

The 50% reduction in stone may be a bit harsh, but it is still going to be more mats than pre-rebuild. Moreover, most maps also have tons of stone accessible from stone ore at the start of the game (eg. elantris - easy 3k), so it's not gonna be a 50% reduction in practice - probably 1/3 to 1/4 on most maps.

However, if it will be too harsh of a decrease (need to see in-game) then probably it'll be increased to be something in between.

### DIFF
--- a/Rules/CTF/Scripts/CTF_GiveSpawnItems.as
+++ b/Rules/CTF/Scripts/CTF_GiveSpawnItems.as
@@ -6,7 +6,7 @@
 const u32 materials_wait = 20; //seconds between free mats
 const u32 materials_wait_warmup = 40; //seconds between free mats
 
-const int warmup_wood_amount = 250;
+const int warmup_wood_amount = 200;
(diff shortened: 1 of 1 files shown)


AsuMagic (approved):
lgtm following discussion over at https://discord.com/channels/337250580303052801/337266863841214465/1041405526803755088

PR #1490, opened 2022-11-14
Add missing translation support

eps0003:
## Status

## Description

Search for all occurrences of `.getInventoryName()` and made sure they all have translation support. If it was commented out, I just deleted the unused code.

## Steps to Test or Reproduce

Add translations for the text updated.

### DIFF
--- a/Entities/Items/Crate/MakeCrate.as
+++ b/Entities/Items/Crate/MakeCrate.as
@@ -10,7 +10,7 @@ shared CBlob@ server_MakeCrate(string blobName, string inventoryName, int frameI
 		crate.server_setTeamNum(team);
 		crate.setPosition(pos);
 		crate.set_string("packed", blobName);
-		crate.set_string("packed name", inventoryName);
+		crate.set_string("packed name", getTranslatedString(inventoryName));
 		crate.set_u8("frame", frameIndex);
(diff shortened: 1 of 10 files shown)


Entities/Items/Crate/MakeCrate.as
```diff
@@ -10,7 +10,7 @@ shared CBlob@ server_MakeCrate(string blobName, string inventoryName, int frameI
 		crate.server_setTeamNum(team);
 		crate.setPosition(pos);
 		crate.set_string("packed", blobName);
-		crate.set_string("packed name", inventoryName);
+		crate.set_string("packed name", getTranslatedString(inventoryName));
```
AsuMagic, line 13: Then you also need to update `Crate.as` which uses this property:

```angelscript
			this.setInventoryName("Crate with " + this.get_string("packed name"));
```

And shouldn't we rather use `getTranslatedString` wherever the `"packed name"` property is being read?

AsuMagic (changes requested):

Entities/Items/Crate/MakeCrate.as
eps0003: Done

AsuMagic (approved):

PR #1493, opened 2022-11-14
[fixed] CTF shops never refreshed as you stopped/began passing requirements while open

AsuMagic:
## Status

## Description

**Requires testing on server! Only tested on localhost.** However, I believe it should work because the requirements checking code seems to be client-side in the first place.

Now, shop GUIs are refreshed every tick to update button status. Check reproduction steps.  
It might seem silly to refresh requirements `onTick` but it really shouldn't be a big deal, and should allow us to avoid thinking about every edge case if taking an event-based approach.

## Steps to Test or Reproduce

Examples:

Knight shop example:

1. Build a knight shop
2. Set your coins to 200
3. Buy bombs, see options gray out as you run out of coins

Builder shop example:

1. Build a builder shop
2. Buy things with wood, see options gray out as you run out of resources

Food shop example (requires master as of 4173-hotfix1):

1. Build quarters
2. Hit yourself with a spike
4. Buy food, see the food options gray out as you go full health
5. Drop a spike and quickly open the shop, see the healing options show up as you get damage

Without the PR, options are never updated until you reopen the shops.

This PR currently only affects CTF shops, not other grid menus affected by the same problem. Such code needs to be edited independently, there is no common code we could do to automagically fix it all.

### DIFF
--- a/Entities/Common/Factory/Shop.as
+++ b/Entities/Common/Factory/Shop.as
@@ -450,8 +450,6 @@ void BuildShopMenu(CBlob@ this, CBlob @caller, string description, Vec2f offset,
 	this.set_string("open menu name", caption);
 	this.set_netid("open menu caller blob", caller.getNetworkID());
 
-	print("build shop menu " + getTranslatedString(description));
-
 	if (menu !is null)
 	{
 		if (!this.hasTag(SHOP_AUTOCLOSE))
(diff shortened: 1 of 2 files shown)


Vam-Jam (approved):
asu my beloved

PR #1501, opened 2022-11-18
[dev] fix to shop refresh broken logic

AsuMagic:
## Status

## Description

This is a rather braindead way to do things (as `onTick` could be moved to a rules script there), but it is generally not a big deal. We move the refresh-related properties from shop blob properties to rule properties.

I'd rather not do this, but ATM, the only way to get an open `CGridMenu` appears to be to get it by its caption. Of course, different shops are highly likely to have the same caption.

One could hold onto a `CGridMenu` reference through a generic `get`/`set` but we have seemingly no way of figuring out whether a grid menu is still alive at that point or not.

Considering there is no situation where we can have multiple shops open, this seems to be a functional way to achieve this.

### DIFF
--- a/Entities/Common/Factory/Shop.as
+++ b/Entities/Common/Factory/Shop.as
@@ -39,11 +39,6 @@ void onInit(CBlob@ this)
 			this.set_u8("shop button radius", 16);
 		}
 	}
-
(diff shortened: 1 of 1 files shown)


Vam-Jam (approved):
jenkins go brr

PR #1502, opened 2022-11-19
[fixed] Plants growing off incorrect tile

Gingerbeard5773:
## Status

- **READY**: this PR is ready.

## Description

For plants to grow, they use a tile one block deeper than expected rather than the surface tile. This change simply changes it so the surface tile is always used instead.
Untitled

## Steps to Test or Reproduce

Place a plant of any kind onto the ground. The plant will always grow as long as there is a dirt block underneath the plant.

### DIFF
--- a/Entities/Natural/Scripts/PlantGrowth.as
+++ b/Entities/Natural/Scripts/PlantGrowth.as
@@ -17,10 +17,8 @@ void onInit(CBlob@ this)
 
 void onTick(CBlob@ this)
 {
-	if (getNet().isServer())
+	if (isServer())
 	{
-		Vec2f pos = this.getPosition();
-
 		u8 amount = this.get_u8(grown_amount);
 
 		u8 time = this.get_u8(growth_time);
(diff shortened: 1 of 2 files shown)


mugg91:
I tested with seeds and variants of
```
print("0: " + map.isTileGround(map.getTile(Vec2f(pos.x, pos.y)).type) 
+ " 4: " + map.isTileGround(map.getTile(Vec2f(pos.x, pos.y+4)).type) 
+ " 8: " + map.isTileGround(map.getTile(Vec2f(pos.x, pos.y+8)).type));
```
It seems that it doesn't make a difference if you use `+ 4` or `+ 8`.
The print returning `false true true`.
Therefore it seems to consider the tile right below itself before and after this change.
Not sure if naturally growing plants use `canGrowAt()` and how this PR would affect that, though.

Gingerbeard5773:
Seeds work as intended as you tested. The bug that has been fixed in this PR was regarding growing plants (other than trees) like grain and flowers rather.

mugg91:
Alright. I got to test with a naturally growing grain plant.
Assuming it grows on a layer of dirt following a layer of wood, then the print returns
`true true false`.
So +0 and +4 seems to work?
I guess `else if (canGrowAt(this, (pos + Vec2f(0.0f, 6.0f))))` in `PlantGrowth.as` is at work, and that means it's actually looking for "plant center position + 6.0f (if +0 is used) or + 10.0f (if +4 is used).

AsuMagic (approved):

PR #1506, opened 2022-11-21
No longer alert admins when it shouldn't

eps0003:
## Status

## Description

No longer alert admins when:
* Materials merge into one blob
* Material has zero quantity

## Steps to Test or Reproduce

Slash gold ore and let the gold merge.


ParaLogia (approved):
Tested with various quantities of gold in single player. Everything seems to work correctly.

AsuMagic (approved):

PR #1507, opened 2022-11-21
[fixed] Spikes, Spiker and Saw will properly remember if they are bloody and will instantly update their sprite when g_kidssafe changes

mugg91:
## Status

## Description

```
[fixed] Spikes, Spiker and Saw will properly remember if they are bloody or not and will instantly update their spritelayer when changing `g_kidssafe` from the menu.
[changed] Spikes don't continuously update their animation frame in `onTick()` via calling `onHealthChange()`
[changed] Spiker will change its animation frame rather than adding a spritelayer
```

Spikes remembered its bloody status but only updated its sprite frame if a player moved inside a radius of 124.0f.
It also called `onHealthChange()` to do it on each tick even if the frame doesn't need updating.

Spiker would either be bloody or not bloody forever after the first hit lands due to a bad return condition in `onHitBlob()`.

Saw turns bloody when `g_kidssafe` was off and then stayed that way forever.

This PR makes it so all three (Spikes, Spiker, Saw) will properly remember if they are bloody or not and update their animation frame accordingly when `g_kidssafe` is changed via the menu. When one of the three spawns, then their `onInit()` will run this:

```
	CRules@ rules = getRules();
	if (!rules.hasScript("ToggleBloodyStuff.as"))
	{
		rules.AddScript("ToggleBloodyStuff.as");
	}
```

And `ToggleBloodyStuff.as` will update the animation frames of all Saws, Spikes and Spikers on the map according to whether they are tagged with "bloody" and `g_kidssafe`. This is done via the hook `void OnCloseMenu(CRules@ this)`.

I didn't want to add `ToggleBloodyStuff.as` as a script to all gamemode config files in order to prevent clutter and since I feel that this is something that should always reliably run regardless what gamemode is played. Of course I can do it if a dev wants me to.

Tested in online and offline, no bugs observed.

## Steps to Test or Reproduce

Have `g_kidssafe` off and jump into some Spikes, Spikers and Saws.
Turn `g_kidssafe` on via settings menu.
Notice that the things that were bloody before are now not bloody.
Turn `g_kidssafe` off again.
Notice that the things that were not bloody are now bloody again.

## Ideas to make this Better

- Have the bloodiness fade away depending how much time passed. This way, you can see which spikes/Saw killed a player most recently.
- Make blood particles change the environment, like turning blocks bloody depending what side it's hit from. Have it go away after some time.

### DIFF
--- a/Entities/Industry/Saw/Saw.as
+++ b/Entities/Industry/Saw/Saw.as
@@ -15,6 +15,12 @@ void onInit(CBlob@ this)
 	this.addCommandID(sawteammate_id);
 
 	SetSawOn(this, true);
+	
+	CRules@ rules = getRules();
+	if (!rules.hasScript("ToggleBloodyStuff.as"))
+	{
+		rules.AddScript("ToggleBloodyStuff.as");
+	}
 }
 
 //toggling on/off
@@ -64,7 +70,7 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
 		bool set = !getSawOn(this);
 		SetSawOn(this, set);
 
-		if (getNet().isClient()) //closed/opened gfx
+		if (isClient()) //closed/opened gfx
 		{
 			CSprite@ sprite = this.getSprite();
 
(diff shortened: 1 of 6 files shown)


mugg91:
I made a commit that fixed that the first placed blob wasn't considered. Setting back to "ready".

mugg91:
I will open another PR to replace this one.

PR #1511, opened 2022-11-24
[changed] Tunnel, Storage, Quarry and Outpost won't get a window added to them

### LINKED ISSUE #1314: Can't see background of Tunnel, Storage and Quarry

mugg91:
## Status

## Description

[changed] Tunnel, Storage, Quarry and Outpost don't have a window. (Fixes #1314)
[changed] Condense `AddTilesBySector()` and `AddTilesBySectorWithWindow()` into one function.

This makes it so certain buildings where the window cannot be seen are not receiving a window upon building them.

Knightshop, Archershop, Buildershop, Boatshop, Vehicleshop, Quarters and unconstructed Building will get tagged `"has window"` in their `onInit()`.

Then `AddTilesBySector.as` - which places the building's background tiles - will check for that tag. If it does have the tag, it will not place a background tile over the center tile.

Tunnel, Storage, Quarry and Outpost are not tagged with `"has window"`, therefore they will have background placed over the center tile.

I included Quarry because it doesn't hurt to update it while I'm at it.

Because `AddTilesBySector()` and `AddTilesBySectorWithWindow()` are almost duplicate and because of this change, I made it into one function.

Tested in offline, should work in online. No bugs observed.

## Steps to Test or Reproduce

Place a building or use `!knightshop`.
Notice there will be a window.
Construct or spawn a tunnel / quarry / outpost / storage.
Notice there is no window; you cannot place a background tile over the center tile.

### DIFF
--- a/Entities/Structures/Common/AddTilesBySector.as
+++ b/Entities/Structures/Common/AddTilesBySector.as
@@ -1,29 +1,6 @@
-void AddTilesBySector(Vec2f ul, Vec2f lr, string sectorName, TileType tile, TileType omitTile = 255)
+void AddTilesBySector(Vec2f ul, Vec2f lr, string sectorName, TileType tile, TileType omitTile = 255, bool hasWindow = false)
 {
-	if (getNet().isServer())
-	{
-		CMap@ map = getMap();
-		const f32 tilesize = map.tilesize;
-		Vec2f tpos = ul;
-		while (tpos.x < lr.x)
-		{
-			while (tpos.y < lr.y)
-			{
-				if (map.getSectorAtPosition(tpos, sectorName) !is null && map.getTile(tpos).type != omitTile)
-				{
-					map.server_SetTile(tpos, tile);
-				}
-				tpos.y += tilesize;
-			}
-			tpos.x += tilesize;
-			tpos.y = ul.y;
-		}
-	}
-}
-
-void AddTilesBySectorWithWindow(Vec2f ul, Vec2f lr, string sectorName, TileType tile, TileType omitTile = 255)
-{
-	if (getNet().isServer())
+	if (isServer())
 	{
 		CMap@ map = getMap();
 		const f32 tilesize = map.tilesize;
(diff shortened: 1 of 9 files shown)


Entities/Structures/Common/AddTilesBySector.as
```diff
@@ -34,7 +11,9 @@ void AddTilesBySectorWithWindow(Vec2f ul, Vec2f lr, string sectorName, TileType
 		{
 			while (tpos.y < lr.y)
 			{
-				if (map.getSectorAtPosition(tpos, sectorName) !is null && map.getTile(tpos).type != omitTile && tpos != center)
+				if (map.getSectorAtPosition(tpos, sectorName) !is null 
+					&& map.getTile(tpos).type != omitTile
+					&& !(tpos == center && hasWindow))
```
eps0003, line 16: Rearranging this set of conditions like so makes it a bit more readable in my opinion. Feel free to change this if you agree.
```suggestion
					&& (!hasWindow || tpos != center))
```

eps0003 (approved):

Issue #1520, opened 2022-11-26
Can't change visibility of spritelayers of blobs that were put in inventory


mugg91:
## Description

1) Blob with spritelayer A exists.
2) A is visible.
3) Blob is put in an inventory.
4) A is set to non-visible.
5) Blob is taken from the inventory.
6) A is visible now.

It seems, either the game puts the whole sprite and all its layers to visible, or it sets the spritelayer to the same visibility as when it was put in inventory. In either case, this prevents setting the visibility ourselves.
Needs testing.

PR #1521, opened 2022-11-26
[fixed] Blobs will keep burning when in inventory

### LINKED ISSUE #1401: Resolve in-inventory burning
## Current in-game behaviour.

When putting a burning item in your inventory, it will not finish burning.
It will continue burning and taking burn damage after it has been put out of inventory.

## Proposed Change

Continue burning even when put in an inventory blob.

mugg91:
## Status

## Description

[fixed] Blobs will keep taking burn damage when in an inventory blob. (Fixes #1401)

I added this code to `IsFlammable.as` which makes it so the script keeps ticking when the blob is currently in inventory.
The blob will keep taking burn damage and the burning will resolve.
```
void onThisAddToInventory(CBlob@ this, CBlob@ inventoryBlob)
{
	// keep burning in inventory
	this.doTickScripts = true;

}
```

The extinguish sound didn't play when burning resolved in inventory, so I removed `PlaySound("/ExtinguishFire.ogg");` from `FireAnim.as` and make it play in `IsFlammable.as` in the `//check if we're extinguished` section.

Since `FireAnim.as` would tick only every 12 ticks, removing a blob whose burning has resolved in-inventory would still show its fire spritelayer for a short moment. I tried a lot of things but unfortunately, I couldn't fix it.
You can't check for `blob.isInInventory()` in `FireAnim.as` since it doesn't even tick when the blob is in inventory.
Setting the spritelayer to non-visible in `IsFlammable.as` in the `//check if we're extinguished` section didn't work.
Using something like
```
void onThisRemoveFromInventory(CBlob@ this, CBlob@ inventoryBlob)
{
	CSpriteLayer@ fire = this.getSprite().getSpriteLayer("fire_animation_large");
	if (fire !is null)
	{
		fire.SetVisible(false);
		this.getSprite().ReloadSprites(0,0);
	}
}
```
also didn't get the desired result. When the blob is taken from a crate into your hand or when the crate is destroyed, the blob would still show the fire spritelayer. Setting tick frequency of `FireAnim.as` to 1 seems too draining.
So I'm leaving it for now and because of these observations, I started issue #1520

## Testing Steps

Use a grain for testing, it takes burn damage but never gets destroyed.
Hit the grain with a fire arrow, then see how it looks if it finishes burning on the ground / in your inventory / in a crate etc.

Or make a log burn and place the log in an inventory.
Notice that it gets destroyed in inventory eventually.

### DIFF
--- a/Entities/Common/FireScripts/FireAnim.as
+++ b/Entities/Common/FireScripts/FireAnim.as
@@ -30,6 +30,7 @@ void onInit(CSprite@ this)
 
 void onTick(CSprite@ this)
 {
+	this.getCurrentScript().tickFrequency = 24;
 	CBlob@ blob = this.getBlob();
 	CSpriteLayer@ fire = this.getSpriteLayer("fire_animation_large");
 	
--- a/Entities/Common/FireScripts/FireAnim.as
+++ b/Entities/Common/FireScripts/FireAnim.as
@@ -30,9 +30,9 @@ void onInit(CSprite@ this)
 
 void onTick(CSprite@ this)
 {
-	this.getCurrentScript().tickFrequency = 24; // opt
 	CBlob@ blob = this.getBlob();
 	CSpriteLayer@ fire = this.getSpriteLayer("fire_animation_large");
+	
 	if (fire !is null)
 	{
 		//if we're burning
@@ -40,7 +40,8 @@ void onTick(CSprite@ this)
 		{
 			this.getCurrentScript().tickFrequency = 12;
 
-			fire.SetVisible(true);
+			if (!fire.isVisible())
+				fire.SetVisible(true);
 
 			//TODO: draw the fire layer with varying sizes based on var - may need sync spam :/
 			//fire.SetAnimation( "bigfire" );
@@ -53,11 +54,7 @@ void onTick(CSprite@ this)
 			}
 		}
 		else
-		{
-			if (fire.isVisible())
-			{
-				this.PlaySound("/ExtinguishFire.ogg");
-			}
+		{	
 			fire.SetVisible(false);
 		}
 	}
--- a/Entities/Common/FireScripts/IsFlammable.as
+++ b/Entities/Common/FireScripts/IsFlammable.as
@@ -83,7 +83,10 @@ void onTick(CBlob@ this)
 		this.set_netid("burn starter blob", 0);
 		
 		CSprite@ sprite = this.getSprite();
-		sprite.PlaySound("/ExtinguishFire.ogg");
+		if (sprite !is null)
+		{
+			sprite.PlaySound("/ExtinguishFire.ogg");
+		}
 	}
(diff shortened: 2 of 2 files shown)


Entities/Common/FireScripts/FireAnim.as
```diff
@@ -30,17 +30,19 @@ void onInit(CSprite@ this)
...
 	if (fire !is null)
 	{
 		//if we're burning
 		if (blob.hasTag(burning_tag))
 		{
 			this.getCurrentScript().tickFrequency = 12;
 
-			fire.SetVisible(true);
+			if (!fire.isVisible())
+				fire.SetVisible(true);
```
eps0003, line 45: Why is this change required? The added `if` statement looks redundant.

Entities/Common/FireScripts/IsFlammable.as
```diff
@@ -81,6 +81,9 @@ void onTick(CBlob@ this)
 	{
 		server_setFireOff(this);
 		this.set_netid("burn starter blob", 0);
+		
+		CSprite@ sprite = this.getSprite();
+		sprite.PlaySound("/ExtinguishFire.ogg");
```
eps0003, line 86: Add a `null` check for the sprite just in case it doesn't exist for some reason.

eps0003 (changes requested):

Entities/Common/FireScripts/FireAnim.as
mugg91: It is not required, I just figured it would be good practise to do it this way.
I tested with
```
	for (int z = 0; z < 500000; z++)
	{
		if (!fire.isVisible())
			fire.SetVisible(true);
	}
```
I barely could see any lag and I couldn't spot a difference in performance between having `if (!fire.isVisible())` vs. not having it. That also means there is no gain in having `if (!fire.isVisible())`. Should I revert it?

Entities/Common/FireScripts/IsFlammable.as
mugg91: Done.

Entities/Common/FireScripts/FireAnim.as
AsuMagic: It is indeed not useful because IIRC it is just a mere boolean flag in engine. Some other things, like SetLight, involve more involved logic when the state is _actually_ changed, but the check is usually redundant anyway.

AsuMagic: Generally speaking the bottleneck in this kind of call-heavy scripts is the FFI (i.e. calling engine functions from the script VM) in which case causing more calls could actually lower performance. Then again, that code is really not a hot loop, so it doesn't matter much. That is especially since it ticks at a throttled rate, which skips script execution completely, and while not free, it is probably as much of a performance save as you can get.

PR #1522, opened 2022-11-27
Hide resupply text above inventory GUI if help is hidden

eps0003:
## Status

## Description

It is very obnoxious and shouldn't need to be shoved in the face of most players.

## Steps to Test or Reproduce

1. Have resupply available
2. Toggle help with F1

### DIFF
--- a/Rules/CTF/Scripts/CTF_GiveSpawnItems.as
+++ b/Rules/CTF/Scripts/CTF_GiveSpawnItems.as
@@ -199,7 +199,7 @@ void displayResupply(CRules@ this, string player_class, string resupply_class, V
 
 		GUI::DrawTextCentered(text, Vec2f(x, y), color);
 	}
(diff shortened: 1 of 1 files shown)


AsuMagic:
Random thought: Could we maybe move this + the "resupply in _x_ seconds" message in the item bar as an icon or badge in some way (but so that it remains obvious and easily discoverable even for newbies)?  
Those messages do feel like they clutter up the screen somewhat (at least at lower resolutions) but at least they're explicit.

AsuMagic:
Forgot about this. An alternative to this PR should be worked on soon based on the above comment, probably based on one of these sprites:

resupply draft

eps0003:
> Forgot about this. An alternative to this PR should be worked on soon based on the above comment, probably based on this sprite:
> 
> resupply icon

Who created these sprites? I was unable to find them in the KAG files.

AsuMagic:
Superseded by #1587

Issue #1516, opened 2022-11-25
Crate doesn't autopickup things.


Hobey:
Crates used to pickup logs when you walk over them or if you throw the crate at a log. 

Since Build 4234, crates don't do that anymore.

Reproduce:
- Buy a crate in a buildershop
- spawn a log
- Throw the crate at the log, or walk over it.

Hobey:
It seems that crates actually can't autopickup anything anymore. I just noticed with logs first I guess.

eps0003:
Crates don't pick up anything in all the tagged commits I've tried which makes me suspect the regression was caused by an engine change.

PR #1524, opened 2022-11-27
Warboat crate drops gold when destroyed done properly && fixed crate auto-pickup

bunniewormy:
## Status

## Description

As per mugg's comment on the old PR:

https://github.com/transhumandesign/kag-base/pull/1475

The gold drops for Outpost crate and Warboat crate are declared in two different places.

Gold drop for Outpost crate is added via VehicleShop.as onCommand()
whereas epsilon's PR added Gold drop for Warboat crate in Crate.as onInit().

It also caused an issue with crates not being able to auto-pickup items.
Doing `blob.get_string("packed")` means `blob.exists("packed")` will return true every time after that (which is stupid, but I suppose you're supposed to do an .exists check before getting it every time),
so `crateTake()` in CratePickupCommon.as was returning false due to that, breaking the autopickup.

### DIFF
--- a/Entities/Industry/CTFShops/BoatShop/BoatShop.as
+++ b/Entities/Industry/CTFShops/BoatShop/BoatShop.as
@@ -57,21 +57,21 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
 	if (cmd == this.getCommandID("shop made item"))
 	{
 		this.getSprite().PlaySound("/ChaChing.ogg");
-		bool isServer = (getNet().isServer());
+
 		u16 caller, item;
-		if (!params.saferead_netid(caller) || !params.saferead_netid(item))
+		string name;
+
+		if (!params.saferead_netid(caller) || !params.saferead_netid(item) || !params.saferead_string(name))
 		{
 			return;
 		}
-		string name = params.read_string();
-		{
-			if (name == "warboat")
-			{
-				CBlob@ crate = getBlobByNetworkID(item);
 
-				crate.set_Vec2f("required space", Vec2f(10, 6));
-				crate.set_s32("gold building amount", CTFCosts::warboat_gold);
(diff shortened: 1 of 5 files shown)


Entities/Industry/CTFShops/BoatShop/BoatShop.as
```diff
@@ -57,5 +57,21 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
 	if (cmd == this.getCommandID("shop made item"))
 	{
 		this.getSprite().PlaySound("/ChaChing.ogg");
+		bool isServer = (getNet().isServer());
```
eps0003, line 60: Unused variable.

Entities/Industry/CTFShops/BoatShop/BoatShop.as
```diff
@@ -57,5 +57,21 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
...
+		u16 caller, item;
+		if (!params.saferead_netid(caller) || !params.saferead_netid(item))
+		{
+			return;
+		}
+		string name = params.read_string();
```
eps0003, line 66: saferead should be used as you did previously.

eps0003 (changes requested):

mugg91:
This will resolve #1516.

Entities/Industry/CTFShops/BoatShop/BoatShop.as
bunniewormy: I was just copypasting code from VehicleShop.as so that it looks the same. I've changed all cases of "shop made item" command in all CTF shops now so that they all use saferead.

(also changed getNet().isServer() to isServer() in the files)

AsuMagic (approved):

PR #1477, opened 2022-11-13
Add ability to crouch through crates

eps0003:
## Status

- **READY**: this PR is (to the best of your knowledge) ready to be incorporated into the game.

## Description

Resolves #1117

## Steps to Test or Reproduce

Crouch through crates exactly like with team platforms.

PR #1527, opened 2022-12-01
[fixed] Archer can't clip through wall when crouching and getting in a Crate 

mugg91:
## Status

## Listing

[changed/fixed] 
When getting in crate, if you are holding the crate, the crate assumes the caller's position.
Therefore, Archer hold Crate and crouching before a wall when getting in the crate will not clip you through the wall

[dev] 
Removed unused `CBlob@ mine = null;`.

## Description

Archer holding Crate and getting inside while crouching in front of a wall would occasionally clip you through the wall.
This PR makes it so the Crate will assume the player blob's position when getting in, but only if you are holding the crate.

## Steps to Test or Reproduce

Stand in front of a wall as Archer, hold a crate and crouch, tap the "get in" button.
Notice you will occasionally clip through the wall. **After this PR, not anymore.**

Keep showing buttons while throwing a crate away.
Click on the "get in" button while the crate lies on the ground.
Notice you will get in the Crate and the Crate will stay in the same position. **Unchanged after this PR.**

Place a Mine in a Crate.
Change Team.
Get in the Crate.
Notice that you will not get in and Mine flies out (booby trap). **Unchanged after this PR.**

### DIFF
--- a/Entities/Items/Crate/Crate.as
+++ b/Entities/Items/Crate/Crate.as
@@ -359,7 +359,6 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
 			CInventory@ inv = this.getInventory();
 			u8 itemcount = inv.getItemsCount();
 			// Boobytrap if crate has enemy mine
-			CBlob@ mine = null;
 			for (int i = 0; i < inv.getItemsCount(); i++)
 			{
 				CBlob@ item = inv.getItem(i);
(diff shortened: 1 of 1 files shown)


Gingerbeard5773:
In testing I wasn't able to reproduce the bug that this is supposed to fix

mugg91:
I tested and can't reproduce the issue either, for some reason.
I recall the issue being easy to reproduce. Not sure if the Christmas patch changed something...

AsuMagic:
Could be #1477 if you didn't rebase over master when making the PR

mugg91:
Maybe we should close this for now, at least until I see the exploit arise again.

PR #1533, opened 2022-12-13
Fixed seed placement bug.

### LINKED ISSUE #1514: Can't place seed as builder, if have no enough resources
Builder sometimes can't place seed like building.
Edit: Checked this again, and get regularity

Case 1:
1.Put seed to inventory
2.Select wooden platform(or something blob)
3.Build it to use all wood you have
4.Pickup seed
5.Can't place the seed, dialog that need wood to build is shown
It resets on detach, it is happen even select block or put to inventory again

Case 2:
1.Put seed to inventory
2.Build workshop to get less than 150 wood
3.Pickup seed
4.Same thing is happen

I think we need to check BlobPlacement.as or PlacementCommon.as

Hobey:
## Status
- **READY**
- I have tested locally.
- I have tested multiplayer on a server (with 2 people on it for 5 minutes).

## Description
Resolves #1514.
See comments in ~~[BuilderLogic.as (line 657)](https://github.com/Hobey/kag-base/blob/4c9786c7a95aaac49aa796da5a5ecdd2abf144b5/Entities/Characters/Builder/BuilderLogic.as#L657)~~ [BuilderLogic.as (line 656)](https://github.com/transhumandesign/kag-base/blob/19a821681aae5597949a12bec6aeeeb6a8686a11/Entities/Characters/Builder/BuilderLogic.as#L656)
Seeds can now be placed after running out of mats while building a blob (e.g. wooden door).

## Steps to Test or Reproduce
- Put a (tree) seed in your inventory; (e.g. do "!tree_pine", then kill tree)
- Then you start building blobs (e.g. doors) until you run out ot mats for that blob;
- Then you take the seed out of your inventory
- Then you try to place the seed but you can't because you get a "Requires 30 wood" message;

## Demo
- [Without fix](https://youtu.be/VHu_iz9FrOE)
- [With fix](https://youtu.be/9Nq-5q5r3I0)
- [Fix, but without the 'if' that filters for "seed" blobs only](https://youtu.be/ie7KtTIHEvQ)

## Some useful information

git branch used:  seed_fix
https://github.com/Hobey/kag-base/tree/seed_fix

### DIFF
--- a/Entities/Characters/Builder/BuilderLogic.as
+++ b/Entities/Characters/Builder/BuilderLogic.as
@@ -651,27 +651,10 @@ void onAddToInventory(CBlob@ this, CBlob@ blob)
 
 void onAttach(CBlob@ this, CBlob@ attached, AttachmentPoint @attachedPoint)
 {
-	if (attached is null) return;
-	if (attached.getName().getHash() == 1346747564) // (attached.getName() == "seed")
+	if (attached !is null && attached.getName() == "seed")
 	{
-		// NOTE(hobey): 
-		// See issue: https://github.com/transhumandesign/kag-base/issues/1514
-		//
-		// The issue happened when:
-		//   - Put a (tree) seed in your inventory; (e.g. do "!tree_pine", then kill tree)
-		//   - Then you start building blobs (e.g. doors) until you run out ot mats for that blob;
-		//   - Then you take the seed out of your inventory
-		//   - Then you try to place the seed but you can't because you get a "Requires 30 wood" message;
-		//
-		// Setting the buildblob to -1 on seed-attach fixes this, 
+		// Setting the buildblob to -1 on seed-attach resolves issue #1514, 
 		// since the mat-requirement logic doesn't fire if buildblob is -1.
-		//
 		this.set_u8("buildblob", -1);
-		//
-		// Note that doing this for for all blobs instead of only for "seed" blobs causes "replenishing of build blobs" to not work (i.e. you will place a door but then you don't automatically get a second one in your hand
-		//
-		// https://youtu.be/VHu_iz9FrOE [Without fix]
-		// https://youtu.be/9Nq-5q5r3I0 [With fix]
-		// https://youtu.be/ie7KtTIHEvQ [Fix, but without the 'if' that filters for "seed" blobs only]
 	}
 }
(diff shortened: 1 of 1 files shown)


Entities/Characters/Builder/BuilderLogic.as
```diff
@@ -648,3 +648,30 @@ void onAddToInventory(CBlob@ this, CBlob@ blob)
 		SetHelp(this, "help inventory", "builder", "$Help_Block1$$Swap$$Help_Block2$           $KEY_HOLD$$KEY_F$", "", 3);
 	}
 }
+
+void onAttach(CBlob@ this, CBlob@ attached, AttachmentPoint @attachedPoint)
+{
+	if (attached is null) return;
+	if (attached.getName().getHash() == 1346747564) // (attached.getName() == "seed")
```
eps0003, line 655: There is no need to optimise this by comparing the hash. Just compare the name and combine the two `if` statements into one.

Entities/Characters/Builder/BuilderLogic.as
```diff
@@ -648,3 +648,30 @@ void onAddToInventory(CBlob@ this, CBlob@ blob)
...
+		// Setting the buildblob to -1 on seed-attach fixes this, 
+		// since the mat-requirement logic doesn't fire if buildblob is -1.
+		//
+		this.set_u8("buildblob", -1);
+		//
+		// Note that doing this for for all blobs instead of only for "seed" blobs causes "replenishing of build blobs" to not work (i.e. you will place a door but then you don't automatically get a second one in your hand
+		//
+		// https://youtu.be/VHu_iz9FrOE [Without fix]
+		// https://youtu.be/9Nq-5q5r3I0 [With fix]
+		// https://youtu.be/ie7KtTIHEvQ [Fix, but without the 'if' that filters for "seed" blobs only]
```
eps0003, line 675: Too many comments for a one-line fix. Shorten the comment down to one or two lines max.

eps0003 (changes requested):

eps0003:
Please test on a local dedicated server rather than just localhost. Ask in Discord if you are unsure how to do this.

Hobey:
Did testing on a server with trollbird just now. Fix seems to work. PR is ready now.

AsuMagic:
Bit of a painful workaround :P

I'm not even sure what code in the first place allows seed placement. Would be worth checking if there is a more generic fix to it.

Gingerbeard5773:
There is a more straight forward solution that could be done in BlobPlacement.as at line 298. Add a check to make sure we are using correct requirements for the the carried blob. It would look something like
``if (block !is null && block.name == carryBlob.getName())``
this approach is a lot smarter because we aren't hardcoding the seed's name into the code, and will work for any theoretical blob that can be placed like the seed. Also its a one line fix for this bug. BTW I haven't tested this code extensively... do your own testing if you use this method. @Hobey @AsuMagic

PR #1548, opened 2022-12-29
[fixed] You can click on "change class" buttons from right corners of class shops

### LINKED ISSUE #1505: Can change class at shops from left corners, but not from right corners
## Current in-game behaviour.

You can change class at shops from left corners, but you cannot do it from right corners.

Can select button from left corner - https://imgur.com/nbwWlxu.png
Cannot select button from right corner - https://imgur.com/Ry2brVf.png

## Proposed Change

Ideally, you could make shop buttons selectable depending on distance between player from the center of the shop/thing you want to interact with.

mugg91:
## Status

## Description

Fixes #1505.

This is a simple fix that makes it so you can click the "change class" button from right corner of a class shop.
This was achieved in `OneClassAvailable.as` by increasing `this.get_u8("class button radius");` in `onInit()` which is applied to the button's `enableRadius` in `getButtonsFor()`.
To determine a higher "class button radius", instead of calculating the average of shape with and shape height, I take whatever is higher and add 8 more to it. So the "class button radius" for class shops is now 48 instead of 32.

There is a more sophisticated solution that will enable or disable the class change button dependent on your distance with the shop, but I decided against it for now.
The solution being:
- Add this to each class shop's `onInit()`:
```
	this.getCurrentScript().runFlags |= Script::tick_blob_in_proximity;
	this.getCurrentScript().runProximityTag = "player";
	this.getCurrentScript().runProximityRadius = 8.0f;
```
- Add this for each class shop:
```
void onTick(CBlob@ this)
{
	CButton@ button = getHUD().getButtonWithCommandID(SpawnCmd::changeClass);
	if (button !is null)
	{
		f32 distance = (this.getPosition() - button.getOwner().getPosition()).Length();
		f32 allowed_distance = this.get_u8("class button radius");
		
		button.SetEnabled((distance < allowed_distance) ? true : false);
	}
}
```
- Remove the line `button.enableRadius = this.get_u8("class button radius");` from `GetButtonsFor()` in `OneClassAvailable.as`.

Although I didn't see lag after placing hundreds of Knightshops, I'm still afraid if a ton of proximity checks could drain performance. The solution should also be applied to all buttons, not just "class change" buttons.

## Steps to Test or Reproduce

Build Archershop, Knightshop or Buildershop and place blocks so that you can stand inside the left and right corners without being able to pass through.
Show buttons, you will notice that you can change class from the left corners but not from the right corners. 
**After this PR, you can change class from both sides.**

### DIFF
--- a/Entities/Common/Respawning/OneClassAvailable.as
+++ b/Entities/Common/Respawning/OneClassAvailable.as
@@ -16,7 +16,7 @@ void onInit(CBlob@ this)
 		CShape@ shape = this.getShape();
 		if (shape !is null)
 		{
-			this.set_u8("class button radius", Maths::Max(this.getRadius(), (shape.getWidth() + shape.getHeight()) / 2));
+			this.set_u8("class button radius", Maths::Max(this.getRadius(), Maths::Max(shape.getWidth(), shape.getHeight()) + 8));
 		}
 		else
 		{


Entities/Common/Respawning/OneClassAvailable.as
```diff
@@ -16,7 +16,7 @@ void onInit(CBlob@ this)
 		CShape@ shape = this.getShape();
 		if (shape !is null)
 		{
-			this.set_u8("class button radius", Maths::Max(this.getRadius(), (shape.getWidth() + shape.getHeight()) / 2));
+			this.set_u8("class button radius", Maths::Max(this.getRadius(), Maths::Max(shape.getWidth(), shape.getHeight()) + 8));
```
eps0003, line 19: Use the tile size CMap property rather than the hardcoded `8`.

eps0003 (dismissed):

Entities/Common/Respawning/OneClassAvailable.as
mugg91: I applied your suggestion and also applied it to the line that said 16.
Let me know if there are more suggestions.

AsuMagic (approved):

PR #1552, opened 2022-12-30
[changed] You can replace Wood Doors with Stone Doors / You can repair Spikes

### LINKED ISSUE #1433: Make it possible to replace wood doors with stone doors
## Current in-game behaviour.

You can replace low health wood doors with new wood doors directly.
However, you cannot replace a (full or low health) wood door with a stone door.

## Proposed Change

Make it possible to replace wood doors and low-health stone doors with new stone doors, just like the wood tile counterparts.

### LINKED ISSUE #1530: Can't replace mossy castle tile
## Current in-game behaviour.

Go to a map with mossy tiles or have PR #1465 applied and wait for moss to grow.
Try to replace a mossy background with a normal background; it works.
Try to replace a mossy stone block with a normal stone block; it doesn't work.

## Proposed Change

Make it possible to replace mossy stone block with a normal stone block.

mugg91:
## Status

- **IN DEVELOPMENT**
- Remaining issues:
Repairing Spikes should get rid of blood.
Repairing Spikes will make a "take item" sound instead of "spikes being placed" sound.
- After testing, I found Stone door replacing wood door doesn't work in online.

## Description

[changed] You will be able to replace mossy tile with a normal castle tile - Fixes #1530
[changed] You will be able to place Stone Door on a Wood Door in order to replace it - Fixes #1433
[changed] You will be able to repair Spikes

I made some changes to the Blob Placement logic.
Most notably, instead of having a bunch of `isPlatform`, `isDoor`, `isSpikes`, I add a generic Tag `"repairable"` to `Spikes.as`, `Bridge.as`, `WoodenPlatform.as` and `SwingDoor.as`.

For the stone type of `SwingDoor`, I added Tag `"can replace"` and a strings array `names_to_replace` which is saved to the SwingDoor blob. This array currently only contains `wooden_door`. The same principle could be used in the future for other blobs.

Unfortunately, `PlaceBlob()` in `BlobPlacement.as` seems to be called from the engine-side with parameters such as `repairing`. So I couldn't add my own `replacing` boolean. Instead, I check if overlapping blobs contain a "name to replace" and if yes, save the net id of that blob and check for it in `PlaceBlob()`. If the net id blob exists and the blob placement occurs successfully, then the blob will be replaced (`server_Die()` is applied).

So far, I have tested this only in offline. It needs to be tested in online. Given the complexity of the blob placement logic, it would be good to test this in a larger match setting.

### DIFF
--- a/Entities/Common/Building/BlobPlacement.as
+++ b/Entities/Common/Building/BlobPlacement.as
@@ -16,7 +16,7 @@ bool PlaceBlob(CBlob@ this, CBlob @blob, Vec2f cursorPos, bool repairing = false
 			return false;
 
 		u32 delay = getCurrentBuildDelay(this);
-		SetBuildDelay(this, delay / 2); // Set a smaller delay to compensate for lag/late packets etc
+		SetBuildDelay(this, delay / 2); // Set a smaller delay to compensate for lag/late packets etc.
 
 		CShape@ shape = blob.getShape();
 		shape.server_SetActive(true);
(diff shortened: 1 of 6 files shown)


AsuMagic:
Converted to draft assuming the PR message is up to date

PR #1560, opened 2023-01-05
[boat rebuild] Improved vehicle capturing

Gingerbeard5773:
## Status

- **READY**

## Description

A few minor improvements to vehicle capturing.

* Commit 1: Fixed a bug where the mounted bow on a warship would not change teams.
* Commit 3: Attached vehicles can only be converted by their parent (The vehicle that they are attached to).
* Commit 4: Introduces customizable capture zones, which fixes a problem with outposts being capped outside of walls. More in detail farther down.

## Steps to Test or Reproduce

A: Test to see the warboat's mounted bow change team when the warboat is captured.
B: Attach a catapult to a warboat/longboat. Now try to capture the catapult. You should only see the boat being captured.
C: Place one layer of blocks around an outpost. You shouldn't be able to capture it outside of the blocks.

devs & modders listen:
Capture zones are customizable. For example, here is the zoning I've added for boats;

Capture Zones

The zones may seem a bit small, but capturing is triggered when you touch the zone, not when you are all the way inside. You can see zones in game by using ``/g_debug 2`` in your console.

If a blob's script doesn't define a capture zone, its default capture zone will be based off its radius.

### DIFF
--- a/Entities/Vehicles/Common/VehicleConvert.as
+++ b/Entities/Vehicles/Common/VehicleConvert.as
@@ -4,18 +4,22 @@
 
 const string counter_prop = "capture ticks";
 const string raid_tag = "under raid";
+
+const string capture_time_prop = "capture time";
 const int capture_half_seconds = 30;
-const int short_capture_half_seconds = 10;
 
 const string friendly_prop = "capture friendly count";
 const string enemy_prop = "capture enemy count";
-
-const string short_raid_tag = "short raid time";
 
 void onInit(CBlob@ this)
 {
 	this.getCurrentScript().tickFrequency = 15;
 	this.getCurrentScript().runFlags |= Script::tick_not_attached;
+
+	if (!this.exists(capture_time_prop))
+	{
+		this.set_u16(capture_time_prop, capture_half_seconds);
+	}
 
 	if (isServer())
 	{
@@ -36,7 +40,7 @@ void ResetProperties(CBlob@ this)
 {
 	this.set_u16(friendly_prop, 0);
 	this.set_u16(enemy_prop, 0);
-	this.set_u16(counter_prop, GetCaptureTime(this));
+	this.set_u16(counter_prop, this.get_u16(capture_time_prop));
 	this.Untag(raid_tag);
 }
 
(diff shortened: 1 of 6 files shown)


mugg91:
It seems your commit 1 and 3 are already included in #1251 amongs lot of other fixes.
`[fixed] Bow on Warboat now gets converted along with the Warboat.`
`[fixed] Converting a Vehicle that is attached to another vehicle will detach instead of converting the other vehicle.`

Your commit 4 is interesting, that should be used on outpost to prevent capture through 1 block tall ceilings.

Gingerbeard5773:
``Without having tested it yet, if there is a red warboat with a red catapult on it, and the warboat gets captured by blue.
Does it instantly convert the warboat and catapult to blue?``

Yes. Here's a before and after comparison of what happens. The purpose of commit 3 is to remove pointless HUD clutter as seen in the first video. There is no gameplay difference.

https://user-images.githubusercontent.com/68350259/210921832-cb3a205b-049d-4668-b81e-cdac2209d96d.mp4

https://user-images.githubusercontent.com/68350259/210921824-05c847b4-fdca-45d6-ad15-e2023645e9eb.mp4

Entities/Vehicles/Common/VehicleConvert.as
```diff
@@ -38,145 +73,110 @@ void onAttach(CBlob@ this, CBlob@ attached, AttachmentPoint @attachedPoint)
...
 			sprite.PlaySound("/VehicleCapture");
 		}
-
-		ConvertPoint(this, "VEHICLE");
-		ConvertPoint(this, "DOOR");
 	}
 }
 
-void ConvertPoint(CBlob@ this, const string pointName)
+void ConvertPoints(CBlob@ this, const string pointNames)
```
AsuMagic, line 161: `pointNames` really should be a `string[]`

AsuMagic (changes requested):

AsuMagic:
Also, it looks like dinghies could benefit from having a custom capture zone.

AsuMagic:
One interesting thing is:
image
Once my hitbox touches the capture zone, it correctly starts capturing. Once I leave it by a few pixels, it doesn't stop. I have to walk away further for capturing to stop. It likely has to do with a bug/weird intentional behavior with moving sectors in engine rather than code in this PR though. It's minor enough not to be an issue.

Entities/Vehicles/Common/VehicleConvert.as
Gingerbeard5773: Resolved, review again? :D

AsuMagic (approved):

PR #1562, opened 2023-01-13
[fixed] No materials on next map

### LINKED ISSUE #1344: No resupplies when starting a new match

Gingerbeard5773:
## Status

- **READY**

## Description
Resolves #1344 & closes PR #1526

This PR decisively fixes the bug where players would spawn on next map without receiving their materials.

Why did the bug occur?
When going next map, players would be auto-balanced, and will change teams. Changing a player's team causes the player to request a new spawn. This kills their blob and spawns a new one. However, players receive their spawn mats before the second spawn happens, and they end up dropping their materials when they die.

TLDR: Players changing teams on next map causes them to drop their mats.

## Steps to Test or Reproduce

a) Go onto a dedicated server.
b) Spawn at least 3 bots with ``!bot``. You will probably need moderator/rcon access to use the chat command.
c) Keep going next map. If you are testing *without* the applied fix you will notice that you will receive no mats every time you switch teams by auto balance. Otherwise, with the applied fix you will receive mats as usual.

### DIFF
--- a/Rules/CTF/Scripts/CTF.as
+++ b/Rules/CTF/Scripts/CTF.as
@@ -184,6 +184,8 @@ shared class CTFSpawns : RespawnSystem
 
 		if (info is null) { warn("CTF LOGIC: Couldn't get player info ( in bool canSpawnPlayer(PlayerInfo@ p_info) ) "); return false; }
 
+		if (getGameTime() < 15) { return false; } //allow time for team balancing
+
 		if (force) { return true; }
 
 		return info.can_spawn_time <= 0;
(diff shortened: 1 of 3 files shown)


AsuMagic:
.. That explains things :)

Why does the team swap logic kick in *after* that players have spawned? Could we reliably trigger it earlier without the game time check hack?

Gingerbeard5773:
Its caused due to script priority in the gamemode config. DefaultBalanceTeams.as is triggered after CTF.as. I tried reordering scripts but it didn't work. The whole system is really sensitive and I didnt want to risk causing more bugs, which is why I ended up with a hack.

AsuMagic (approved):
:(

PR #1563, opened 2023-01-16
[fixed] Incorrect saw sprite on player join

### LINKED ISSUE #1378: Disabled Saw has a wrong sprite when joining an online server
## Description

Sometimes a Saw will look as though it is active but without animation.

## Reproduction

It seems this occurs when you join a server that has a disabled Saw.
The Saw will look active but not be animated at all.

Gingerbeard5773:
## Status

- **READY**

## Description

Resolves #1378 & Closes #1395 

The difference between this PR and #1395 is that the saws will also synchronize the saw blade's bloody state. Sync is also done just once instead of every tick.

## Steps to Test or Reproduce
1) Go onto a dedicated server.
2) Create a saw, jump in it to make it bloody, then turn it off.
3) Disconnect and then rejoin the server. With the applied fix, the saw should look exactly the same as before.

Note for reviewers:
Commit 1 is the necessary changes, Commit 2 has no functional change.


Gingerbeard5773:
failure

PR #1569, opened 2023-01-22
[modified] Update tunnel menu to display options in consistent order

### LINKED ISSUE #1568: Use total ordering of tunnels in travel menu, instead of relative distance-based ordering

The order that tunnels appear in the travel menu is inconsistent, depending on which tunnel you're traveling from. This is unintuitive and makes it hard to describe a given tunnel (i.e. you can't say "third tunnel under attack" without ambiguity).

As a demonstration, here's a little puzzle. Consider the following arrangement of tunnels, with tunnels labeled 1-6: 

The goal is to visit each of the tunnels once in sequential order, using only tunnel warps. (i.e. warp to tunnel 1, then warp from 1 to 2, then 2 to 3, etc.) 

I've labeled the buttons of the menu from left-to-right with A-F. What sequence of button presses would allow you to achieve this goal? (I covered the icons to emphasize that the labels are for the _positions_ of the buttons in the menu.)

Video solution: 

https://user-images.githubusercontent.com/22576116/213889978-c3ef009d-f3d0-48e3-b8ae-14861be38a3e.mp4

Did anyone guess A, D, C, C, D, D? 

Even if you did, you'd _still_ be wrong, because it actually depends on the order that the tunnels were created! Here's a video of the same setup, except the answer is now A, D, C, **D**, D, D: 

https://user-images.githubusercontent.com/22576116/213890042-1ddbd044-e6ca-45a2-bae7-e3a72e3777a5.mp4

There are two main cases where inconsistencies happen:
1. Multiple tunnels placed on the same x-coordinate. 
    - If you try to travel from one of these tunnels, the other tunnels on the same x-coordinate will always appear on the right side of the current one.
    - If there are three vertically aligned tunnels that are equally spaced apart, the button order of the middle tunnel depends on whether the top tunnel or bottom tunnel was built first.
2. When there is one tunnel that is farther along the x-axis than another tunnel, but still closer according to euclidean distance (because the other tunnel is higher or lower vertically).
    - Video demo: [https://youtu.be/CcjNL2XTZjU](https://youtu.be/CcjNL2XTZjU)
 
**Proposed Change**

The order of tunnels in the travel menu should follow a universal total ordering based on their x- and y-coordinates. (First sort by `x`, then by `y` for ties in `x`.)

This ordering will be more intuitive for players. In the above puzzle, you could visit each tunnel 1-6, just by clicking buttons A-F. And in general, you will always be able to map one given button in the menu to one given tunnel in the map. This will greatly reduce the cognitive load of checking the direction of the arrow and figuring out which tunnel is further than another.

**Additional context**

 - These inconsistent cases may seem niche, but they happen surprisingly often in maps with multiple routes (top/bottom) or in very tall towers. Some offensive builders also like to create "backup" tunnels below existing ones if they have an excess of gold.
 - Any implementation of issue #1190 would have a pretty terrible UX without this change.
 - There may be some discussion for whether to flip the vertical sorting for red/blue team.
 - Relevant code: [https://github.com/transhumandesign/kag-base/blob/master/Entities/Industry/Tunnel/TunnelTravel.as#L55](https://github.com/transhumandesign/kag-base/blob/master/Entities/Industry/Tunnel/TunnelTravel.as#L55)
 - Of course, everything here applies to outposts and halls as well as tunnels.

ParaLogia:
## Status

- **READY**

## Description

Resolves #1568 by sorting tunnels according to map position (x-coordinate, then y-coordinate) instead of sorting by distance from the current tunnel.

## Steps to Test or Reproduce

Create a series of tunnels/halls/outposts at different elevations, and try to visit each tunnel. You should be able to visit a given tunnel by clicking the same button position (e.g. 2nd from left) consistently, no matter which tunnel you're traveling from.

Example video:

https://user-images.githubusercontent.com/22576116/213897695-50560e17-c67c-4ee8-893e-8256e6bf498e.mp4

Note: I mentioned in the issue that we may wish to flip the vertical sorting for red team (i.e. bottom-to-top instead of top-to-bottom). I did not implement that in this PR.

### DIFF
--- a/Entities/Industry/Tunnel/TunnelTravel.as
+++ b/Entities/Industry/Tunnel/TunnelTravel.as
@@ -5,6 +5,28 @@
 #include "KnockedCommon.as";
 #include "GenericButtonCommon.as";
 
+// Wrapper class that overrides opCmp to allow sorting by position
+class PositionComparableCBlobWrapper 
+{
+	CBlob@ blob;
+
+	PositionComparableCBlobWrapper(CBlob@ p_blob)
+	{
+		@blob = p_blob;
+	}
+
+	// Compare by x-coordinate, then y-coordinate
+	int opCmp(PositionComparableCBlobWrapper@ other)
+	{
+		int xDiff = blob.getPosition().x - other.blob.getPosition().x;
+		if (xDiff != 0)
+		{
+			return xDiff;
+		}
+		return blob.getPosition().y - other.blob.getPosition().y;
+	}
+}
+
 void onInit(CBlob@ this)
 {
 	this.addCommandID("travel");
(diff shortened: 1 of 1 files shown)


Gingerbeard5773 (approved):
Tested and found no issues. Good change 👍

AsuMagic (approved):

PR #1571, opened 2023-01-24
[fixed] Seed placement using requirements

### LINKED ISSUE #1514: Can't place seed as builder, if have no enough resources

Gingerbeard5773:
## Status

- **READY**

## Description

Closes #1533 and resolves #1514.

An alternative method.
## Steps to Test or Reproduce

1) Go on localhost or dedicated server.
2) Build blobs until you see "missing requirements".
3) Grab a seed out of your inventory.
4) Attempt to place the seed. Without the applied fix, it will be not be possible.

Credit given to Hobey

### DIFF
--- a/Entities/Common/Building/BlobPlacement.as
+++ b/Entities/Common/Building/BlobPlacement.as
@@ -295,7 +295,7 @@ void onTick(CBlob@ this)
 
 	u8 blockIndex = this.get_u8("buildblob");
 	BuildBlock @block = getBlockByIndex(this, blockIndex);
-	if (block !is null) {
+	if (block !is null && block.name == carryBlob.getName()) {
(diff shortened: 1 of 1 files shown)


AsuMagic (approved):
Tested on net with high latency, doesn't seem to cause issues with block selection

PR #1580, opened 2023-01-29
[dev] Simplified CTF flag scripts

### LINKED ISSUE #1319: CTF flag timer bugs

### LINKED ISSUE #1604: Map-prebuilt background tiles behind flags get destroyed (except dirt)
**Description**

If you add wooden or stone background tiles behind where a flag base would be in a map `.png`, then the background tiles get turned into air. This probably is caused by the code that also places bedrock tiles beneath flag_base.

**Reproduction**

Load CTF map `Asu_Kelsier_v2`

**Expected Behavior**

Flag backwall should remain

**Screenshots**

image
image

**Your Setup:**

N/A

**Additional context**

N/A

Gingerbeard5773:
## Status

- **READY**

## Description

This PR includes #1325 and will close it. Resolves #1319, resolves #1604

## Steps to Test or Reproduce

Test as instructed in mentioned issues above.

### DIFF
--- a/Entities/Meta/CTFMusic.as
+++ b/Entities/Meta/CTFMusic.as
@@ -217,7 +217,7 @@ void GameMusicLogic(CBlob@ this, CMixer@ mixer)
 					for (uint i = 0; i < flagBases.length; i++)
 					{
 						CBlob @b = flagBases[i];
-						if (b.hasTag("flag missing"))
+						if (!b.hasAttached())
 						{
 							chosen = world_battle;
 							break;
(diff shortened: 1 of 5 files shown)


AsuMagic:
Removed the NFC tag as this actually includes a bugfix

AsuMagic (approved):

PR #1582, opened 2023-02-01
[dev] Costs optimization

Gingerbeard5773:
## Status

- **READY**

## Description

As suggested by @Vam-Jam in #1370

Optimizations
* Costs.cfg is read only once. The values are stored in a dictionary in rules.
* Other scripts will read from the dictionary. When those scripts receive the values, they do not read costs ever again.

## Steps to Test or Reproduce
No intended functional difference.

Though, will this even make a beneficial difference enough to be important? Let me know

### DIFF
--- a/Entities/Common/Includes/Costs.as
+++ b/Entities/Common/Includes/Costs.as
@@ -4,6 +4,8 @@
 //	the values are set in InitCosts,
 //	to prevent them from lingering between
 
+bool costs_loaded = false;
+
 //// CTF COSTS ////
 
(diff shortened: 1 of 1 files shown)


AsuMagic:
> Though, will this even make a beneficial difference enough to be important? Let me know

I have never seen it in profiling but avoiding to reload config files is probably not a bad idea.  
That being said I doubt it really ever translates to a significant performance penalty in the real world because the file is probably never going to be evicted from page cache.

AsuMagic (approved):

PR #1583, opened 2023-02-01
[changed] Blobs on fire aren't allowed to be put into inventories

### LINKED ISSUE #1401: Resolve in-inventory burning

Gingerbeard5773:
## Status

- **READY**

## Description

Closes #1521 & resolves #1401

## Steps to Test or Reproduce

1) Burn a drill, log or grain.
2) Attempt to place it into your inventory. With the applied change, it will not be possible until the blob stops burning.

### DIFF
--- a/Entities/Common/FireScripts/IsFlammable.as
+++ b/Entities/Common/FireScripts/IsFlammable.as
@@ -119,3 +119,8 @@ void onTick(CBlob@ this)
 
 	// (flax roof cottages!)
 }
+
+bool canBePutInInventory(CBlob@ this, CBlob@ inventoryBlob)
+{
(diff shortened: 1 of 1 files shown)


AsuMagic (approved):
Will see if people like it in testing

mugg91:
I think this also prevents players from getting inside crates while burning.
Which is not what the original issue/my PR wanted to address or implement.
I vaguely remember that it used to be possible.
Maybe we want to consider allowing again, getting inside crates while burning.

PR #1585, opened 2023-02-01
Slightly optimize `RegrowPlants`' `findTileByCoords`

AsuMagic:
## Status

## Description

Partly mitigates #1354

Profiling shows that object copies within the function is a significant performance hit, which the AS optimizer cannot/is not able to elide.  
This is not a complete fix but a quick improvement because the function's complexity is still `O(n)` with regards to the tile count and should be solved to close the issue.

## Steps to Test or Reproduce

See issue.


ParaLogia (commented):
LGTM

PR #1588, opened 2023-02-04
[fixed] Allow slashing/jabbing team corpses and disallow hitting held corpses

### LINKED ISSUE #1578: Can't slash/jab corpses of your own team
**Description**

Knights can't slash/jab corpses of their own team to get rid of them. This is a problem because it's common for your teammates' corpses to hold your doors open, and it's difficult to clear them out now.

**Reproduction**

1. Join any gamemode and go knight
2. Find a corpse that belongs to your team
3. Try to slash or jab the corpse to damage it
4. The corpse is unaffected

**Expected Behavior**

The corpse should be damaged by your slashes/jabs.

**Video**

https://user-images.githubusercontent.com/22576116/215299279-14094b0b-062c-4506-a3a8-aa46eefb9530.mp4

ParaLogia:
## Status

- **READY**

## Description

Resolves #1578 by re-enabling the ability to slash/jab your teammates' corpses.
Also prevents held corpses from getting hit - a missed case from #1083/#1102.

## Steps to Test or Reproduce

1. Set up a dedicated server
2. Join with at least two players/clients
3. Kill someone to create a corpse
4. Ensure that the corpse can be damaged by slashing/jabbing from knights of any team
5. Have a player pick up the corpse
6. Ensure that the nobody can slash/jab the held corpse, regardless of the team of the holder/attacker/corpse

### DIFF
--- a/Entities/Characters/Scripts/RunnerDeath.as
+++ b/Entities/Characters/Scripts/RunnerDeath.as
@@ -40,12 +40,20 @@ f32 onHit(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@ hitt
 			}
 		}
 
-		// add pickup attachment so we can pickup body
+		// add/convert pickup attachment so we can pickup body
 		CAttachment@ a = this.getAttachments();
 
 		if (a !is null)
 		{
-			AttachmentPoint@ ap = a.AddAttachmentPoint("PICKUP", false);
+			AttachmentPoint@ point = a.getAttachmentPointByName("PICKUP");
+			if (point is null)
+			{
+				AttachmentPoint@ ap = a.AddAttachmentPoint("PICKUP", false);
+			}
+			else
+			{
+				point.socket = false;
+			}
 		}
 
 		// sound
--- a/Entities/Characters/Scripts/RunnerDeath.as
+++ b/Entities/Characters/Scripts/RunnerDeath.as
@@ -40,20 +40,19 @@ f32 onHit(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@ hitt
 			}
 		}
 
-		// add/convert pickup attachment so we can pickup body
+		// add pickup attachment so we can pickup body
 		CAttachment@ a = this.getAttachments();
 
 		if (a !is null)
 		{
 			AttachmentPoint@ point = a.getAttachmentPointByName("PICKUP");
-			if (point is null)
+			if (point !is null && point.socket)
 			{
-				AttachmentPoint@ ap = a.AddAttachmentPoint("PICKUP", false);
+				// Invalidate existing pickup attachment
+				point.name = "PICKUP_OLD";
 			}
-			else
-			{
-				point.socket = false;
-			}
+
+			AttachmentPoint@ ap = a.AddAttachmentPoint("PICKUP", false);
 		}
 
 		// sound
--- a/Entities/Characters/Scripts/RunnerDeath.as
+++ b/Entities/Characters/Scripts/RunnerDeath.as
@@ -45,13 +45,6 @@ f32 onHit(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@ hitt
 
 		if (a !is null)
 		{
-			AttachmentPoint@ point = a.getAttachmentPointByName("PICKUP");
-			if (point !is null && point.socket)
-			{
-				// Invalidate existing pickup attachment
-				point.name = "PICKUP_OLD";
-			}
-
 			AttachmentPoint@ ap = a.AddAttachmentPoint("PICKUP", false);
 		}
 
--- a/Entities/Characters/Knight/KnightLogic.as
+++ b/Entities/Characters/Knight/KnightLogic.as
@@ -1742,24 +1742,23 @@ void SetFirstAvailableBomb(CBlob@ this)
 // Blame Fuzzle.
 bool canHit(CBlob@ this, CBlob@ b)
 {
-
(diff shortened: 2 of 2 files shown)


Entities/Characters/Scripts/RunnerDeath.as
```diff
@@ -40,12 +40,20 @@ f32 onHit(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@ hitt
...
 		}
 
-		// add pickup attachment so we can pickup body
+		// add/convert pickup attachment so we can pickup body
 		CAttachment@ a = this.getAttachments();
 
 		if (a !is null)
 		{
-			AttachmentPoint@ ap = a.AddAttachmentPoint("PICKUP", false);
+			AttachmentPoint@ point = a.getAttachmentPointByName("PICKUP");
```
AsuMagic, line 48: This part of the logic seems to cause a change in the offset of picked up corpses:

Before:  
before image

After:  
after image

This is probably undesirable as it reduces the usefulness of corpses for hiding your team color.

AsuMagic (changes requested):

Entities/Characters/Scripts/RunnerDeath.as
ParaLogia: Is there a method to delete or move the offset an existing AttachmentPoint? 

I believe the issue with the previous code was that it was creating duplicate "PICKUP" AttachmentPoints, so the existing "socket" AttachmentPoint was throwing off this check: https://github.com/transhumandesign/kag-base/blob/8e936d2ebec1fa126213b6e8b185baca07a605ed/Entities/Characters/Knight/KnightLogic.as#L1754

But maybe I'm just misunderstanding how AddAttachmentPoint works. Is there any documentation for methods (with parameters) on these classes? There's nothing about it in [http://doc.kag2d.com/](http://doc.kag2d.com/)

ParaLogia: I figured out how to update the offset to `(0,0)`, but that _still_ had a regression with builder corpses because builders already have two attachment points defined in `Builder.cfg`, one as a socket and one as a plug - for building I guess. 
https://github.com/transhumandesign/kag-base/blob/8e936d2ebec1fa126213b6e8b185baca07a605ed/Entities/Characters/Builder/Builder.cfg#L209

That plug is used when you pick up the corpse, instead of the newly added one from `RunnerDeath.as`. And since the existing plug is offset vertically, it actually covers more of the player than a new `(0,0)` plug.

So for now, I'm "deleting" the existing pickup socket by re-naming it. Please let me know if there's a better way to remove it.

AsuMagic: Oh, I didn't realize why the double `PICKUP` attachment problem was even a thing.

I don't think you can remove attachment points. I'd be wary of any side effects of being able to remove an object like this (such as questionable indexing or sneaky dangling references). The engine code seems to index into the internal attachment list by index in a few places, so removing attachment points would seem risky to me either way.

The plug vs socket case is really annoying. It seems like a flaw in the original API that it is possible to query a single attachment point by name. Searching for `getAttachmentPointByName("PICKUP")` in Base reveals that this `KnightLogic.as` line you linked is seemingly the only one that might be looking for a socket/plug pickup attachment point on a player blob.

Using `getAttachmentPoints` and manually looking for a plug `PICKUP` point (while commenting why that is necessary) might be a cleaner way to achieve this.

ParaLogia: > Using `getAttachmentPoints` and manually looking for a plug `PICKUP` point (while commenting why that is necessary) might be a cleaner way to achieve this.

The trouble I'm getting with this approach is that a builder holding an item can't be slashed/jabbed. That's because `isAttachedToPoint()` takes a string name and not an `AttachmentPoint`, so I can't check if the target is specifically attached to the plug `PICKUP` point rather than the socket. Is there a better method for that?

And that takes me back to my earlier question: Is there a way to check the methods/properties of a class like `AttachmentPoint`? Do I need to use reflection (if that's even a thing)?

AsuMagic: > And that takes me back to my earlier question: Is there a way to check the methods/properties of a class like AttachmentPoint? Do I need to use reflection (if that's even a thing)?

`Manual/Interface/` in the KAG directory. For class methods, either in `Objects.txt` or `Objects/`. Those are only the function signatures though, there is no documentation (other than `doc.kag2d.com`).

Also, after looking stuff up, I noticed that there _is_ a method to `CAttachment` get an attachment point specifically of a desired socket/plug type:

```
	AttachmentPoint@ getAttachmentPoint(const string&in)
	AttachmentPoint@ getAttachmentPoint(const string&in, bool socket)
	AttachmentPoint@ getAttachmentPoint(const string&in, bool socket, bool occupied)
```

There is also `getAttachmentWithBlob(CBlob@ blob)`. So you could

1. Check that the hit candidate is a `PICKUP`, occupied plug
2. Check the name for the attachment with a common blob to ensure that its name is `PICKUP`

Sprinkled with some sanity null checks just in case.

ParaLogia: > 2. Check the name for the attachment with a common blob to ensure that its name is `PICKUP`

I don't understand what you mean by this.

Pushed a new version that uses `getAttachmentPoint(string, bool, bool)`, which seems to work. It seems like the PICKUP plug defined in Builder.cfg doesn't actually get occupied when you select blocks.

I also added a check for the `"player"` tag so that players carried by gregs can still be hit. (This was previously working by accident for the same reason that held corpses could get slashed.)

AsuMagic (approved):

PR #1590, opened 2023-02-05
[fixed] Health bar not scaling with camera zoom

Gingerbeard5773:
## Status

- **READY**

## Description

* Fixes minor visual bug where the health bar moves far away from it's blob when the camera is zoomed out

## Steps to Test or Reproduce

Spawn a vehicle/outpost and zoom out your camera. With the fix the health bar won't move too far away from the vehicle.


ParaLogia (approved):
Tested in sandbox with various vehicles/animals/outposts.

AsuMagic (approved):

PR #1594, opened 2023-02-08
[boat rebuild] Vehicle scripts cleanup

Gingerbeard5773:
## Status
- **READY**
## Description
Vehicle scripts are simplified. Some other improvements were added along the way.

* [fixed] desync where firing sounds wouldn't play.
* [fixed] certain vehicle sprites not showing their correct damage frame.
* [fixed] boats not colliding with platforms.
* [changed] pop off wheels button doesn't appear when the vehicle is attached.

## Steps to Test or Reproduce
All vehicles' functionality should be tested thoroughly.

## Updating mods
Modders pay attention! Assuming this PR is merged, some mods may go out of date. Updating your mod may be neccessary if it modifies vehicle scripts or uses vehicle scripts. It is recommended to update because of bug fixes included.
**How to update:**
The functions ``Vehicle_onFire`` and ``Vehicle_canFire`` are deprecated. They have been moved into each vehicles class. You can view the implementation here at ``catapult.as`` or ``ballista.as``. This change was done so vehicle scripts don't REQUIRE the functions unlike before.
There are other functions and class members that are also deprecated. If your mod uses them, errors will be raised.
If needed, message me on discord if you need help updating your mod.

### DIFF
--- a/Entities/Structures/MountedBow/MountedBow.as
+++ b/Entities/Structures/MountedBow/MountedBow.as
@@ -3,21 +3,41 @@
 
 // Mounted Bow logic
 
-const Vec2f arm_offset = Vec2f(-6, 0);
+class MountedBowInfo : VehicleInfo
+{
+	void onFire(CBlob@ this, CBlob@ bullet, const u16 &in fired_charge)
+	{
+		if (bullet !is null)
+		{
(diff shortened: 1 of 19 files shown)


AsuMagic (approved):
Reviewed and tested to the extent that my brain allowed me to, generally LGTM and will need a bunch of testing

PR #1598, opened 2023-02-14
[modified] Kick AFK to spectator when seeding servers

### LINKED ISSUE #1398: AFK kick should kick to spectator when seeding servers
**Reproduction**

1. Join an empty server
2. Wait for players
3. Be AFK kicked

**Expected Behavior**

Spectators are still counted towards the total player count, so when player count is less than or equal to a certain threshold (probably 1 player, or something like 4) in an appropriate gamemode (e.g. not sandbox), AFK kick could be replaced to a kick to the spectator team.  
This should make seeding easier, as the AFK player(s) are probably waiting for players to come, while avoiding letting _actually_ AFK players stay in a team.

To avoid accidental AFK still, an actual kick could still be applied with a much more generous grace period (e.g. an hour or several hours).

Gingerbeard5773:
## Status

- **READY**

## Description

Resolves #1398 

Implementation as follows:
* AFK Players will be sent into spectator if the server is below half-capacity (threshold subject to change).
* AFK players will be kicked off servers regardless of player count on Sandbox and Challenge game modes.

## Steps to Test or Reproduce
1) Go onto a dedicated server.
2) Wait to be kicked. If the player count is below 50% capacity then you will be sent into spectator.
3) Use bots to simulate AFK kicking over 50% capacity.

### DIFF
--- a/Rules/CommonScripts/KickAFK.as
+++ b/Rules/CommonScripts/KickAFK.as
@@ -1,6 +1,7 @@
 // kicks players that dont play for a given time
 // by norill
 // move people with kick immunity to spec instead of doing nothing -mazey
+// move people into spec if seeding servers -gingerbeard
 
 #define CLIENT_ONLY
 #include "AdminLogic.as"
@@ -13,6 +14,9 @@ const uint checkInterval = 90;
 const uint totalToKickSeconds = 60 * 2 + 30;
 const uint warnToKickSeconds = 60;
 const uint idleToWarnSeconds = totalToKickSeconds - warnToKickSeconds;
+const f32 seedingPercent = 0.5f;
+
+const string[] ignoredSeedingGamemodes = { "Sandbox", "Challenge" };
 
 void onTick(CRules@ this)
 {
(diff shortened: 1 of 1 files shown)


Rules/CommonScripts/KickAFK.as
```diff
@@ -68,17 +72,21 @@ void onTick(CRules@ this)
...
-		if (!kickImmunity)
+		if (!kickToSpectator)
 		{
 			getNet().DisconnectClient();
 		}
-		else
+		else if (kickImmunity)
 		{
-			joinNewSpecTeam(this, getLocalPlayer()); //Force-swap to spec team.
+			joinNewSpecTeam(this, p); //Force-swap to spec team.
```
ParaLogia, line 83: Does this need to be kept? The current way where admins need to type `!m` to rejoin their old team is annoying and the chat message just seems like a hacky workaround.

Gingerbeard5773: All it does is put admins into moderator mode when afk kicked. Otherwise, it would just send admins into spectator without moderator visuals. I'm not sure which is preferred

>The current way where admins need to type !m to rejoin their old team is annoying

Changes to adminlogic should be done in another PR. put up an issue if you think its important

ParaLogia: My opinion is that there's no point in putting admins into moderator mode when they go AFK. 

But looking through the history, it looks like that was added so that players could be swapped to spec even if the game mode doesn't have a spec team. (#460)

I guess that raises the question of whether that issue will re-appear with this PR. Maybe instead of a blacklist a la `ignoredSeedingGamemodes`, there should be a whitelist of "seed-able" game modes?

ParaLogia: The concern wasn't with admins specifically, but with what happens if you try to swap any player to spec when there is no spec team. From testing in sandbox, it seems like it just respawns the player. Is that desirable? Not sure what might happen in mods.

It's likely a nonissue. Just floating some points of potential discussion.

AsuMagic (approved):

PR #1619, opened 2023-04-10
Reduce occurences of falldamage caused by trampolines

asumagic:
## Status

## Description

Even when `bullet` mode (aka CCD) is enabled, players falling fast onto trampolines will still get fall stun due to the fact that Box2D is not aware of how trampolines work, as they behave like a non-solid collision. Within the same tick, the player might hit both the trampoline and the ground.

This PR introduces a somewhat hacky workaround which makes it so that `doesCollideWithBlob` tags the player not to get fall damage this tick, should we expect the blob to be bounced.

It would be possible to make it to that the trampoline is solid to those blobs instead, however, it seems to make it so that the trampoline is pushed back by the collision (which is undesired), and there are probably other unintended side effects.

## Steps to Test or Reproduce

**N/A** until physics `bullet` mode is properly fixed, so you cannot try this ATM until this is changed.

Assuming a build is available with `bullet` mode unconditionally enabled for all entities (or until a PR is merged to introduce this to player and other entities):

1. Test both localhost and a local server
2. Place a trampoline
3. Try to fall very fast on the trampoline and in various ways, you shouldn't get stun
4. Press down while falling: you should still get damage

### DIFF
--- a/Entities/Characters/Scripts/RunnerDefault.as
+++ b/Entities/Characters/Scripts/RunnerDefault.as
@@ -6,6 +6,8 @@
 
 void onInit(CBlob@ this)
 {
+	this.getShape().getConsts().bullet = true;
+
 	this.getCurrentScript().removeIfTag = "dead";
 	this.Tag("medium weight");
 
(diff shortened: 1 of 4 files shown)


Gingerbeard5773 (changes requested):
https://user-images.githubusercontent.com/68350259/232255311-14f16f42-1c13-49f8-a79e-730acc812ac9.mp4

bug with multi tramp collisions

PR #1630, opened 2023-05-04
[modified] flipping siege vehicles requires capturing them first

asumagic:
## Status

## Description

Makes it so you can only flip siege vehicles (catapult, ballista) when you are in the same team as the vehicle. You get crushed when that happens.

Supersedes the implementation of that feature from #1273.

This was quickly hacked together, maybe we should instead take steps so that flipping siege can't crush a player in the first place instead. I don't think this has massive gameplay implications, though.

## Steps to Test or Reproduce

Attempt to flip a catapult or ballista when in a different team - you should only be able to load ammo. Switch teams, and try again.

### DIFF
--- a/Entities/Vehicles/Ballista/Ballista.as
+++ b/Entities/Vehicles/Ballista/Ballista.as
@@ -226,12 +226,15 @@ void GetButtonsFor(CBlob@ this, CBlob@ caller)
 {
 	if (!canSeeButtons(this, caller)) return;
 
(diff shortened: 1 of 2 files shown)


Gingerbeard5773:
Collides with #1594. maybe put this into it?

AsuMagic:
Can always merge #1594 first then rebase (if we don't scrap this flip button team one in the first place)

PR #1632, opened 2023-05-05
Use `getActiveCommandPlayer` in shop logic

AsuMagic:
## Status

## Description

When the `caller` blob occasionally had a `null` player, then the shop script would crash. We could just add a null player check, but we might as well not read the caller from the stream. Getting the player from the netcode is more robust.

This also makes it so that the command handler early returns if `!isServer`. It should not be the case that all players receive the `"shop buy"` command...

## Steps to Test or Reproduce

Test shops on dedicated servers for robustness. Test corner cases.


AsuMagic:
Was checked externally and seemed ok; will see in testing.

Issue #1498, opened 2022-11-16
Animation speed of Outpost is inconsistent


mugg91:
## Current in-game behaviour.

Outpost's flag moves slower than Ballista's flag and Warboat's flag.
Also Outpost's flag will sometimes flicker visually when the daytime starts turning dark.

https://www.youtube.com/watch?v=-dxySPebEPs

## Proposed Change

Make all flags move at the same speed.
Use either the slow or the fast variant.
Look into why the Outpost's flag flickers and try finding a fix.

mugg91:
I checked in online server. There still seems to be an issue with the flag animation speed.
I spawn two warboats or two outposts next to each other. One of them will have a fast animation speed (~18 frames) and the other will have a slower one (~24 frames).
https://imgur.com/YeJhwzY

PR #1641, opened 2023-05-15
Make flag animations consistent

its-kos:
## Status

## Description

This PR addresses [issue#1498](https://github.com/transhumandesign/kag-base/issues/1498). Flag animation speeds have been made consistent and based on the `CTF_FLAG` blob.

## Steps to Test or Reproduce

!spawn Outpost
!spawn Ballista
!spawn Warboat

In sandbox to view their flags.

- All gamemodes relevant
- Objects affected are: Outpost, Ballista, Warboat


asumagic (approved):

mugg91:
I pointed out [here](https://github.com/transhumandesign/kag-base/issues/1498#issuecomment-1765384302) that the flag animation speed is still inconsistent. Yes, all blobs that have a flag will use the same duration `XORRandom(3) + 3` but since `XORRandom(3)` is random each time, the flag animation speed will be slow or fast. My suggestion was that it should always be the same for every flag blob.

PR #1642, opened 2023-05-19
[fixed] Vehicle arms not facing correctly

Gingerbeard5773:
## Description

Fixes the bug where attached vehicles' arms would look incorrect when changing direction.
(mounted bow, catapult, ballista)

## Steps to Test or Reproduce

1) Attach a catapult to a warboat. Change the warboat's facing direction. Without the fixes, the catapult's arm will look incorrect.


mugg91:
I had made a fix here one year ago, [PR #1251](https://github.com/transhumandesign/kag-base/pull/1251) that makes it so Mounted Bow and vehicles don't look bugged when attached to another vehicle.

Gingerbeard5773:
conflict resolved

asumagic (approved):

PR #1677, opened 2023-09-16
Refactor map code and make it 3 options

asumagic:
## Description

Title, essentially. The code should be easy to modify should someone want to set fewer or more maps, or to add back the random button (although this code path is now untested).

Closes #993

## Steps to Test or Reproduce

Test localhost mode, test multiplayer and try as many edge cases as possible.


asumagic:
New appearance: https://media.discordapp.net/attachments/337266863841214465/1152682720896499752/image.png

PR #1678, opened 2023-09-16
Cinematic camera improvements

asumagic:
## Description

Goals:

- [x] Make the cinematic camera smoother and less crazy when a lot of stuff is moving around
- [x] Make it more obvious how to toggle the cinematic camera, but without nagging the user too much. This will be done by adding a small notice that temporarily reveals in certain circumstances (to determine)
- [x] Make it harder to accidentally toggle the camera off or on, e.g. ignore toggling within the first ~second of the user dying to avoid unintentional input
- [x] Add an option to permanently disable the cinematic camera if desired (will require engine PR)
- [x] Preserve disabling the cinematic camera across map changes (the temporary RMB one, not the new permanent setting)

## Steps to Test or Reproduce

Due to the addition of `v_camera_cinematic`, which is also added as a setting in the Misc. menu, this relies on a specific engine PR so this cannot be tested without a test build, which is not released yet.


asumagic:
When entering spectator mode, or when toggling the cinematic camera temporarily with RMB, a small pane shows up on the bottom left for one second. This is not revealed on death. This pane can also be revealed by bringing the cursor very close to it.

demo of example pane

When panning the camera around while the cinematic camera is enabled, a small yellow line shows how long it will take until the cinematic camera takes over again. This is more intended as a visual aid to realize why the camera goes back to normal, and that the cinematic camera is involved. When this happens, the pane is always revealed.  
Since the pane is bright, you don't notice the yellow bar until you actually stare at it.

demo of yellow line

`v_camera_cinematic` was added and provided as a setting in options. Disabling it means the cinematic camera is fully disabled and no pane will ever show up.

setting screenshot

asumagic:
Will merge this in while not very sure of the heuristics to reduce the motion of the camera. It is difficult for me to reproduce a real world CTF scenario, so ideally we would test it with enough players and figure out if it's good.

I tried to make it meaningful for camera tracking in TDM too: Usually, the camera will lock to center which reduces unnecessary motion significantly.

Issue #1545, opened 2022-12-27
Getting inside a Crate while moving is too hard


mugg91:
## Current in-game behaviour.

Stand anywhere, hold a Crate, have the mouse cursor pointed in the far sky. 
Then show buttons and let go in quick succession (= "Tap").
Notice that you will get inside the Crate. 👍

Hold a Crate and climb your team's fortress and bounce yourself off with a trampoline.
Furiously spam "show button" but you will never get inside the Crate until the moment you landed.

## Proposed Change

Make Tapping work even when you are moving through the air at high speed.

mugg91:
I have had no problems getting into crates while moving at high speed. My cursor can be anywhere - as long as I'm tapping, I will get into the crate I'm holding.

mehwaffle10:
I have a rework for my balance mod that uses attachments instead of inventories for players. They're also reusable, auto pickup when getting out, and I've swapped getting in/out to space (which utilizes the activation commands) and emptying to e. Fixes issues with controls and warping, may be worth considering parts of if not all for vanilla.

PR #1690, opened 2023-09-24
[fixed?] Crate button being hard to press at high speeds

MrPotatobird:
## Status

## Description

Tapping E should select your held item's button even when you're going fast.
Addresses #1545 

I don't fully get why this works, but it seems like it does. Haven't tested on a laggy server. Also it might make it slightly harder to tap E to open a shop if you're holding a log or something and going really fast?

## Steps to Test or Reproduce

Get a crate, or a scroll of doom or something
Jump off a cliff
Tap E

### DIFF
--- a/Entities/Common/Controls/StandardControls.as
+++ b/Entities/Common/Controls/StandardControls.as
@@ -174,7 +174,9 @@ void onTick(CBlob@ this)
 	{
 		if (this.get_bool("release click"))
 		{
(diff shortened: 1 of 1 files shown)


asumagic:
> Haven't tested on a laggy server.

I just have. With 100ms, it works, whereas it doesn't without.

> I don't fully get why this works, but it seems like it does.

I think it has to do with the radius checks and attachment position lagging behind player position on servers.

asumagic (approved):
This behavior looks sort of more correct for vanilla anyway...

PR #1694, opened 2023-09-26
Set bool "facing" of sieges false on init, and catapult init arm fix

longwoodbow:
## Status

## Description

In the case of ballista, it does force arm updating while this.getTickSinceCreated() < 30.
but other case, set bool "facing" true and check this.get_bool("facing") != this.isFacingLeft().
mostly force arm updating is needed when facing left and unpack siege, but because "facing" is true, it does force arm updating when facing right.
so i guess we should set it false and do force arm updating when unpack when facing left, not right.

## Steps to Test or Reproduce

Do something in this list
- Unpack a crate with catapult, ballista or mounted bow
- collide with bomber from left/right to siege that facing right/left
And check their arms


longwoodbow:
I changed catapult's script for fixing initial arm angle.
... or should I make a method that set arm and do it in onInit hook?

PR #1732, opened 2023-10-26
Fix archer with no arrows at TDM issue

longwoodbow:
[Fixed] Archer can't get supply arrows when changing class in TDM

## Status

- **READY**

## Description

Because of issue in arrow cleaning system, archer can't get arrows when item arrows are in the radius of spawn.
But then I noticed that we can dupe arrows easily, so I'll try to make resupply timer for prevent duping.

## Steps to Test or Reproduce

1. Go to TDM
2. Get mat_arrows and drop near spawn
3. Become archer at the spawn
4. Currently. all arrows are disappeared and archer can't get arrows


longwoodbow:
Sorry for late, it's ready

PR #1765, opened 2023-12-21
Cmd sanitization archerlogic

bunniewormy:
## Status

## Description

ArcherLogic.as && ArcherCommon.as:
	"shoot arrow" - was client->server, removed, is serverside code now. added "play fire sound" command that just plays a sound
	"grapple sync" - Instead of the client telling the server what they're grappling, the code for grappling is run server-side and client-side at once now. The server then syncs the server-side grapple to the client. I think.
	"pickup arrow" - was client->server->client, is now client->server. Added client-side "pickup arrow client", which just plays the sound.


mugg91:
If you shoot and throw away your arrows in quick succession, you will shoot an arrow without depleting the arrows you threw away.

asumagic (approved):
Will need more testing, seems fine in local testing, not ideal still

PR #1790, opened 2023-12-21
Cmd sanitization shops

bunniewormy:
## Status

## Description

Shop.as
	"shop buy" - is client->server, sanitized a bit more.
	"shop made item" - was server->client and server->server. server->server now uses callbacks. is now server->client "shop made item client".

### DIFF
--- a/Entities/Common/Factory/ShopCommon.as
+++ b/Entities/Common/Factory/ShopCommon.as
@@ -1,3 +1,5 @@
+funcdef void ShopMadeItem(CBitStream@);
+
 shared class ShopItem
 {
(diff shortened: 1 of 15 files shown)


Entities/Common/Factory/ShopCommon.as
```diff
@@ -1,3 +1,5 @@
+funcdef void ShopMadeItem(CBitStream@);
```
asumagic, line 1: Could've just taken the params directly but w/e, can "just be refactored later" (lol)

asumagic (dismissed):

PR #1799, opened 2023-12-21
Cmd sanitization minimapsync

bunniewormy:
## Status

## Description

MinimapSync.as
	The script overall feels unnecessary. The client requests sync on 1st tick after joining and after restart. The server then syncs 3 bools to all clients. There is no reason why the bools would not be synced automatically by the server. Commented the script out; if its removal causes any issues in the future, it should be added back.


asumagic (approved):
yolo, surely can't have negative consequences

Issue #1818, opened 2024-01-04
Make presents immune to fall damage


mugg91:
## Current in-game behaviour.

If a present spawns on a very high ledge or single tile, the parachute will deactivate and the present falls down and despawns due to falling too far.

## Suggestions

Don't despawn presents from falling too high. Make them immune to fall damage.

mugg91:
Alternatively we could make the contents jump out on falling from high.

PR #1826, opened 2024-01-09
[changed] Presents won't take fall damage + won't get sawed

mugg91:
## Status

## Description

[changed] Present will not take fall damage
[changed] Present will not take get sawed in the first 30 ticks
[fixed] that you couldn't damage presents with the same color as your team (via jabbing)

----

`this.Tag("ignore fall");` was added to `Present.as` which makes presents immune to fall damage. 
It looked odd that presents could fall down and just disappear.
Fixes https://github.com/transhumandesign/kag-base/issues/1818

----

`this.Tag("ignore_saw");` was added to `Present.as` which means it will not bounce off of or be standing on a Saw.
It was added just in case and actually doesn't change anything since `Present.as` has:
```
{return blob.getShape().isStatic() && blob.isCollidable();}
```

----

`Christmas.as` actually assigned team numbers to the presents so Blue team couldn't jab-damage blue presents and Red team couldn't jab-damage red presents. Therefore `spawnPresent()` was changed to make presents always team 255. `Present.as` applies 
```
void onInit(CSprite@ this)
{this.ReloadSprites(XORRandom(8), 0);}
```
to make just the sprite team-colored.

----

`Saw.as` was changed to make presents not get sawed if they existed less than 30.
This fixes https://github.com/transhumandesign/kag-base/issues/1540
I tried a general fix in https://github.com/transhumandesign/kag-base/pull/1348 by using tag "can be sawed immediately" but it wasn't implemented so I'm going with this simpler yet hackier fix this time.

----

Tested and it works.

## Steps to Test or Reproduce

Go to CTF.
Spawn Saw.
Spawn present.
Notice the present will not get sawed right away.

Spawn a present and change into the same team color.
Try to jab at the present.
Notice it will take damage now.

Go to top of the map.
Spawn present and have it fall down all the way.
Notice it will not get destroyed.


mugg91:
Fixed file conflict, removed an empty space.
Maybe instead of checking for "present" specifically, we could introduce a tag "dont saw immediately" and give that to blobs.

mugg91:
This had a file conflict or two, that github didn't catch (?)
I opened a new PR to improve upon this one: https://github.com/transhumandesign/kag-base/pull/2249

Issue #1812, opened 2023-12-30
Bombs should be allowed to fly above the ceiling


mugg91:
## Current in-game behaviour

Bombs should be allowed to fly above the ceiling freely because
1) Arrows are allowed to fly above the ceiling, too
2) Bombs are commonly used as projectiles, being thrown to the enemy side. So as projectiles they should have the same privilege like arrows.
3) It looks odd when bombs are thrown to the enemy side and get pushed below the ceiling in the last moment. As a result it would make targeting easier.
4) Bombs explode and despawn anyway, so it's not like they will fly above the ceiling indefinitely.

PR #1827, opened 2024-01-09
[change] Bombs are not limited by the top of the map

mugg91:
## Status

## Description

[changed] Bombs can fly over the top limit of the map

Fixes https://github.com/transhumandesign/kag-base/issues/1812

Bombs should be allowed to fly above the ceiling freely because
1) Arrows are allowed to fly above the ceiling, too
2) Bombs are commonly used as projectiles, being thrown to the enemy side. So as projectiles they should have the same privilege like arrows.
3) It looks odd when bombs are thrown to the enemy side and get pushed below the ceiling in the last moment. As a result it would make targeting easier.
4) Bombs explode and despawn anyway, so it's not like they will fly above the ceiling indefinitely.

Video demonstrating this: https://www.youtube.com/watch?v=NvXtCCtcELU

### DIFF
--- a/Entities/Items/Explosives/Bomb.as
+++ b/Entities/Items/Explosives/Bomb.as
@@ -11,8 +11,12 @@ void onInit(CBlob@ this)
 	this.set_u16("explosive_parent", 0);
 	this.getShape().getConsts().net_threshold_multiplier = 2.0f;
 	SetupBomb(this, bomb_fuse, 48.0f, 3.0f, 24.0f, 0.4f, true);
-	//
-	this.Tag("activated"); // make it lit already and throwable
+
+	//dont collide with top of the map
+	this.SetMapEdgeFlags(CBlob::map_collide_left | CBlob::map_collide_right);
+	
+	// make it lit already and throwable
+	this.Tag("activated");
 }
 
 //start ugly bomb logic :)


mugg91:
I like that you can throw things above the boundary. In my opinion it should apply to **all** blobs, except players.
We just need to make sure blobs can't somehow fly up infinitely or linger around above the boundary (for example, trampolines).

Entities/Items/Explosives/Bomb.as
```diff
@@ -11,8 +11,12 @@ void onInit(CBlob@ this)
 	this.set_u16("explosive_parent", 0);
 	this.getShape().getConsts().net_threshold_multiplier = 2.0f;
 	SetupBomb(this, bomb_fuse, 48.0f, 3.0f, 24.0f, 0.4f, true);
-	//
-	this.Tag("activated"); // make it lit already and throwable
+
+	//dont collide with top of the map
+	this.SetMapEdgeFlags(CBlob::map_collide_left | CBlob::map_collide_right);
```
Vam-Jam, line 16: Use `this.SetMapEdgeFlags(CBlob::map_collide_sides);` instead

Vam-Jam (changes requested):

asumagic:
> I like that you can throw things above the boundary. In my opinion it should apply to **all** blobs, except players. We just need to make sure blobs can't somehow fly up infinitely or linger around above the boundary (for example, trampolines).

That's reasonable, the only concern is maps that have a solid partial roof (esp in TDM), but that's fairly rare and probably a non-issue.

mehwaffle10:
I would go further with this. In my mod we have it so that *everything* can go above max build height cause it's just that: max build height. On map start it scans the top to see if the map is a cave map and dynamically will make it so blobs can go above the map. This is hugely beneficial for getting over towers which can already be very difficult

Issue #1751, opened 2023-12-04
TDM minor issues


mugg91:
1) `MonsterTDM1` has grass tiles above non-dirt tiles.

2) `Biurza_Clogged_Pipes`: The trampolines get destroyed by water, although they have the "invincible" tag. Not sure if this is intentional.

mugg91:
I will move the "bad grass tiles" part to a different issue.

PR #1829, opened 2024-01-09
[fixed] removed bad grass in MonsteriTDM1 + Trampolines in Biurza_Clogged_Pipes won't decay in water anymore

mugg91:
## Status

## Description

[fixed] removed wrong grass file in `MonsteriTDM1`
[changed] decaying will not happening when the blob has tag "invincible"

Fixes https://github.com/transhumandesign/kag-base/issues/1751

----

Removed a grass file above empty space in TDM map `MonsteriTDM1`.
See https://i.imgur.com/99r25D8.png
As a side effect, the new png file takes 671 bytes less.

----

In `Biurza_Clogged_Pipes` the trampoline would get destroyed by the water which is odd because trampoline placed by the map loader has the "invincible" tag.
I edited `DecayCommon.as` to make it so blobs don't decay if they have the "invincible" tag.
Invincible things should not get destroyed.

This also means blobs with `Tag("invincible")` won't decay from being spammed, being left alone, being flipped upside down, or being on land now. But there is no gameplay situation were this could happen since those blobs (upside down catapult, on land warboat, buckets getting spammed etc.) aren't given the "invincible" tag so everything should be normal.
I don't see a possibility for regressions right now. 

## Steps to Test or Reproduce

Load map `MonsteriTDM1`.
Notice the grass looks better.

Load map `Biurza_Clogged_Pipes`.
Wait until the water reaches the trampolines.
Notice they will not decay in water anymore.


mugg91:
I will work on two separate PRs, one fixing the grass in 3 maps and one fixing the decaying.

PR #1836, opened 2024-01-10
[fixed] Greg not grabbing you if you previously got in a crate while grabbed

### LINKED ISSUE #1736: Greg will not grab you after entering crate while grabbed
## Description

1) Hold a crate
2) Get grabbed by a greg
3) get inside the crate while grabbed, then get out
4) Gregs will still try to grab you but will just collide with you without grabbing you.

Tested in offline and online.

mugg91:
## Status

- Needs testing

## Description

```
[fixed] Greg not grabbing you after you got in a crate while grabbed before
[changed] You will not see crate buttons when you are attached
```

Fixes https://github.com/transhumandesign/kag-base/issues/1736

The bug with Greg not grabbing you happened because tag `"picked"` on the targetblob  wasn't untagged after the game detached you from Greg as a result of getting in a crate.

Previously, tag `"picked"` was untagged after Greg.as detaches you.
This PR accounts for the fact that you could get detached outside of Greg.as, by adding `onDetach()` to Greg which untags `"picked"` and sets bool `"no target"`. As far as I tested, Greg now works correctly.

This PR also adds a second fix to make it so that you cannot see crate buttons or get in a crate in the first place, if you are attached to something. As far as I tested, this also works fine but it needs more testing to see if there are other gameplay scenarios that might be affected by this. You could get in a crate while sitting on a catapult's driver seat which wouldn't be possible after this PR goes live (this one is not a big deal).

## Reproduction Steps

Go in Sandbox.
Spawn Greg and crate.
Pick up a crate and wait for Greg to grab you.
While grabbed, get in the crate.
Wait for Greg to grab you again.
Notice he will try to grab you but you will not be grabbed.
**After this PR, you can't get in the crate, but even if you could Greg would properly grab you.**

### DIFF
--- a/Entities/Items/Crate/Crate.as
+++ b/Entities/Items/Crate/Crate.as
@@ -249,7 +249,7 @@ bool isInventoryAccessible(CBlob@ this, CBlob@ forBlob)
 
 void GetButtonsFor(CBlob@ this, CBlob@ caller)
 {
-	if (!canSeeButtons(this, caller)) return;
+	if (!canSeeButtons(this, caller) || caller.isAttached()) return;
 
 	Vec2f buttonpos(0, 0);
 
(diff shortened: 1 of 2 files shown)


mugg91:
The issue still exists and this PR still works. 
It happens in both offline and online, and the fix works the same in both offline and online.

## Before
(Please don't mind the wrong file name)

https://github.com/user-attachments/assets/afd9578d-ee0d-4013-9318-5bf7bbe1b75b

## After

https://github.com/user-attachments/assets/0ad794f3-4372-4e7d-8b28-6f7358b02961

PR #1838, opened 2024-01-12
[added] Mounted Bow to vehicle shop + Mounted Bow changes

### LINKED ISSUE #1741: Add mountedbow to siege workshop
Add mountedbow to siege workshop. It will add "new content" practical without any effort. It seems to me that the optimal price would be: 200 wood and 50 or 60 coins

### LINKED ISSUE #2203: Mounted Bow will fall through gaps
## Description

https://github.com/user-attachments/assets/2d7b8a4f-caad-461f-90fd-ab7be4144f28

Originally mentioned and fixed in https://github.com/transhumandesign/kag-base/pull/1251

### LINKED ISSUE #2204: Mounted Bow will fall through doors and platforms
## Description

https://github.com/user-attachments/assets/c9ff9540-ebb3-4f1c-bc54-92ddde3ac2bf

Originally mentioned and fixed in https://github.com/transhumandesign/kag-base/pull/1251

### LINKED ISSUE #2205: Mounted Bow should change team to player who picks it up
## Description

https://github.com/user-attachments/assets/7c4c2847-3284-4d7e-9367-07fa86e52681

Note that Mounted Bow which is loaded with the map should not be pickable and remain as neutral team when attaching to it in the gunner position.

Originally addressed in https://github.com/transhumandesign/kag-base/pull/1251

mugg91:
## Status

## Description

[added] Mounted Bow to vehicle shop, can be bought for 80 coins
[changed] Mounted Bow can be picked up unless it was loaded with the map
[changed] Mounted Bow health from 2.0 to 2.5
[changed] Mounted Bow properties, making it heavier
[changed] Mounted Bow width is just wide enough to not fit in 1 tile gaps
[changed] Mounted Bow inventory space from 3 slots to 2 slots
[changed] Mounted Bow arrows won't despawn early
[changed] Mounted Bow shoot delay from 25 ticks to 35 ticks
[changed] Mounted Bow will not fall through platforms and doors

Fixes https://github.com/transhumandesign/kag-base/issues/1741
Fixes https://github.com/transhumandesign/kag-base/issues/2205
Fixes https://github.com/transhumandesign/kag-base/issues/2203
Fixes https://github.com/transhumandesign/kag-base/issues/2204

I added Mounted Bow to vehicle shop so it can be used in CTF which was sometimes requested by players, to add a bit more variety to the game.

I made some changes to Mounted Bow:
* It was a featherweight item and could be thrown far across the map. I changed that to make it heavier.
* I gave it more health so it isn't destroyed and 80 coins were wasted right away
* I made it not fit in 1 tile gaps because players could attach themselves to Mounted Bow and fall down through gaps.
* I also changed the shape vertically so it looks like it actually touches the ground (was floating 1 px high in the air previously)
* I gave it less inventory space in an effort to make it less of a camping tool
* I made it so arrows don't despawn as I like it better that way. I don't think making it a short distance tool in CTF is a good idea.

----

~My only concern now is maybe we should make it shoot at a lesser frequency.
It fires slightly too often imo.
What do you think?~

## Steps to Test or Reproduce

Go to Sandbox or CTF.
Spawn vehicle shop, 
purchase a Mounted Bow and play around with it.


mugg91:
Should Mounted Bow be able to convert to the other team? By pickup or by waiting at it?

mugg91:
I made a commit that does the following changes:

I changed how Mounted Bow behaves as a medium weight item.
* Tag "medium weight" was added to Mounted Bow which actually does what I had intended, reducing throw distance
* mass reduced from 150 to 100
* friction changed from 0.5 to 0.6
* elasticity changed from 0.1 to 0.3
* buoyancy changed from 0.1 to 0.2
* drag changed from 1.0 to 0.5

I addressed the issue with how it shoots too frequently; it could be too spammy in CTF
* shoot delay changed from 25 ticks to 35 ticks

I made mounted bow possible to steal from enemy team by pick-up
* Added `SetTeamToCarrier.as` as blob script to `MountedBow.cfg`

GELGELAT:
its really cool. You have to look at it relative to regular archers. I dont think hes supposed to shoot faster than them. Imagine if 3 people shoot it like a turret. maybe make it a team item?

mugg91:
What do you mean by making it a team item?
Right now, it can only be mounted by team players but can be stolen by the other team via pickup.
Should I add a check like in Saw, that it can only be picked from below or when overlapping?

GELGELAT:
> What do you mean by making it a team item? Right now, it can only be mounted by team players but can be stolen by the other team via pickup. Should I add a check like in Saw, that it can only be picked from below or when overlapping?

let's go like a saw, either way its worth a lot of coin to just steal.

mugg91:
I think we should add a wood cost, too. 100 or 150 wood. Because there is usually plenty wood anyway and buying mounted bow shouldn't be something someone can just spam.

> let's go like a saw, either way its worth a lot of coin to just steal.

Of course, bringing it to the battle field is risky.
I've though about it more like fortifying your castle and placing it behind platforms so it can't be picked  (assuming this will be in the next build, too: https://github.com/transhumandesign/kag-base/pull/1864 ). 

I think it's not too far off compared to if you mess up your Keg or 2-3 bombs. Or when you think about bringing Saw to the battle field, that's also a high risk for wasting 100 stone, so Saw is usually used only for industry.

GELGELAT:
> Should we add a wood cost? Like 100 or 150 wood. Because there is usually plenty wood anyway and buying mounted bow shouldn't be something someone can just spam.
> 
> > let's go like a saw, either way its worth a lot of coin to just steal.
> 
> Of course, bringing it to the battle field is risky. I was thinking about it more like fortifying your castle and placing it behind platforms so it can't be picked (assuming this will be in the next build, too: #1864 ).
> 
> I think it's not too far off compared to if you mess up your Keg or 2-3 bombs. Or when you think about bringing Saw to the battle field, that's also a high risk for wasting 100 stone, so Saw is usually used only for industry.


The cost in the form of wood is not so necessary this resource can be obtained everywhere, I think it will only be a distraction. Hmmm.

GELGELAT:
I have an idea. What if you let the bow be picked up when theres already a person sitting in it? What do you think?

GELGELAT:
you could do that with 2 or more units. Sounds creepy, but like for fun

mugg91:
I committed changes:
* Somehow `VehicleShop.as` had file conflicts that are now resolved - Buying from vehicle shop won't cause errors in the console now.
* Mounted Bow will not fall through doors and platforms.
* Mounted Bow can be picked by allies from afar, and from enemies only when overlapping (same as crates).

PR #1841, opened 2024-01-12
[added] HUD background is transparent when mouse hovering

mugg91:
## Status

## Description

`[added] Background of HUD is transparent when mouse-hovering over it`

Screenshots
https://i.imgur.com/8ovhCOW.png
https://i.imgur.com/fb3fWte.png

Similar to how the chat background turns invisible on mouse hover, this PR makes it so the HUD background turns invisible on mouse hover as well.

### DIFF
--- a/Entities/Common/GUI/DefaultActorHUD.as
+++ b/Entities/Common/GUI/DefaultActorHUD.as
@@ -3,6 +3,9 @@
 
 #include "ActorHUDStartPos.as";
 
+bool hide_hud_background = false;
+const int HP_segment_width = 32;
+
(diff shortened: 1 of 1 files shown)


mugg91:
Waiting for review...

Vam-Jam (changes requested):
You should check the mouse position on the start of onRender and pass it to the functions that need it, instead of declaring it as a global and checking at the end of a frame. 

Also not sure if this is a feature people would find useful? I feel like it would be better to have it fade in and out based on how far the mouse is, similar to health bars instead of suddenly disappearing

PR #1840, opened 2024-01-12
[fixed] Archer fake charging bug

mugg91:
## Status

- **READY**: this PR is (to the best of your knowledge) ready to be incorporated into the game.

## Description

`[fixed] Archer fake charging bug`

Fixes https://github.com/transhumandesign/kag-base/issues/1807

Video https://www.youtube.com/watch?v=HTrvW8kj3z4

When you charge up arrow and get jabbed by enemy Knight, you will be looking like you are charging arrow but you actually aren't; letting go of `action1` will do nothing.

I added an `else` code block so archer will charge up a new arrow after getting jabbed and when still holding action1, just like you would charge up a new arrow after getting knocked back and the knockback ends.

```
if ((just_action1 || this.wasKeyPressed(key_action2) || this.hasTag("redo charge") && !pressed_action2) &&
 (charge_state == ArcherParams::not_aiming || charge_state == ArcherParams::fired || charge_state == ArcherParams::stabbing))
{	// ...
}
else if (charge_state == ArcherParams::readying)
{	// ...
}
else if (charge_state == ArcherParams::charging)
{	// ... 
}
else if (charge_state == ArcherParams::no_arrows)
{	// ... 
}
else
{
	charge_state = ArcherParams::not_aiming;
	charge_time = 0;
	this.Tag("redo charge");
}
```

Tested in offline and online, no problems encountered.

## Steps to Test or Reproduce

Go to TDM and use `AddBot("");` in console.
(You might want to edit `KnightBrain.as` so the bot jabs more often.)
Be Archer and charge arrow. While charging, get jabbed by the enemy knight bot.
Notice you will still have your bow drawn as if charging an arrow but nothing will happen.
**After this PR, you will charge up a new arrow after getting jabbed.**

PR #1843, opened 2024-01-13
[Fixed] Archer sometimes holds bow and doesn't pull

longwoodbow:
## Status

## Description

https://github.com/transhumandesign/kag-base/assets/68646058/7e662b77-7f80-41a3-bf8e-71c9f0022878
(For testing, archer gets knocked 2 ticks when press space)

Sometimes archer can't charge, it is fixed when re-click, then I think it is because of short knocking, such as knight's jab.
Probably non my player character get solved by this too, I don't know that the archers holds bow horizonal and doesn't pull but shoots arrows...

## Steps to Test or Reproduce

-Pull bow
-Take jab from knight
-Check that recharge starts


mugg91:
This will not work because holding action2 (grapple) and then pressing action1 (arrow) will cause archer to not charge arrow.

However, I found a solution. This will work, making archer re-charge arrow after stun/jabs:
```

		if ( (!pressed_action2 || (just_action1 && this.wasKeyPressed(key_action2)) ) &&
		        (charge_state == ArcherParams::not_aiming || charge_state == ArcherParams::fired || charge_state == ArcherParams::stabbing))
		{
```

mugg91:
Nevermind, the issue of not charging arrow after getting jabbed still exists when holding action2.
This issue doesn't exist in https://github.com/transhumandesign/kag-base/pull/1840

PR #1847, opened 2024-01-15
[added] Ballista Bolt can drop down from stuck position

### LINKED ISSUE #1750: Ballista bolts should drop after tile it's stuck to collapses
## Description

Not so long ago, arrow was changed to despawn if the tile it hit is removed/collapses or if the door or platform it hit opens.

But ballista bolts could still float in midair after the tile it hit collapses.

I suggest, instead of despawning like arrow, bolt should fall down and have a chance at hitting something once more.

mugg91:
## Status

## Description

```
[fixed] Regular Ballista Bolt will not remain in the air after the thing it stuck to collapses or gets destroyed
[added] Regular Ballista Bolt will drop if the tile/blob it stuck to is destroyed or opened
[changed] Regular Ballista Bolt will hit blobs only at velocity 1.5f or higher
[changed] Can only be climbed like a ladder when static
```

Fixes https://github.com/transhumandesign/kag-base/issues/1750

Previously, Ballista Bolt would remain in the stuck position (possibly in midair) even when the tile or blob it hit is gone.
This PR implements the same change that was previously added to `Arrow.as`.
This makes sense as Ballista bolts are very similar to arrows.
However, instead of simply despawning, Ballista Bolt will drop down and have another chance at hitting something.

That means that if Ballista Bolt hits 
* a door/platform blob and it opens
* a tile and the tile collapses or is destroyed
then Ballista Bolt will drop down, possibly hitting more things.

Since I didn't want it to be too op (by having it insta-kill enemies at the time it drops) I added a check for velocity so it only hits enemies when it gained enough speed (on the 5th tick or so).

In the newly added `SetNonStatic()`, I couldn't make Ballista Bolt forget `"tileWorldPoint"` and `"hitBlob"` and these approaches didn't work:
* `this.clear("tileWorldPoint")`/`this.clear("hitBlob")`
* `this.set_Vec2f("tileWorldPoint", null)` (...)
* `Vec2f nullvec; this.set_Vec2f("tileWorldPoint", nullvec)` (...)
So I had to add booleans for whether to check "hitBlob" or "tileWorldPoint".

I addressed an issue with being able to climb Ballista Bolt in midair ( https://www.youtube.com/watch?v=fH33Qdr_gQo ).
Now you can only climb it when it is stuck.

Some empty lines removed while I was going through the file.

Tested in offline and online, it works as envisioned.

## Steps to Test or Reproduce

Spawn ballista.
Shoot at a stone tile.
Destroy or collapse the stone tile.
Notice the ballista bolt will still stick in the air.
**After this PR, it will drop down, possibly hitting more targets.**

Shoot at a door/platform.
Make the door/platform open.
Notice the ballista bolt will still stick in the air.
**After this PR, it will drop down (...)**


mugg91:
Filed file conflict, works now.

PR #1849, opened 2024-01-15
[changed] Plants don't make cut sfx when burning

mugg91:
## Status

## Description

`[changed] Plants (Bush, flowers, grain...) don't make a cut sfx when taking burn or fire damage`

It looked odd to me that plants would make a cut sound when taking burn damage, so this PR makes it so a cut sound is played but only if the damage source isn't "fire" or "burn". Otherwise the plant will take damage without a sound.

### DIFF
--- a/Entities/Common/Attacks/PlantHitEffects.as
+++ b/Entities/Common/Attacks/PlantHitEffects.as
@@ -5,7 +5,7 @@ f32 onHit(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@ hitt
 {
 	if (damage > 0.05f) //sound for all damage
 	{
-		if (customData != Hitters::fire && customData != Hitters::burn) // not fire or burn
+		if (customData != Hitters::fire && customData != Hitters::burn)
 		{
 			Sound::Play("/cut_grass", this.getPosition());
 		}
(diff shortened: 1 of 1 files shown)


Entities/Common/Attacks/PlantHitEffects.as
```diff
@@ -5,25 +5,16 @@ f32 onHit(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@ hitt
 {
 	if (damage > 0.05f) //sound for all damage
 	{
-		//read customdata for hitter
-		switch (customData)
+		if (customData != 7 && customData != 8) // not fire or burn
```
Gingerbeard5773, line 8: not a big deal but since this is base kag it is probably better to use the actual enums from hitters.as just for reading purposes. you can disregard this comment though if wanted

Gingerbeard5773 (approved):

Entities/Common/Attacks/PlantHitEffects.as
mugg91: I changed it.

Gingerbeard5773 (changes requested):
this will error, you need to add in ``#include "Hitters.as"``

mugg91:
> this will error, you need to add in `#include "Hitters.as"`

It is already included.

Issue #1811, opened 2023-12-25
Make water clean bloody things


mugg91:
## Describe the feature

Water splash from bucket or being underwater should clean Saw, spiker and spikes from being bloody.

mugg91:
If this were to be implemented, it should probably be done in PR https://github.com/transhumandesign/kag-base/pull/1507

Issue #1753, opened 2023-12-09
kidssafe can't deactive small flesh gibs


longwoodbow:
**Description**

small flesh gibs appear on destroying corpses, even kidssafe is activated.

**Reproduction**

1. activate kidssafe
2. destroy the corpse of player

**Additional context**

by the way, kidssafe disallows pain sound too, i think shouldn't disallow it.

mugg91:
On a related note, I had made a - as of yet unimplemented - PR that made Saw, Spiker and Spikes properly remember and toggle "bloody" status. 
https://github.com/transhumandesign/kag-base/pull/1507

PR #1854, opened 2024-01-16
[fixed] Blood on Saw/Spikes/Spiker will be toggled properly on changing g_kidssafe

mugg91:
## Status

## Description

```
[fixed] Spikes, Spiker and Saw will properly remember if they are bloody or not and will instantly update their spritelayer when changing g_kidssafe from the menu
[changed] Spikes don't continuously update their animation frame in onTick() via calling onHealthChange()
[changed] Spiker will change its animation frame rather than adding a spritelayer
[changed] There will be no flesh gibs when g_kidssafe is on
[added] Water splash from bucket, water bomb or water arrow will remove the blood 
[added] Spiker will show blood on its tip even in closed state
```

Fixes https://github.com/transhumandesign/kag-base/issues/1753
Fixes https://github.com/transhumandesign/kag-base/issues/1811

This PR implements the same changes as [PR #1507](https://github.com/transhumandesign/kag-base/pull/1507) and does the following additional changes:

* The conflict with `Saw.as` has been resolved.
* There will be no flesh gibs from player corpses or animals when `g_kidssafe` is on.

Other than that please refer to the write-up of the above linked PR.


mugg91:
I made a change so that "blood" tag is removed when using a water splash from bucket, water bomb or water arrow.

Only watersplash will clear blood, but going underwater won't because it looked odd in testing.

mugg91:
I added a commit that does the following changes:
* Merged `Spike.as` (which was a sprite script) and `GenericSpike.as` (which was a blob script) to `Spike.as` (sprite + blob script).
* Spiker's tip in closed state will now also show blood: When spiker is placed, it saves spiker's network ID to spike. When spike gets bloody or its blood status is toggled, it will now also fetch spiker blob and update its background spritelayer.
* Changed `Spiker.png` to add the bloody spiker frame and to use less space.
Tested offline and online and it works.

mugg91:
There is a problem that the Spike sprite will be behind the Spiker's main frame after rejoining server.
Will work on a fix.

mugg91:
Added a commit that makes the following changes:
* When cleaning bloody blob by water splash, script `CleanSparkles.as` is applied to the blob which will produce sparkle particles and a sparkle sound effect will play. This happens only if "faster graphics" is turned off.
* Sparkle sound effect taken from here: https://pixabay.com/sound-effects/magic-sparkle-190030/

Aforementioned issue with spiker still unfixed as of this writing.

mugg91:
Added a commit that makes it so the toggling doesn't take place when the menu is closed but when the `g_kidssafe` boolean changes.

mugg91:
I will fix the merge issue with this PR sooner or later.

mugg91:
I'm replacing this PR with this new one https://github.com/transhumandesign/kag-base/pull/2271

PR #1861, opened 2024-01-21
[added] Bison and Shark icons to kill feed

mugg91:
## Status

## Description

When you get killed by a bison or shark, the "fall death" icon would be shown in the kill feed.

This PR changes that so the bison or shark would be shown as the death reason in the kill feed.

Ingame look: https://i.imgur.com/6GMMoEk.png

Since `Hitters::flying` is used by things other than Bison, I changed it in `Bison.as` to be `Hitters::bison`. I have not seen problems following this change. I added it to the end of the list in `Hitters.as` since I couldn't be bothered changing all the files.

Since Shark is the only blob right now that uses `EatOthers.as` - which uses `Hitters::bite` - I applied the shark icon directly for it.

As a side effect, KillFeedIcons.png is now 124 bytes smaller (-93%).

## Steps to Test or Reproduce

Get killed by a shark or bison.
Notice the death icon would actually be a shark or bison instead of the fall death icon.


Gingerbeard5773 (approved):

mugg91:
I made a separate PR that adds bison + shark but also trampoline.
https://github.com/transhumandesign/kag-base/pull/2184

So people can decide if they want only biso and shark or all three.

Issue #1342, opened 2022-07-28
Items shouldn't be picked through platforms


mugg91:
## Current in-game behaviour.

When an item is hidden behind a platform facing against the picker, the picking up will still happen.

Screenshot - https://i.imgur.com/3cyHqtf.png

In the screenshot, you can pick the blobs behind platforms and bridge, but not behind the door. My suggestion really only applies for the platforms though, not the team bridges.

## Proposed Change

Make a blob to not be pickable if it is behind a platform that faces the picker.
If a player wants to pick a blob behind platforms, they must be destroyed first.

mugg91:
I couldn't find an easy way to implement this yet.

Assuming there is only a platform between `blob1` and `blob2`,
`getMap().rayCastSolid(blob1.getPosition(), blob2.getPosition(), end);` will collide with a platform both when it is facing towards `blob1` as well as when facing away from `blob1`.
Knowing this, I'm not sure yet, why Arrow - which also uses `rayCastSolid()` - doesn't collide with away-facing platform.

NoahTheLegend:
you can get direction of pickup (pointing from blob to each item in end array) and add some kind of 
`if ((deg < 45.0f || deg > 315.0f) && vely > 0.0f) //up		
		{
			return true;
		}`
		this will check the rotation between a 90 degree triangle rotated by 45 for preventing small digit deviation (e.g. rotated to right platform can be having angle of 90.00001 etc)

mugg91:
@NoahTheLegend 
If I understand your solution correctly, this would apply to the whole block.
[So Player B in this image wouldn't be able to pick the item.](https://i.imgur.com/ffTQIFe.png)
I'm interested in a solution where Player A can't pick the item, but Player B can.

Maybe, in addition to your solution, one could use `CShape@ --> ShapePlatformDirection@ getPlatformDirection(int index)` to see where exactly the platform is. 
You could assume two lines, one from the player to the picked item, the other from the one corner to the other corner of the platform block. Check if they cross.
I may look into it when I have more time.

NoahTheLegend:
adding a check if player is beyond the facing direction with x | y coordinates may help
`if (blobPos.x > itemPos.x && (angle_degree < 45.0f || angle_degree > 315.0f)) //up { return true; }`
pick direction passing tru platform which is facing to right, blob's pos is more to right from the item, so it will prevent picking up item in this way

PR #1864, opened 2024-01-23
[fixed] You can not pick item through platform that faces toward you

mugg91:
## Status

## Description

`[fixed] You can not pick items through platform that faces toward you`

Fixes https://github.com/transhumandesign/kag-base/issues/1342

This PR does some small optimizations to and adds code to `StandardPickup.as` to render items unpickable if we are trying to pick through the border of a platform.

Please consider [this image](https://i.imgur.com/Eq4dFLf.png).
Picker is trying to pick the "blob to pick" at different positions (picker 1, picker 2, picker 3) through a right-facing platform.

* Picker 1: The raycast is not set false because picker 1's x pos is less than the border's x pos.
* Picker 2: The raycast is not set false because `hi.hitpos` is less than the border's x pos.  
* Picker 3: The raycast is set false because `hi.hitpos` is equal to the border's x pos.

This same logic is applied for all four platform directions.

Thanks @NoahTheLegend for your suggestion in determining the platform's facing direction.
Noah said platforms could have an angle that is off by ~0.01 so checking for e.g. `90.0f` (angle degrees) may fail.
However, I'd like to point out that `RunnerMovement.as` does check platform's angle for a direct value (`angle == 90.0f`) at line 834. I guess it doesn't matter after all?

## Screenshots

[unpickable because we are facing through the right border of the platform](https://i.imgur.com/QMedYVV.png)
[pickable because we are facing through the platform but not through the right border](https://i.imgur.com/WPwxoWD.png)

## Steps to Test or Reproduce

Try to pick item through a platform. 
It will work if the platform is facing away from you but it will not work if the platform is facing toward you.

### DIFF
--- a/Entities/Common/Controls/StandardPickup.as
+++ b/Entities/Common/Controls/StandardPickup.as
@@ -615,29 +615,89 @@ bool canBlobBePickedUp(CBlob@ this, CBlob@ blob)
 
 	float maxDist = Maths::Max(this.getRadius() + blob.getRadius() + 20.0f, 36.0f);
 
-	Vec2f pos = this.getPosition() + Vec2f(0.0f, -this.getRadius() * 0.9f);
-	Vec2f pos2 = blob.getPosition();
-
-	Vec2f ray = pos2 - pos;
-	bool canRayCast = false;
-
-	CMap@ map = getMap();
+	Vec2f picker_pos 	= this.getPosition() + Vec2f(0.0f, -this.getRadius() * 0.9f);
+	Vec2f blob_pos 		= blob.getPosition();
+	Vec2f ray 			= blob_pos - picker_pos;
+	f32 ray_length		= ray.Length();
+	bool canRayCast 	= false;
+	CMap@ map 			= getMap();
 
 	HitInfo@[] hitInfos;
-	if(map.getHitInfosFromRay(pos, -ray.getAngle(), ray.Length(), this, hitInfos))
+	if (map.getHitInfosFromRay(picker_pos, -ray.getAngle(), ray_length, this, hitInfos))
 	{
 		for (int i = 0; i < hitInfos.length; i++)
 		{
 			HitInfo@ hi = hitInfos[i];
-			CBlob@ b = hi.blob;
+			CBlob@ b 	= hi.blob;
 
-			// collide with anything that isn't a platform
-			// could do proper platform direction check but probably not needed
-			if (b !is null && b !is this && b !is blob && b.isCollidable() && b.getShape().isStatic() && !b.isPlatform())
+			if (b is null || b is this)
 			{
-				canRayCast = false;
-				break;
+				continue;
+			}
 
+			if (b !is blob && b.isCollidable() && b.getShape().isStatic())
+			{
+				if (!b.isPlatform())
+				{
+					canRayCast = false;
+					break;
+				}
+				else
+				{
+					// don't allow pick through platforms
+					
+					f32 angle 	= b.getAngleDegrees();
+					
+					if (angle > 45.0f && angle < 135.0f		// right-facing platform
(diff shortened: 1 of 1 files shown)


Entities/Common/Controls/StandardPickup.as
```diff
@@ -615,29 +615,89 @@ bool canBlobBePickedUp(CBlob@ this, CBlob@ blob)
...
+						f32 border_y = b.getPosition().y - map.tilesize / 2;
+						
+						if (picker_pos.y < border_y
+							&& Maths::Round(hi.hitpos.y) == Maths::Round(border_y))
+						{
+								canRayCast = false;
+								break;
+						}
+					}
+				}
```
Gingerbeard5773, line 700: there is a bunch of strange indents/tabs which looks shoddy

also the platform code seems bloaty for what it does.
heres a bool that is mostly copied over from rock.as that is much less bloaty but does most of your platform stuff
```
bool CollidesWithPlatform(Vec2f ray, CBlob@ platform)
{
	const f32 platform_angle = platform.getAngleDegrees();	
	Vec2f direction(0.0f, -1.0f);
	direction.RotateBy(platform_angle);
	const f32 velocity_angle = direction.AngleWith(ray);

	return !(velocity_angle > -90.0f && velocity_angle < 90.0f);
}
```

Gingerbeard5773, line 700: heres code that functions the same but is less bloaty, using CollidesWithPlatform bool instead.
```suggestion
			if (b !is blob && b.isCollidable() && b.getShape().isStatic())
			{
				//bypass platform if it is facing away from us
				const bool insidePlatform = (b.getPosition() - picker_pos).Length() <= b.getRadius();
				if (b.isPlatform() && (!CollidesWithPlatform(ray, b) || insidePlatform))
				{
					canRayCast = true;
					break;
				}
				canRayCast = false;
				break;
			}
```

Gingerbeard5773 (changes requested):

Entities/Common/Controls/StandardPickup.as
mugg91: I tested your code, it will still pick through platform regardless what direction it's facing.

mugg91: I got both your solutions to work. But it will not work the same as I envisioned because it seems [your solution doesn't allow picker 1 and picker 2 to pick the item](https://imgur.com/Eq4dFLf) whereas my PR does. 

I will try to cut down the repetitive code though.

mugg91:
By modifying the `CollidesWithPlatform()` code that @Gingerbeard5773 posted, I worked out a solution that cuts down on the repetitive code.

The `CollidesWithPlatform()` code I applied will check if we are facing against the platform as well as if the hit position is overlapping with the border position. I used `(angle + 45.0f) % 180.0f) < 90.0f` to check if the x position or the y position should be checked. 

I used `border_offset.RotateBy(Maths::Round(angle));` so angle is rounded and not off by 0.001

Works as before in testing.

Entities/Common/Controls/StandardPickup.as
mugg91: About tabs or indents, idk because it looks good on my end in notepad++
About the code itself, see my latest post below.

mugg91: Although I like tabs, I edited the file on github website itself to removed them and it should be good now.

mugg91:
Added a commit that changes the code inside the hitinfo loop because the code 

```
		if(map.isTileSolid(hi.tile))
		{
			canRayCast = false;
			break;
		}
```
couldn't be run if the loop returns if `b is null`.
Not tested but most likely works as intended.

Vam-Jam:
Closing in favour of https://github.com/transhumandesign/kag-base/pull/2145

PR #1873, opened 2024-01-27
[fixed] Autopicking from someone's hand

### LINKED ISSUE #1866: Items can be picked from your hand by AutoPickUp
## Description

1. Pick up a `mat_arrows`
2. Another player is playing archer and happens to walk over the `mat_arrows` just after you picked it up
3. The `mat_arrows` are "auto picked" by the archer even though they were in your hands.

Considering that "picking item from another player's hand" is something we fixed in `StandardControls.as` ( https://github.com/transhumandesign/kag-base/issues/1323 )
we should adjust `ArcherAutoPickup.as`, `KnightAutoPickup.as` and `BuilderAutoPickup.as`, too, by adding the same return conditions. 
The function should be returned from if the item is in inventory, is attached or can "not be picked up".

----

The fix that was applied to `StandardControls.as`:
```
if (owner !is null 
		    && !owner.isInInventory()
		    && !owner.isAttached()
		    && pick !is null 
		    && !pick.isAttached()
		    && pick.canBePickedUp(owner))
```

mugg91:
## Status

## Description

`[fixed] Autopicking from someone's hand`

Fixes https://github.com/transhumandesign/kag-base/issues/1866

This PR adds return conditions to `ArcherAutoPickup.as`, `BuilderAutoPickup.as` and `KnightAutoPickup.as` to make it so an item will not be "auto-picked" if it has been picked up by someone else already. 

I added `blob.isAttached()` and `!blob.canBePickedUp(this)` as return conditions that also exist in `StandardControls.as` when regularly picking up an item.
However, I did not add `owner.isInInventory()` and `owner.isAttached()` therefore still allowing auto-picking items when e.g. driving over materials while in a catapult. Let me know if it's ok or not.

I couldn't test the issue itself but it is extremely likely it will work as intended. I tested for regressions and didn't see any. Crate auto picking still works.

## Steps to Test or Reproduce

Join a server with a laggy friend.
Have laggy friend be archer.
Spawn a mat_arrows.
Pick up the mat_arrows and have the laggy friend walk over it in quick succession.
Notice the mat_arrows will be picked up by you but then be grabbed from your hand.
**After this PR, it will stay in your hand**

### DIFF
--- a/Entities/Characters/Archer/ArcherAutoPickup.as
+++ b/Entities/Characters/Archer/ArcherAutoPickup.as
@@ -9,7 +9,10 @@ void onInit(CBlob@ this)
 
 void onCollision(CBlob@ this, CBlob@ blob, bool solid)
 {
-	if (blob is null || blob.getShape().vellen > 1.0f)
+	if (blob is null 
+		|| blob.isAttached()
+		|| !blob.canBePickedUp(this)
+		|| blob.getShape().vellen > 1.0f)
 	{
 		return;
 	}
--- a/Entities/Characters/Builder/BuilderAutoPickup.as
+++ b/Entities/Characters/Builder/BuilderAutoPickup.as
@@ -43,7 +43,10 @@ bool pickupCriteria(CBlob@ this, CBlob@ blob, uint16 quantity)
 
 void onCollision(CBlob@ this, CBlob@ blob, bool solid)
 {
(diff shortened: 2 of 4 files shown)


mugg91:
I added a commit that applied the same logic to Crate in `CrateAutoPickup.as`.
Crate overlapping with an item that we just picked up will not yoink them from our hands.

Issue #1876, opened 2024-01-27
Sandbox: Make blobs spawned via "/spawn" assume the same facing direction as the caller


mugg91:
## Current in-game behaviour.

Ballistas and runners spawned by the player should face in the same direction as him.

In Sandbox, you can't build a soccer field and have runners face to the left because they would always face to the right by default.

mugg91:
We need to make sure certain buildings are always set to face right. 

Or we could allow mirrored buildings but need to fix some bugs:
Spritelayers' facing direction in quarters should be the same as the building.
Quarry should spawn stone depending on facing direction.

PR #1878, opened 2024-01-27
[changed] "/spawn" will spawn blobs with caller's facing direction

mugg91:
## Status

## Description

[changed] Blobs spawned with "/spawn" will assume the caller's facing direction

Fixes https://github.com/transhumandesign/kag-base/issues/1876

This PR will make blobs spawned via "/spawn" assume the facing direction of the player who used the chat command.
Quarters getting mirrored will show the player's head rotated correctly.
Quarry will spawn mat_stone in the correct position.
Making buildings "mirror proof" is probably a good thing.

And allowing blobs to get spawned not just "always facing left" but also being able to face right would add more variety to Sandbox.

## Steps to Test or Reproduce

Face left and use `/spawn <blob>`.
The spawned blob will face left.

Then face right and use `/spawn <blob>`.
The spawned blob will face left.

**After this PR, the first blob faces left and the second blob faces right.**

### DIFF
--- a/Entities/Industry/CTFShops/Quarry/Quarry.as
+++ b/Entities/Industry/CTFShops/Quarry/Quarry.as
@@ -201,9 +201,11 @@ void spawnOre(CBlob@ this)
 	int remainder = amountToSpawn % 5;
 	amountToSpawn += (remainder < 3 ? -remainder : (5 - remainder));
 	//setup res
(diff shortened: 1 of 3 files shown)


Vam-Jam:
Works online + offline

Vam-Jam (approved):

Issue #1875, opened 2024-01-27
Some Components will leave blackness when destroyed


mugg91:
## Description

For example, spawn a bunch of component tiles, such as motion sensor.

Have it be destroyed by orb, bomb or keg.

https://i.imgur.com/ByFIFcK.png

It is not that much of a problem, but it seems like buggy behavior so I made this issue.

Issue #1874, opened 2024-01-27
Obstructor hit by Keg will become bugged


mugg91:
## Description

When obstructor is hit by an exploding keg, it will cause obstructor to become bugged.

* The obstructor will not change its state between solid and non-solid anymore, even when power is supplied.
* Builder can build over the obstructor with new tiles.

It seems obstructor is not killed/despawned properly.

Happens easily in offline and online in Sandbox. Sometimes happens in TDM matches.

PR #1879, opened 2024-01-27
[fixed] Obstructor and Sensor not getting destroyed properly

mugg91:
## Status

## Description

```
[fixed] Obstructor and Sensor didn't get destroyed properly when their health depleted beyond 0 HP.
```

Fixes https://github.com/transhumandesign/kag-base/issues/1875
Fixes https://github.com/transhumandesign/kag-base/issues/1874

This PR changes the check for health in `DummyOnStatic.as` from `== 0` to `<= 0`.

I changed wording of the comment right after the health checking to make it clear _when_ Obstructor or Sensor would destroy background if it weren't for those 3 lines of code: `//builder removing Obstructor and Sensor from hand won't break background`

## Steps to Test or Reproduce

Place a buch of sensors and then `/spawn orb`.
It should look like this https://i.imgur.com/ByFIFcK.png

Place some obstructors and then either have 1 keg explode next to them or 6 bombs inside them.
You can also have 1 bomb explode inside a non-solid obstructor and then pickaxe the obstructor until destruction.
Notice the obstructor will not get destroyed properly as the sprite will remain on the map.

**After this PR, both issues are fixed. There will be no "black tiles" and obstructor will get destroyed properly.**

### DIFF
--- a/Entities/Structures/Common/DummyOnStatic.as
+++ b/Entities/Structures/Common/DummyOnStatic.as
@@ -18,7 +18,7 @@ void onSetStatic(CBlob@ this, const bool isStatic)
 				map.server_SetTile(POSITION, this.get_TileType(Dummy::TILE));
 				server_setDummyGridNetworkID(map.getTileOffset(POSITION), this.getNetworkID());
 			}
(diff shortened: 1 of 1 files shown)


Vam-Jam:
Closing in favour of #2219

Issue #1870, opened 2024-01-25
Last arrow shot doesn't make a sound


mugg91:
## Description

As archer, when shooting your last arrow, it won't make a shoot sound effect.

mugg91:
Looks like it was fixed, based on [my comment from last year ](https://github.com/transhumandesign/kag-base/pull/1880#issuecomment-2537190002)

PR #1880, opened 2024-01-28
[fixed] No sound when shooting your last arrow (on server) + Grapple sound on collision

mugg91:
## Status

## Description

```
[fixed] Last arrow doesn't make a sound effect
[added] Grapple will make a sound on collision
```

Fixes https://github.com/transhumandesign/kag-base/issues/1870

This PR changes `ArcherLogic.as` to allow a shoot sound effect when shooting your last arrow(s) in online.

Sound effect removed from command `"shoot arrow"`.
A new command `"shoot sound"` is added which is sent in `CreateArrow()`. Only the client processes the command and it will play the sound effect. Some logic was added to make it play the sound only once in a tick, not three times for legolas.

This way, the sound will play when the arrow is actually created.

Tested offline and online, no problems.

## Steps to Test or Reproduce

Shoot your last arrow in online.
Notice there is no sound.
**After this PR, there is a sound.**

### DIFF
--- a/Entities/Characters/Archer/ArcherCommon.as
+++ b/Entities/Characters/Archer/ArcherCommon.as
@@ -84,6 +84,7 @@ shared class ArcherInfo
 };
 
 const string grapple_sync_cmd = "grapple sync";
+const string grapple_sound_cmd = "grapple sound";
 
 void SyncGrapple(CBlob@ this)
 {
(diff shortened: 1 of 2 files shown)


mugg91:
I added a commit that will play a `BombBounce.ogg` sound effect when grapple collides with something.
It is processed only on client.
I tested in offline and online and it works.

I don't know if the sound will happen just with you or with all archers on the map. I have no way to test it right now but if so it should be easy to fix it.

mugg91:
I made a small change that makes the sound play amplied by 0.5f (less annoying) and at the grapple's position rather than the archer's.

mugg91:
In the middle of merging all my PRs, I saw `this.getShape().getConsts().net_threshold_multiplier = 0.5f;` in `ArcherLogic.as` being there for no reasons. It needs to be removed before you use this PR.

Vam-Jam:
If you could fix up merge conflicts, I'll review it when I get a chance (if this is still an issue)

mugg91:
I made a new PR for the grapple sound https://github.com/transhumandesign/kag-base/pull/2234
Closing this now.

PR #1885, opened 2024-01-30
[fixed] invincible players getting hit by certain damage sources

### LINKED ISSUE #1740: Archer is awarded coins when hitting invulnerable enemy player with arrow
## Description
Arrow colliding with invulnerable player causes blood sprite and I think it also makes a hurt sound.
It rewards the archer with coins.

Whereas Knight hitting slash causes no blood, no hurt sound and no coin reward.

## Suggestion

Archer should despawn after colliding with invulnerable player without causing blood, without causing hurt sound, without rewarding coins.

mugg91:
## Status

- Needs testing.

## Description

```
[fixed] Invincible player can't get hit by boulder
[fixed] Invincible player can't get hit by catapulted boulder
[fixed] Invincible player can't get hit by arrow
[fixed] Invincible player can't get hit by stomp
[fixed] Invincible player can't get hit by spikes
[fixed] Invincible player can't get hit by fire
[fixed] Invincible player can't get hit by burn
[fixed] Invincible player can't get hit by mine
[fixed] Invincible player can't get hit by shield
[fixed] Invincible player can't get hit by explosion while in a crate
[fixed] Invincible player can't get hit by ballista bolt
        Explosion will still happen although the invincible player will not get by it either
[fixed] Invincible player can't get hit by water bomb
[fixed] Invincible player can't get hit via HurtOnCollide (e.g. raft, vehicles)
[fixed] Invincible player can't get hit by being run over by vehicle
[fixed] Invincible player can't get hit by being on a spiker that activates
[fixed] Invincible player won't make stun sound when his jab is blocked
[fixed] Invincible player won't make stun sound when getting shield-slided at
[changed] Stun sound when getting shield-slided at is now emitted from the hit blob not the shield-sliding blob
```

Fixes https://github.com/transhumandesign/kag-base/issues/1740

Invincible players (e.g. those that recently spawned from tent) could get hit by certain damage sources.
Blood splatter, hurt or stun sound and knockback would happen and the attacker would get rewarded coins.

This PR fixes this by checking those damage sources for the attacked blob's "invincible" tag before going through with the attack.

Some instances of `getNet().isServer()` and `getNet().isClient()` have been changed to `isServer()` and `isClient()` respectively. One instance of `this.getMap()` changed to `getMap()`.

This PR still needs to be tested, but it should work as envisioned.

## Steps to Test or Reproduce

Go to CTF.
`AddBot("");` in console a few times.
( `/restartmap` may cause bots to turn into builders)
`getRules().SetCurrentState(GAME);`
Go to the enemy side and try - while they are invincible - :
* throwing a boulder at them
* shooting (fire) arrow at them
* having the background behind tent burn
* throwing water bomb at them
* throwing raft at them
* riding a catapult against them
* etc.

Notice they will get hit, as indicated by blood splatter, hurt/stun sound, knockback and coin gain.
**After this PR, they won't.**


mugg91:
When I jab another knight it says "Command cycle not found; Add it first in onInit()".
Not sure where is the source of the problem since I didn't touch commands in `KnightLogic.as`.

I noticed that I changed shieldstun to play the stun sound from the hitblob but jab should also play it from the hitblob.
On the last hit, the willhelm shout on death should not overlap with the stun sound.

Vam-Jam:
Would it be more ideal if we make a change engine side to have server_hit not do anything if the invincible tag is applied (or we can add a new flag like `setInvincible(bool state)`?

mugg91:
> Would it be more ideal if we make a change engine side to have server_hit not do anything if the invincible tag is applied (or we can add a new flag like `setInvincible(bool state)`?

Yes, not processing the hit if the blob is invincible is a good idea, as well as blob functions `setInvincible()` and `isInvincible()`.
But I still think it is important to early out of functions when detecting a blob is invincible. It's been a while I worked on this PR but I think it's not just about the `server_Hit()` instances, also other things around that should be avoided if blob is invincible.

mugg91:
Fixed file conflicts

Issue #1882, opened 2024-01-29
Make drill only emit steam from water hit if its heat is above 0


mugg91:
## Current in-game behaviour.

Drill shouldn't emit steam from getting by water arrow if it wasn't hot in the first place.

PR #1886, opened 2024-01-30
[fixed] drill emitting steam when there's no heat

mugg91:
## Status

## Description

`[fixed] Drill doesn't emit steam when getting hit by water and while having no heat`

Fixes https://github.com/transhumandesign/kag-base/issues/1882

## Steps to Test or Reproduce

Spawn a drill.
Shoot a water arrow at it.
Notice there is steam even though the drill had no heat.
**After this PR, there won't be steam.**

### DIFF
--- a/Entities/Industry/Drill/Drill.as
+++ b/Entities/Industry/Drill/Drill.as
@@ -446,9 +446,13 @@ f32 onHit(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@ hitt
 	if (customData == Hitters::water)
 	{
 		s16 current_heat = this.get_u8(heat_prop) - heat_max * heat_reduction_water;
-		if (current_heat < 0) current_heat = 0;
+		
+		if (current_heat <= 0) 
+			current_heat = 0;
+		else if (current_heat > 0)
+			makeSteamPuff(this);
+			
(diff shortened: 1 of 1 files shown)


mugg91:
I thought this was a simple fix but I'm encountering more problems with drill not cooling down properly.
I'm also considering adding gib to it (it just vanishes when it is destroyed).
Please wait while I address these.

Entities/Industry/Drill/Drill.as
```diff
@@ -445,10 +445,14 @@ f32 onHit(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@ hitt
 
 	if (customData == Hitters::water)
 	{
-		s16 current_heat = this.get_u8(heat_prop) - heat_max * heat_reduction_water;
-		if (current_heat < 0) current_heat = 0;
+		s16 current_heat = this.get_u8(heat_prop);
```
Vam-Jam, line 448: this should be a u8

Vam-Jam (changes requested):

Entities/Industry/Drill/Drill.as
mugg91: I applied the change and tested for a bit in offline. 
There should be no problems from changing f32 to u8 in line 136 and changing s16 to u8 in line 448.

PR #1887, opened 2024-01-31
[added] "/team _" will also convert held items with script "SetTeamToCarrier.as"

mugg91:
## Status

## Description

When switching team via `/team x` in Sandbox, held item will also switch to the new team if it has script `"SetTeamToCarrier.as"` (keg, drill, saw, boulder...)

Switching team via menu won't work. Tested online and offline successfully.

## Steps to Test or Reproduce

Go to Sandbox.
Have a drill in hand.
`/team 5` in chat.
Notice the drill will not have changed team.
**After this PR, held item will change to the new team**


mugg91:
Since this should only happen with `"/team -`", we need to edit chat command, not `RunnerDefault.as`
I will make a new PR.

PR #1888, opened 2024-01-31
[added] "/team _" will also convert held items with script "SetTeamToCarrier.as"

mugg91:
## Status

## Description

When switching team via `/team x`, held item will also switch to the new team if it has script `"SetTeamToCarrier.as"` (keg, drill, saw, boulder...)

This is a better solution than https://github.com/transhumandesign/kag-base/pull/1887

## Steps to Test or Reproduce

Go to Sandbox.
Have a drill in hand.
`/team 5` in chat.
Notice the drill will not have changed team.
**After this PR, held item will change to the new team**

### DIFF
--- a/Rules/CommonScripts/ChatCommands/PlayerCommands.as
+++ b/Rules/CommonScripts/ChatCommands/PlayerCommands.as
@@ -95,6 +95,16 @@ class TeamCommand : ChatCommand
 		}
 
 		blob.server_setTeamNum(team);
+		
+		if (blob.hasAttached())
+		{
+			CBlob@ heldblob = blob.getCarriedBlob();
+			
+			if (heldblob !is null && heldblob.hasScript("SetTeamToCarrier.as"))
+			{
+				heldblob.server_setTeamNum(team);
+			}
+		}
 	}
 }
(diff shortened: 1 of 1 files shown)


mugg91:
Inventory items will now also be converted.

Vam-Jam (dismissed):
LGTM

Rules/CommonScripts/ChatCommands/PlayerCommands.as
```diff
@@ -95,6 +95,32 @@ class TeamCommand : ChatCommand
...
+		
+		// convert items in inventory
+		CInventory@ inv = blob.getInventory();
+		if (inv !is null)
+		{
+			for (int i = 0; i < inv.getItemsCount(); i++)
+			{
+				CBlob@ item = inv.getItem(i);
+				
+				if (item !is null)
```
asumagic, line 107: Why no such `SetTeamToCarrier.as` check here?

mugg91: I can't think of a reason why not, so I included it.

Issue #1883, opened 2024-01-30
Fishy air drown should be reverted or reset if it gets back into water before dying


mugg91:
## Description

Fishy will die if on land long enough.
If it gets back in water while still alive, it will not reset the "air time".
So if Fishy gets out of water again, it will continue adding onto the "air time" instead of starting fresh from 0.

The air time should be reset or reverted at the same rate as it is added onto on land.

PR #1889, opened 2024-01-31
[fixed] Fishy airdrown gets reverted in water over time

mugg91:
## Status

## Description

```
[fixed] airdrown didn't get reverted over time if fishy got back in water
```

Fixes https://github.com/transhumandesign/kag-base/issues/1883

Prior to this PR, `AirDrown.as` (which is a blob script for Fishy) added 15 `airdrown_ticks` every 15 ticks Fishy was outside of water. When Fishy got back in water, the value stayed the same.

This PR makes the following changes:
* Script runs both inside and outside of water
* Every 15 ticks, 15 `airdrown_ticks` are added if outside of water or depleted if inside water.
* When blob is dead, script is removed.
* I left `onRender()` test code commented out, just in case someone wants to test it in the future. `AirDrown.as` needs to be added as a sprite script to the blob (e.g. in `Fishy.cfg`) for it to work.

## Steps to Test or Reproduce

Spawn a Fishy.
Have Fishy be outside of water for a few seconds.
Let Fishy get back in water just before dying.
Wait for a few seconds.
Get Fishy out of water.
Fishy will die very soon.
**After this PR, Fishy's airdrown is reverted in water over time so he will not die just after getting outside of water.**


mugg91:
We can remove the commented out `onRender()` if you want.

PR #1896, opened 2024-02-05
[fixed] Camera issues

### LINKED ISSUE #1512: Camera gets stuck in TDM when switching from spectator to player
## Description

Play on TDM server.
Have a new match start and go spectator.
Make sure you haven't died in the current match and there are other players so the match keeps going in your absence.
Switch from spectator to player.
Notice it says "No respawning - wait for match to end" and you can't move the camera.

## Needs Testing

CTF/TTH: Go from spectator to player after match has started (no red barrier), check if this issue happens there, too.

## Additional

Might be related PR: #1123

### LINKED ISSUE #1862: Camera for spectator should start in center of map instead of at the top left (CTF)
## Description

1. Play in a team in (small) CTF
2. Go spectator.
3. Wait for the match to end and a new map to load
4. Notice the camera will be positioned in the top left

## Suggestion

Make the camera start in the center, possibly zoomed out.

mugg91:
## Status

- Needs testing

## Description

```
[fixed] Camera getting stuck when going spectator, a new match starts and then switching from spectator to team
[fixed] Camera couldn't scroll as spectator in a match with bots
[fixed] Camera would scroll to somewhere else when moving character forward and then going spectator
[fixed] Camera would initialize in position (0, 0) instead of center map (in online)
```

Fixes https://github.com/transhumandesign/kag-base/issues/1862
Fixes https://github.com/transhumandesign/kag-base/issues/1512

This PR does some changes to address some issues. 

The reason for camera initialising in position (0, 0) when a new match loads while spectator is with `ViewEntireMap()` being called in `Reset(CRules@ this)` at the time map is not initialised yet, so map dimension is still (0, 0). I changed the code so that `ViewEntireMap()` is called in `onInit(CMap@ this)` instead which allows the camera to initialize in center of map properly.

I moved `ViewEntireMap()` from `CinematicCommon.as` to `Spectator.as` for no particular reason. There is no consequence in having it in one or the other script.

There are many issues with bots but I addressed one I found annoying. When in a match with bots and you go spectator, camera would be stuck on one of the bots and you couldn't move the camera. This PR fixes that by adding a check for `!targetPlayer().isBot()` in `Spectator.as`. Right after that, I moved `posActual = camera.getPosition();` inside the previous code block which seems to make camera movable in situations where it previouslyn wasn't. This seems to fix camera getting stuck when switching from spectator to team and waiting the respawn timer, after a match has gone from build phase to battle phase.

Tested offline and online. However it should be tested with other human players in an online server to check for regressions.

## Steps to Test or Reproduce

I added this in `TipCommand` in `MiscCommands.as` for testing purposes:

```
		if (index == 1)
		{
			getRules().SetCurrentState(GAME);
		}
		else if (index == 2)
		{
			getRules().SetCurrentState(GAME_OVER);
		}
		else if (index == 3)
		{
			AddBot("");
		}
```

1) 
Join CTF online server.
`/tip 4` (add bot)
Go spectator.
Notice the camera is stuck on the bot.
**After this PR, you can freely move it around.**

2)
Join CTF online server.
`/tip 3` (game over)
Go spectator.
When the new map loads, camera will be in top left corner.
**After this PR, it will be centered on the map.**

3)
Join CTF online server.
Move a distance away.
Go spectator.
Notice the camera would scroll back to the tent for odd reasons.
**After this PR, the camera would stay where you despawned.**

4)
Join CTF online server.
Go spectator.
`/tip 3` (game over)
Wait for new map to load.
`/tip 2` (battle phase start)
Join a team.
Notice the camera will be stuck.
**After this PR, it is movable.**

### DIFF
--- a/Rules/CommonScripts/CinematicCommon.as
+++ b/Rules/CommonScripts/CinematicCommon.as
@@ -287,20 +287,6 @@ bool focusOnBlob(CBlob@[] blobs)
 	return false;
 }
 
-void ViewEntireMap()
-{
-	CMap@ map = getMap();
-
-	if (map !is null)
-	{
-		Vec2f mapDim = map.getMapDimensions();
-		posTarget = mapDim / 2.0f;
-		Vec2f zoomLevel = calculateZoomLevel(mapDim.x, mapDim.y);
-		zoomTarget = Maths::Min(zoomLevel.x, zoomLevel.y);
-		zoomTarget = Maths::Clamp(zoomTarget, 0.5f, 2.0f);
-	}
-}
-
 bool cinematicEnabled = true;
 bool cinematicForceDisabled = false;
 
--- a/Rules/CommonScripts/PlayerCamera.as
+++ b/Rules/CommonScripts/PlayerCamera.as
@@ -146,23 +146,30 @@ void SpecCamera(CRules@ this)
 {
 	//death effect
 	CCamera@ camera = getCamera();
-	if (camera !is null && getLocalPlayerBlob() is null && getLocalPlayer() !is null)
+	if (camera !is null && getLocalPlayer() !is null)
 	{
-		const int diffTime = deathTime - getGameTime();
-		// death effect
-		if (!spectatorTeam && diffTime > 0)
+		if (getLocalPlayerBlob() is null)
 		{
-			//lock camera
-			posActual = deathLock;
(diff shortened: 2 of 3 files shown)


mugg91:
It seems camera getting stuck with bots only happens when bot has an empty "" name.

Issue #1513, opened 2022-11-24
Gold despawning in Sandbox will trigger Admin Alert


mugg91:
## Description

Go to offline Sandbox.
`!gold`
Wait for gold to decay.
When the last pebble of gold disappears, it say something like: "8 gold fell into the void"
or if you held it in hand before: "Player x caused 8 gold to fall into the void"

## Proposed Behavior

Don't show this admin alert message when gold decays by itself.

mugg91:
In theory this can happen in TDM servers, too, if you wait long enough. 
Because `TDM.as` doesn't remove `DecayQuantity.as` from gold like the other gamemodes do but TDM rules aren't tagged `"quick decay"`.
Therefore `DecayQuantity.as` is in effect but only ticks every 66 seconds (2000 ticks).

mugg91:
The bug can be seen in this video.
https://www.youtube.com/watch?v=e-rpXITQNA0

eps0003:
I took a quick look into how this could be resolved but I couldn't think of a straightforward solution that covers all decay situations. For example, `DecayUnused.as` uses `server_SetTimeToDie()` which prevents us from tagging the blob with `"AdminAlertIgnore"` right before it dies.

Even if it were as simple as tagging the blob to fix this issue, we would still need to do the same for every other situation that destroys gold but shouldn't alert an admin. Ideally, we shouldn't be tagging blobs with `"AdminAlertIgnore"` in various scripts. Instead, we should be tagging blobs with the way they are destroyed (e.g. `"decayed"`, `"voided"`, `"burned"`, etc.) so `AlertAdminOnDie.as` can be the one to decide which methods should alert admins.

Gingerbeard5773:
Was this resolved?

AsuMagic:
> Was this resolved?

Doesn't look like it.

PR #1898, opened 2024-02-06
[fixed] Gold decaying would alert admin (TDM)

mugg91:
## Status

## Description

```
[fixed] False admin alert when gold vanished by decaying
```

Fixes https://github.com/transhumandesign/kag-base/issues/1513

Pebble of gold could decay after 66 sec in TDM causing a false admin alert message.

Although @eps0003 suggested that blobs should be tagged by the way they are destroyed and letting `AlertAdminOnDie.as` decide which tags are getting reported, this would mean changing every script that has `server_Die()` or `server_Hit()` in it.

For now, I simply added `this.Tag("AdminAlertIgnore");` to `DecayQuantity.as` so the specific situation of gold decaying and alerting admin doesn't happen.

Admin alert message is for when players throw gold into the void, not for when materials decay by themselves.

Note: Tag `"AdminAlertIgnore"` is also used for when materials merge, which is a similar use-case.

## Steps to Test or Reproduce

Go to TDM.
Slash at a gold tile so a pebble of gold falls out.
Wait 66 sec so the gold pebble despawns.
Admin alert message can be read in console log.
**After this PR, there is no admin alert message.**

### DIFF
--- a/Entities/Common/Decaying/DecayQuantity.as
+++ b/Entities/Common/Decaying/DecayQuantity.as
@@ -27,6 +27,7 @@ void onTick(CBlob@ this)
 
 	if (step >= quantity)
 	{
+		this.Tag("AdminAlertIgnore");
 		this.server_Die();
 		return;
 	}


mugg91:
> It would be better if we had alert only work on specific gamemodes, rather than just marking decaying gold as AdminAlertIgnore

Perhaps, but decaying should still not alert admins since no player caused it.

Vam-Jam (approved):

Issue #1300, opened 2022-06-28
Changing class should copy "on fire" state


mugg91:
## Description

When changing class while on fire, you will no longer be on fire.
Since even stun is copied to the new Blob, I think fire should be copied, too.
I was unable to implement this, so I have to open this issue.

Adding this code to StandardRespawnCommand.as makes the new Blob to be on fire forever without losing health:
```
#include "FireCommon.as";
#include "Hitters.as";

// ...

//copy fire
if (caller.hasTag(burning_tag))
{
	newBlob.getShape().getConsts().isFlammable = true;
	newBlob.Tag(burning_tag);
	newBlob.set_s16(burn_duration, caller.get_s16(burn_duration));
	newBlob.set_s16(burn_timer, caller.get_s16(burn_timer));
	newBlob.set_u8(burn_hitter, Hitters::burn);
}
```

Adding a line `server_setFireOn(newBlob);` will cause a null pointer access exception error.

## Reproduction

Go to Sandbox, spawn a Tent or TDM_spawn, be Archer and shoot a fire arrow into the grass.
When on fire, try to change class near the Tent/Spawn.

PR #1905, opened 2024-02-11
[fixed] "on fire" state wasn't copied when changing class

mugg91:
## Status

## Description

```
[fixed/added] "on fire" will be copied when changing class
```
Fixes https://github.com/transhumandesign/kag-base/issues/1300

This PR makes it so that changing class will not get rid of you being on fire. Instead your fire state will be copied to the new blob.

Because `this.getCurrentScript().runFlags` etc. would cause an exception when the script it's in is being used as an "include script" rather than a "blob script", I removed it and made `IsFlammable.as`'s `onTick()` only run when the condition `!this.hasTag(burning_tag) && !this.isInFlames()` is false.

Tested online and offline and it works as it should.

## Steps to Test or Reproduce

Spawn a tdm_spawn.
Be archer.
Shoot fire arrow into the grass and wait for it to burn and set you on fire.
Run into the tdm_spawn and change class.
Notice the burning will keep going as long as it would have without changing class.


GELGELAT:
On one hand this mechanic was cool, on the other hand you could say it was a bug. Honestly I would have left it as it was, because it was used by knowledgeable people and it showed the skill of the player that he knows the mechanics.

Vam-Jam:
Works from testing, marking as discussion for now

Issue #1699, opened 2023-09-27
Inconsistent font size on tips pane (ctf spectating state)


NoahTheLegend:
**Description**

When a match is ended, there is font changed to much bigger one, somewhere before rendering tips

**Reproduction**

Join ctf, stay with enabled tips and die, then trigger the "match" to end
Notice that tips font changed and go out of background pane bounds

**Expected Behavior**

put GUI::SetFont("menu") before rendering tips

mugg91:
I tested in offline.
Being dead and then using chat command `/endgame` will not cause the issue.
But being dead and then the ctf flag being captured will cause it.

It looks like this - https://i.imgur.com/hfx2Z6u.png

mugg91:
https://i.imgur.com/pHzI66z.png
Top right button

PR #1906, opened 2024-02-12
[fixed] wrong font being used on tips and queue button when game state banners are shown

mugg91:
## Status

## Description

```
[fixed] wrong font being used on tips and queue button when game state banners are shown
```
Fixes https://github.com/transhumandesign/kag-base/issues/1699

Tested online, works.

## Steps to Test or Reproduce

You can add
```
		if (index == 1)
		{
			getRules().SetCurrentState(GAME);
		}
		else if (index == 2)
		{
			getRules().SetCurrentState(GAME_OVER);
		}
```
in `TipCommand` in `MiscCommands.as`.

Go to CTF.
Place spikes.
Jump in the spikes so the next jump would kill you.
Still during build phase, type `/tip 2` = battle phase start, banner is shown. Then ...

... Quickly jump in the spikes again.
Notice the font on the tip is wrong.
**After this PR, it is correct.**

... Quickly go spectator.
Notice the font on the queue button is wrong.
**After this PR, it is correct.**

### DIFF
--- a/Rules/CommonScripts/ShowTipOnDeath.as
+++ b/Rules/CommonScripts/ShowTipOnDeath.as
@@ -56,6 +56,7 @@ void RenderTip()
 	Vec2f br(scrw / 2 + w / 2, scrh - offset);
 
 	GUI::DrawButton(tl, br);
+	GUI::SetFont("menu");
 	GUI::DrawText(tip, tl + Vec2f(10, 10), br - Vec2f(10, 10), color_white, true, true, false);
 }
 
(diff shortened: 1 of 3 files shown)


mugg91:
Added a commit that places `GUI::SetFont("menu");` before `GUI::GetTextDimensions(vote_title, text_dim);` in `VoteCore.as` to fix the vote window looking like this:

screen-24-04-03-21-07-46

asumagic (approved):

Issue #1739, opened 2023-11-05
Archer can shoot arrows through red barrier


mugg91:
## Current in-game behaviour.

In small CTF, on some maps Archer can shoot arrows through red barrier during build time.

## Proposed Change

Shooting arrows through red barrier should not be possible.
Stronger force should be applied to repel the arrows.

PR #1913, opened 2024-02-16
[fixed] Archers can not shoot through red barrier

mugg91:
## Status

## Description
```
[fixed] Prevent archer from shooting through red barrier
```
Fixes https://github.com/transhumandesign/kag-base/issues/1739

This PR changes the logic of `RedBarrier.as` to prevent archers from shooting through red barrier.

This is done by checking for spawn points (tent, hall) and pushing the blob towards the spawn point.
That means, we are not pushing from the center to the side, but we are pushing toward the spawn point that matches the blob's team num, for the entire distance of red barrier.
Even on maps with very thin red barrier (e.g. `NoahTheLegend_Stalactite`) and using a trampoline, I wasn't able to shoot arrow through the red barrier.

If a spawn point isn't found (which should never be the case, but just in case it does happen) the old way will be used, pushing from the center of red barrier.

Tested briefly in offline.

## Steps to Test or Reproduce

Go to Small CTF maps, e.g. `NoahTheLegend_Stalactite`.
Assuming both team's tents are placed properly.
Be archer.
Shoot through the red barrier.
Notice you will easily shoot through the red barrier.
**After this PR, the arrows will not make it through the red barrier.**

### DIFF
--- a/Rules/CommonScripts/RedBarrier.as
+++ b/Rules/CommonScripts/RedBarrier.as
@@ -90,7 +90,7 @@ void onTick(CRules@ this)
 				(b.getTeamNum() < 100 || b.hasTag("no barrier pass") || 
 				 b.hasTag("material") || b.getName() == "spikes"))
 			{
-				PushBlob(b, (tl.x + br.x) * 0.5, tl.x, br.x);
+				PushBlob(b, tl.x, br.x);
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


Gingerbeard5773 (approved):
Tested on server, found no issues regardless of CTF map I tested.

Vam-Jam:
That or we can just increase the pushback vel the closer you are to the center instead of having to calculate spawns and what not

Issue #1914, opened 2024-02-17
Make lamp in sandbox shine in team color


mugg91:
## Current in-game behaviour.

Instead of always glowing yellow, I think it would be nice to have it shine in team color.
So blue team lamps would shine blue, red team lamps would shine red, etc.

mugg91:
I'm thinking if motion sensor should receive the same treatment...

PR #1916, opened 2024-02-17
[added] Lamp is now team-colored

mugg91:
## Status

## Description
```
[added] Lamp is now team-colored
```
Fixes https://github.com/transhumandesign/kag-base/issues/1914

This PR makes Lamp shine in its team-color rather than always yellow.

I added a spritelayer and set it to use `RenderStyle::light` when lamp is turned on. This allows for a better lighting effect. I didn't set the sprite to use `RenderStyle::light` directly because it causes the blob to become extremely bright when hitting it by pickaxe.

Looks like this in-game: https://i.imgur.com/tJ1r6uN.png

The file `Lamp.png` was reduced in size by 265 bytes, 59%.

## Steps to Test or Reproduce

Switch to any team, then place some lamps and turn them on.


mugg91:
Although minor, there is still an issue that needs looking into: https://github.com/transhumandesign/kag-base/issues/1950

Vam-Jam (approved):
Works correctly

PR #1917, opened 2024-02-21
Trampoline rework

MrPotatobird:
## Status

- **NEEDS TESTING**

PR is probably ready to merge, but it's a lot of changes and could use more testing and balance opinions.

I can break this PR up if needed

## Description

**Motivation:**
- Trampolines set up on top of a tower are kind of OP at flinging people across the map
- If you use a trampoline in any other situation you're probably either feeding or annoying your team
- They're just really finicky to use

**Goals:**
- Make them less effective horizontally - they should be better than jumping but not comparable to a bombjump or catapult
- Make them more reliable in other situations, like bouncing your teammates up a wall
- Make them more convenient to set up the way you want
- Make them less annoying to be around by giving players more control over when and how they bounce

**Changelog**
```
mod/tramp/nerf
	[modified] Trampoline nerf - Can't glide as far horizontally

mod/tramp/buff
	[modified] Hold W to bounce higher up than before on a trampoline

[modified] Can only pick up enemy trampolines from "below" (ala saw)
[modified] Allow RMB when holding trampoline (shield/grapple/pickaxe)
[modified] Trampolines have more health
[modified] Thrown trampolines will bounce things on the ground
[added] Destruction sprite for trampoline
[modified] Trampolines float better and don't decay in water
[modified] Hold W to pass through downward-facing team trampolines

mod/tramp/jump
	[modified] Tap W to bounce from below on a trampoline you're touching
[modified] Manually bouncing on overlapping tramps won't launch as high

mod/tramp/feet
	[modified] Freeze held trampoline's angle with spacebar
	[modified] Trampolines can be set up with feet so they don't roll
	[modified] Trampoline feet get ripped off by explosions
	[modified] Add mapping support for alpha trampolines

mod/tramp/fold
	[modified] Trampoline auto folds while it's unlocked + held
	[modified] Folded trampolines weigh less
	[modified] Press E when holding folded tramp to unfold and rotate slowly
	fix/tramp/buttons
		[fix] not being able to click buttons while holding trampoline
[modified] Folded trampolines can be caught by enemies from any angle

mod/tramp/physics
	[modified] Trampoline physics are more natural
	[modified] Use old tramp physics if closer to held direction (WASD)

fix/tramp-stun
	[fix] Fix taking fall damage on trampolines
	[fix] Kegs from blowing up on trampolines
	[fixed?] wall run overriding trampo bounce (thanks mugg91)

[fixed?] Arrows not consistently bouncing off of trampolines
```
Includes #1689 and I cherry-picked #1897 because it seems to work and I already had a prop for that

**Possible Issues:**
- Having to pickup enemy tramps from below is pretty obtuse, and could probably use some kind of indicator. 
- Engine bug: When you're still touching a trampoline but stop overlapping one of its feet, you no longer count as overlapping the trampoline (issue with `AddShape`) which makes manually triggering a bounce inconsistent
- Tramps in water tend to wobble a little more than I'd like.
- Could maybe add a help for pressing E on the trampoline to unfold it? The permanent helps already kind of annoy people
- The `AddShape` function used with the feet can apparently cause crashes, although I haven't had that happen. If needed I could revert the feet to be a visual layer and just prevent the trampoline from rotating after throwing. This would be a little more unnatural
- Arrows still tend to pierce trampolines
- Putting feet on an upside down trampoline doesn't really do anything. I could try to make it so the feet stick out of the sides or something but it would be kind of a PITA

## Steps to Test or Reproduce

- Bounce yourself and items off of the trampoline at different angles/velocities
- Try picking up and throwing allied/enemy/map trampolines from different angles when folded/unfolded
- Use Spacebar to add/remove the trampoline's feet
- Press E to unfold a trampoline without locking its angle
- Tap WASD to manually bounce on an overlapping tramp - some angles will require tapping two buttons at once, eg. W+A to bounce on an up+left tramp


TerminalHash:
this pr need rework after command system update, currently will work only with `legacy_cmd = true`

Issue #1532, opened 2022-12-10
NoAmmo sound when buying or building with just enough materials/coins


mugg91:
## Description

While playing in an online server, especially in a high activity CTF server, if you have just enough coins/materials and buy something, there will sometimes be a NoAmmo sound.
E.g. you have exactly 30 coins and buy fire arrows for 30 coins.

When you have exactly 150 wood and you build a building for 150 wood, there will always be a NoAmmo sound.

Needs testing.

mugg91:
This might be a false alarm, like when trying to buy stuff as Archer when having no arrows.
Still not completely sure.

PR #1918, opened 2024-02-22
[fixed] Sounds not playing when building a building

mugg91:
## Status

## Description

Partially fixes https://github.com/transhumandesign/kag-base/issues/1532

This PR fixes these problems:
* When having 150 wood remaining and building a building, a "noammo" sound would play and the "construct" sound wouldn't play both in offline and online. Now the "noammo" sound should play only when the building isn't created and the "construct" sound will play when creating a building with 150 wood remaining.
I might provide more information later.


Gingerbeard5773:
This has been resolved by #2116 , this can be closed

PR #1927, opened 2024-02-25
[changed] You won't get stomped by the blob that you recently stomped

### LINKED ISSUE #1408: Make it so you don't get stomped by an enemy that you stomped
## Current in-game behaviour.

You fall on an enemy to stomp him.
You somehow fall beneath the enemy.
The game registers the stomp and applies downward velocity to the enemy.
The enemy stomps you (and you die).

## Proposed Change

Add a timer to the stomped enemy of 30 ticks in which he cannot stomp someone else.

## Additional context

Stomping someone should be rewarding, not punishing.

mugg91:
## Status

## Description
```
[changed] You won't get stomped by the blob that you recently stomped
```

Fixes https://github.com/transhumandesign/kag-base/issues/1408

This PR changes `Stomp.as`.
The blob that got stomped will have stomp timestamp and stomper netid set to it.

`doesCollideWithBlob()` hook is added to `Stomp.as` which checks if we have any timestamp and stomper info.
Collision doesn't happen if the blob we try to collide with is the blob that recently (within 60 ticks) stomped us and if we are traveling downward faster than that blob.

Tested in offline only. I suggest to test it in live matches.

## Steps to Test or Reproduce

1) Build something like this https://i.imgur.com/59IY7rn.png
2) Spawn knight on the ladder at the top tunnel.
3) Change team
4) Travel to the top and jump on the knight you spawned, try to tap the "use" key so you travel the tunnel.
5) If you successfully traveled, you will notice the knight you just stomped with fall through you and will not stomp you. On the next bounce after that, he will stomp you though.

### DIFF
--- a/Entities/Characters/Scripts/Stomp.as
+++ b/Entities/Characters/Scripts/Stomp.as
@@ -2,6 +2,8 @@
 #include "/Entities/Common/Attacks/Hitters.as";
 #include "KnockedCommon.as"
 
+const u8 STOMP_AGAIN_THRESHOLD = 15;
+
 void onCollision(CBlob@ this, CBlob@ blob, bool solid)
 {
 	if (blob is null)   // map collision?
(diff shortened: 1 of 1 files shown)


Vam-Jam (changes requested):
60 ticks is way too long from my testing, should be reduced to 15.

mugg91:
> 60 ticks is way too long from my testing, should be reduced to 15.
I applied your suggestion.

Issue #1939, opened 2024-03-11
/seed should allow flowers, grain_plant and bush


mugg91:
## Current in-game behaviour.

"/seed" only allows bushy and pine tree.

## Proposed Change

Should also allow flowers, grain_plant and bush.

PR #1941, opened 2024-03-14
[added] "/seed" chat command can now spawn grain plant, flowers and bush

mugg91:
## Status

## Description
```
[changed] "/seed" can now be used to spawn grain_plant, bush and flowers, too. "/tree" removed.
[changed] When "/seed" is used without a type, it will spawn a random seed rather than a pine tree seed.
```
Fixes https://github.com/transhumandesign/kag-base/issues/1939

This PR makes changes to `BlobCommands.as` to allow `/seed` to be able to spawn `grain_plant`, `bush` and `flowers`.

When only `/seed` is used without an argument, it will spawn a random seed instead of always a pine seed.

Removed `/tree` since the command is now also for types of seeds that aren't trees. The user will have to type in `tree_pine` and `tree_bushy` instead of just `pine` and `bushy`, as a small side effect.

Tested in offline, works without problems.

## Steps to Test or Reproduce

Type `/seed`
Type `/seed test`
Type `/seed tree_pine`
Type `/seed tree_bushy`
Type `/seed grain_plant`
Type `/seed bush`
Type `/seed flowers`

### DIFF
--- a/Rules/CommonScripts/ChatCommands.cfg
+++ b/Rules/CommonScripts/ChatCommands.cfg
@@ -18,7 +18,7 @@ commands =
 	allarrows; 0; 1;
 	fishies; 0; 1;
 	chickens; 0; 1;
-	tree; 0; 1;
+	seed; 0; 1;
 	water; 1; 1;
 	crate; 0; 1;
 	scroll; 0; 1;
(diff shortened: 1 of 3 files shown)


Gingerbeard5773 (approved):
tested and works as intended

Vam-Jam (approved):

Issue #1937, opened 2024-03-11
/scroll command should tell valid names


mugg91:
## Current in-game behaviour.

Type "/scroll" or "/scroll some_name".
Nothing happens.
"Scroll not found" error in logs.

## Proposed Change

Game should tell you in chat valid names for scroll.
"/scroll" should maybe spawn a default scroll.

mugg91:
Short summary what needs to be done.
The command doesn't work in modes other than TTH, period. So need to find out why and fix that.

The next problem is, even if the scroll does spawn, special scrolls like "midas" do work as intended, but "saw" and "drill" don't as the button to use the scroll won't show up.

`getButtonsFor()` is added to special scrolls but not to normal scrolls. So need to fix that, too.

Issue #1944, opened 2024-03-15
Dinghy scroll is invisible


mugg91:
## Description

1. go to TTH tutorial
2. `/scroll dinghy`
3. Invisible scroll should spawn. You can still pick it up and you can use it in a tradingpost.

## Suggestion

Make it have a visible frame.

PR #1945, opened 2024-03-15
[fixed] "/scroll <name>" chat command not working

mugg91:
## Status

## Description

```
[fixed] "/scroll <name>" will now work in modes other than TTH
[changed] "/scroll <name>" with an invalid name will list all valid names in chat
```

Fixes https://github.com/transhumandesign/kag-base/issues/1937

This PR changes `BlobCommands.as`.

The command `/scroll <name>` didn't work outside of TTH mode.
I fixed it by executing `SetupScrolls()` if `"all scrolls"` doesn't exist in CRules@.
In testing, `SetupScrolls()` will only be executed once, the first time you try to use `/scroll <name>`.

When an invalid name is used, all valid names will be listed in chat.

When no name is used, a random scroll will be spawned.

This is the same behavior as `/seed <name>` from the seed PR here: https://github.com/transhumandesign/kag-base/pull/1941

## How to use

`/scroll ` --> a random scroll spawns
`/scroll abc` --> says in chat to specify a valid name and lists all valid names
`/scroll midas` --> spawns a midas scroll.

Notice scrolls will actually spawn in Sandbox rather than nothing happening.

### DIFF
--- a/Rules/CommonScripts/ChatCommands/BlobCommands.as
+++ b/Rules/CommonScripts/ChatCommands/BlobCommands.as
@@ -85,8 +85,30 @@ class ScrollCommand : BlobCommand
 			SetupScrolls(getRules());
 		}
 
-		// no name specified, show all valid names in chat
+		// no name specified, spawn a random scroll
 		if (args.size() == 0)
+		{
+			ScrollSet@ allScrolls = getScrollSet("all scrolls");
+			
+			if (allScrolls !is null)
+			{
+				string[] scrolls_list = allScrolls.names;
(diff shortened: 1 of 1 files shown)


mugg91:
Note that this allows midas scroll (turns stone ore into gold ore nearby), drought (removes nearby water) and carnage (destroys nearby players) in Sandbox. I suggest no blacklisting for now, but we should try and see if it becomes too much of a grief in next build.

The issue with dinghy scroll being invisible ( https://github.com/transhumandesign/kag-base/issues/1944 ) still needs addressing.

Vam-Jam (approved):

Issue #1921, opened 2024-02-24
Devices/components don't have minimap tile


mugg91:
## Description

Devices such as obstructor don't show up on the minimap.

mugg91:
Ok, the issue is actually just with obstructor, which shows up as a sky blue tile when it is solid and dark grey when it is non-solid.

Should be vice versa. Instead of sky blue, should show the background tile.

Issue #1948, opened 2024-03-16
isTileGroundBack() and isTileGroundStuff() don't return true for tiles that are placed on top of dirt back tiles


mugg91:
## Description

While trying to fix https://github.com/transhumandesign/kag-base/issues/1921 , I found out that the CMap@ functions `isTileGroundBack()` and `isTileGroundStuff()` and the condition `tile == CMap::tile_ground_back` all return false if the tile in question is placed on top of a dirt back wall whereas I think they all should return true.

When an obstructor is placed on a dirt wall, the game will return false for the three conditions listed above.
This makes it impossible to recognize dirt backwalls when assigning tile colors for the minimap in `MinimapHook.as`.

Gingerbeard5773:
This suggested change is unnecessary. You can correctly identify if a tile has a dirt backwall using ``uint8 getTileDirt(uint offset)``. this function will return ``80`` if the tile has dirt.
Close?

PR #1949, opened 2024-03-16
[fixed] Obstructor has proper minimap colors now

mugg91:
## Status

## Description
```
[fixed] Obstructor has proper minimap tiles depending on its state
```
Fixes https://github.com/transhumandesign/kag-base/issues/1921

Obstructor had a minimap color `tile_castle_back` when it was non-solid, and had none, defaulting to sky color, when it was solid.

This PR changes `MinimapHook.as` to make Obstructor have a minimap tile `tile_castle_back` when it is solid and a brighter `tile_castle_back` color when it is non-solid.

I would like to make Obstructor show up as nothing when it is non-solid (i.e. sky color when it has nothing behind it, or dirt color when it has dirt backwall behind it). But there is no way to recognize such ( see: https://github.com/transhumandesign/kag-base/issues/1948 ) so this PR is the best solution for now.

I didn't see any regressions that arose from the changes.

### DIFF
--- a/Scripts/MapLoaders/MinimapHook.as
+++ b/Scripts/MapLoaders/MinimapHook.as
@@ -55,23 +55,29 @@ void CalculateMinimapColour( CMap@ map, u32 offset, TileType tile, SColor &out c
 	{ 
 		col = color_castle;
 	} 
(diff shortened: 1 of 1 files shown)


mugg91:
AAA
BBB

Vam-Jam (approved):

Issue #1936, opened 2024-03-08
When TDM match is over, don't subtract coins for buying


mugg91:
## Current in-game behaviour.

When TDM match has ended, do either of the following:

- When buying, make it not subtract any coins
- Disable buying

Buying an expensive item and getting your coins depleted just at the same time your teammates already handled the enemy team is stupid.

PR #1952, opened 2024-03-16
[added] You can't purchase items after match ended (TDM)

mugg91:
## Status

## Description
```
[added] You can't purchase items after match ended in TDM
```
Fixes https://github.com/transhumandesign/kag-base/issues/1936

Bool `no purchase post match = true` is written to CRules@ in `TDM.as` in `Reset()`.

A new function `AddMatchNotEndedRequirement()` is added to `Requirements.as` which fires when the match is over and the above bool exists in CRules@.

`TDM_Trading.as` applies `AddMatchNotEndedRequirement()` for each item that is set up in trading post via `AddItemForCoin()`.

This means you cannot buy any items after a match is over in TDM, removing annoying and unnecessary situations where a player wastes his coins post match. Since the bool is only set in `TDM.as`, the requirement only exists in TDM. You can still buy items post match in other game modes, including Challenges - which also uses trading posts and `AddItemForCoin()`.

Tested in online and offline, works as intended.

## Steps to Test or Reproduce

Play TDM.
Buy items as usual in trading post.
Win a match.
Try to buy items in trading post, get told "Match has ended" and all item buttons are greyed out.

### DIFF
--- a/Entities/Common/Includes/Requirements.as
+++ b/Entities/Common/Includes/Requirements.as
@@ -213,7 +213,7 @@ bool hasRequirements(CInventory@ inv1, CInventory@ inv2, CBitStream &inout bs, C
 		{
 			CBlob@ blob = inv1 !is null ? inv1.getBlob() : null;
 			CRules@ rules = getRules();
-			if (rules !is null && rules.exists("no purchase post match") && rules.isGameOver())
+			if (rules !is null && rules.exists("no purchase post match") && rules.get_bool("no purchase post match") && rules.isGameOver())
 			{
 				AddMatchNotEndedRequirement(missingBs);
 				has = false;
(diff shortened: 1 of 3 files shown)


Entities/Common/Includes/Requirements.as
```diff
@@ -123,6 +132,11 @@ bool ReadRequirement(CBitStream &inout bs, string &out req, string &out blobName
 	{
 		return true;
 	}
+	
+	if (req == "match not ended")
+	{
+		return true;
+	}
```
Vam-Jam, line 139:
```suggestion
        if (req == "hurt" || req == "match not ended")
	{
		return true;
	}
```

Entities/Common/Includes/Requirements.as
```diff
@@ -195,6 +209,16 @@ bool hasRequirements(CInventory@ inv1, CInventory@ inv2, CBitStream &inout bs, C
 				has = false;
 			}
 		}
+		else if (req == "match not ended")
+		{
+			CBlob@ blob = inv1 !is null ? inv1.getBlob() : null;
```
Vam-Jam, line 214: You can remove this var as you do not use it

Vam-Jam (changes requested):
Seems to work fine in game, does not break other gamemodes.

Issue #1697, opened 2023-09-27
We cant collect coins on too low map position


longwoodbow:
**Description**

We can't collect dropped coins on too low map point.

**Reproduction**

1. Go to the TDM map bunnie_Wojciech_Jaruzelski_v2
2. Open the chest
3. We can't collect coins from chest

**Additional context**

When I was making maps, I suffered about this issue, then I tried to lift up the height of ground.
I guess 7 blocks height is the limit we can collect dropped coins.

Vam-Jam:
Engine patch submitted

PR #1962, opened 2024-03-19
[fixed] Couldn't collect coins from chests in "bunnie_Wojciech_Jaruzelski_v2"

mugg91:
## Status

## Description
```
[changed] map bunnie_Wojciech_Jaruzelski_v2 as a workaround to the chest bug
```
Fixes https://github.com/transhumandesign/kag-base/issues/1697 (work around, not an actual fix)

On said TDM map, because of a bug, you can't collect coins from the chests because the map position is too deep.
This PR adds more tiles to the bottom as a workaround fix for the bug.

The old and new png are the same size.

Note that the original map in the dormant folder, `bunnie_Wojciech_Jaruzelski` is left untouched.

## Repro

Load map `bunnie_Wojciech_Jaruzelski_v2`
Open chests
You can't collect coins
**After this PR, you will be able to collect the coins**


Vam-Jam:
Closing, engine side patch has been submitted instead.

Issue #1960, opened 2024-03-18
Many TDM spawns are misplaced


mugg91:
## Description

On a lot of TDM maps, the spawn is one tile inside the ground.

PR #1963, opened 2024-03-19
[fixed] TDM_spawn and tent won't be floating above ground or be one tile too deep

mugg91:
## Status

## Description
```
[fixed] TDM_spawn and tent will have the correct position
```
Fixes https://github.com/transhumandesign/kag-base/issues/1960

This PR adds `shape.PutOnGround()` to `Init(CBlob@ this)` for tdm_spawn and tent.
So if the pixel for the "main spawn" in a png file is too low or too high, it will not matter as the blob will automatically place itself on the ground.

During testing I have not seen it misbehave. The only condition for this to work is that there is enough empty space (64x64 for tdm_ruins and 40x40 for tent).

The alternative would have been to change like 30 png files which is much painful work. And map makers will mess up pixels in the future so this generic solution should be good.

When spawning tdm_ruin or tent in Sandbox, naturally they will warp down to the next nearby ground. Good enough, since tents and ruins aren't supposed to be floating anyway.

## Steps

Load a map with the correct pixel position for tdm_spawn, e.g. `RattleCage`
Load a map with the pixel being too deep by 1 pixel, e.g. `Vault`
Notice that Vault's tdm_spawns are one pixel deep in the ground.
**After this PR, tdm_spawn is placed on the ground correctly on both maps.**

### DIFF
--- a/Entities/Special/TDM/TDM_Ruins/TDM_Ruins.as
+++ b/Entities/Special/TDM/TDM_Ruins/TDM_Ruins.as
@@ -20,6 +20,12 @@ void onInit(CBlob@ this)
 	this.Tag("change class drop inventory");
 
 	this.getSprite().SetZ(-50.0f);   // push to background
+	
+	CShape@ shape = this.getShape();
+	if (shape !is null)
+	{
+		shape.PutOnGround();
+	}
 }
 
 void onTick(CBlob@ this)
--- a/Entities/Special/Tent/TentLogic.as
+++ b/Entities/Special/Tent/TentLogic.as
@@ -21,6 +21,12 @@ void onInit(CBlob@ this)
 
 	// defaultnobuild
 	this.set_Vec2f("nobuild extend", Vec2f(0.0f, 8.0f));
+	
+	CShape@ shape = this.getShape();
+	if (shape !is null)
+	{
+		shape.PutOnGround();
+	}
 }
 
 void onTick(CBlob@ this)


mugg91:
TDM_Ruins isn't spawning where it should on some maps. 
I will work on a fix.

mugg91:
I added a commit that made the following changes.

TDM_Ruins was misplaced on some maps. 
It spawned inside walls to the top on:
- SJD360_CrowbarsSwitchblades_v2
- Ferrezinhre_CursedWell_v2
- Biurza_Edge_Lord

And it spawned on a close-by ledge on:
- SzczurWodny 
- HomekBunnie_StrusBorzy

I removed `shape.PutOnGround()` and wrote my own logic in a new script `AlignToGround.as` which is added as a blob script to TDM_Ruins and Tent. 
When a TDM_Ruins or Tent is spawned, it will be placed exactly on the ground.

If there is no solid tile below or above, the TDM_Ruins or Tent will be removed. This doesn't happen unless you use `/spawn tent` above the void or inside a wall near the top of the map.

`AlignToTiles.as` is removed from Tent as it isn't needed.

It all should work as intended, now.

Entities/Common/Scripts/AlignToGround.as
```diff
@@ -0,0 +1,25 @@
+
+void onInit(CBlob@ this)
+{
```
Vam-Jam, line 3: I would set this script to server only
```suggestion
#define SERVER_ONLY

void onInit(CBlob@ this)
{
```

Entities/Common/Scripts/AlignToGround.as
```diff
@@ -0,0 +1,25 @@
...
+	CShape@ shape = this.getShape();
+	
+	// put on ground based on shape's position
+	CMap@ map = getMap();
+	uint ts = map.tilesize;
+	Vec2f pos = shape.getPosition();
+	pos = Vec2f(Maths::Floor(pos.x / ts) * ts, Maths::Floor(pos.y / ts) * ts); // align to tiles
+
+	while (!map.isTileSolid(pos))
+	{
```
Vam-Jam, line 13: My only suggestion is to add a failsafe to this loop, maybe stopping after trying to shift the tent up 10 times. Mappers may accidentally cause this to loop forever when making maps

Vam-Jam (changes requested):
Seems to work fine from testing, could not find any maps that were broken.

mugg91:
@Vam-Jam 
I applied your suggestions. It should probably be tested again just in case. 
It seems I can't resolve file conflicts but it should be trivial here.

Vam-Jam (approved):

Issue #1712, opened 2023-10-16
After a victory, the losing team's tent would explode


NikitaVNsv:
After a victory, the losing team's tent would explode. It would be desirable for the explosion to cause damage only to buildings (but this is not obligatory)

mugg91:
If anyone wants to implement this, here are some suggestions:
- After game ends, wait about 5 seconds (fanfare should have ended)
- Perhaps a brief shaking animation, similar to the cannon [here](https://www.youtube.com/watch?v=mLI2sZmJdk4&t=2940s) (must be on 720p60 to see it).
- Red flash, [sparke.ogg](https://github.com/transhumandesign/kag-base/blob/master/Entities/Items/Explosives/Sparkle.ogg)
- Explode like a keg would.
Players can see it going to explode so the explosion should kill nearby players.

PR #1965, opened 2024-03-20
[added] Loser team's Tent explodes

mugg91:
## Status

## Description
```
[added] Loser team's tent explodes
```
Fixes https://github.com/transhumandesign/kag-base/issues/1712

This PR changes Tent to make it explode  as soon as there is a winner team and if the winner team doesn't equal the tent's team num. In other words, the loser team's tent explodes.
This also happens after a successful surrender vote.

I added more tent sprites to give it a fuse and a red flashing frame, like bomb.
I changed a few lines here and there to make it all work without changing existing logic.

The thing is that bombs will run `SetupBomb()` in their `onInit()` whereas tent will run it in `onTick()` (at a later point).
A check `if (this.hasTag("exploding"))` in `onInit()` in `BombTimer.as` was added because otherwise `BombFuseOn()` would be executed for the tents as soon as the map loads in online. 

Some instances of `getNet().isServer()` changed to `isServer()`.

Since some sprite stuff is done in the tent .as file, I renamed it from `TentLogic.as` to `Tent.as`.

I made it so the "no build" area at the tent's location is removed as soon as "shes_gonna_blow" animation plays, so the ground can take damage and other stuff can be built and interacted with there after tent is gone. Other "no build" areas still remain active.

Tent's explosion is basicly a bomb explosion with higher damage and slightly larger radius. Anyone who is very close will die and gib, and anyone who is somewhat close will die without gibbing.

I encountered a problem that in ctf tutorial when the map reloads, the game would assume the tents to have an s32 `bomb_timer` value of 0 for some reason so `onTick(CSprite@ this)` fires and changes the tents sprite frames unwantedly.
So I had to set `this.set_s32("bomb_timer", -1);` in `onInit(CBlob@ this)` in `Tent.as` as a counter-measure.

Added support for an offset for where the fuse sparks should come out. 
`this.set_Vec2f("custom_sparks_offset", Vec2f(0,-24));` is used in Tent's `onInit(CBlob@ this)` which makes the sparks come out at the top.
If no offset is given the sparks will come out in the blob's center position as before.

I refrained from implementing some ideas I had here - https://github.com/transhumandesign/kag-base/issues/1712#issuecomment-1765439626

Tested in offline and online, no known remaining bugs. I suggest to test it with other players before adding it to a live build.

## Idea for improvement

Although not necessary for now it's of course possible to reduce on the sprite frames and make the fuse as a spritelayer. But I'm not sure if it can be easily done without messing with the existing logic too much.

## Steps to Test or Reproduce

Play CTF or CTF tutorial online or offline.
Collect the other team's flags or make it so a surrender vote gets passed.
Notice the loser team's tent will explode and deal damage.


mugg91:
Here is what it looks like in-game.
https://www.youtube.com/watch?v=7VL7QcwtGFw

mugg91:
I added a commit with the following improvements:
* `BombTimer.as` `onTick()` is only ticked if the blob has tag `"exploding"`. Therefore `UpdateBomb()` doesn't get called each tick when the bomb isn't even exploding yet.
* Bombs that are about to explode were supposed to have their sparks sound speed and volume increase. But the line `if (timer < fuseTicks / 2)` in `BombCommon.as` `UpdateBomb()` didn't work. By saving `fuseTicks` separately in `SetupBomb()`, it works correctly now.

mugg91:
I added a commit that makes it so Tent will explode with 16 8x8 gib particles of itself.
Previously it only made 10 wood gib particles which were set in `Tent.cfg`.

mugg91:
This PR still works. The only thing is that `Activatable.as` needs to be removed.

## Bomb Sparkle Volume and Pitch fixed

**Before**

 https://github.com/user-attachments/assets/80fcd9d7-1e32-44b2-97fd-ee06bd46107f  

**After**

https://github.com/user-attachments/assets/b215b408-f97e-4809-8756-4442d5bdacee  

## Tent Explosion

https://github.com/user-attachments/assets/0b6cd3a0-b3e9-44b0-a082-bcec6230df89

Issue #1968, opened 2024-03-21
Class change death bug


mugg91:
## Description

In TDM, I was archer and had 1 heart health remaining.
I died because I got killed by an arrow while changing from archer to knight.
I correctly became spectator and was rendered dead on the scoreboard, but knight did still spawn.
The knight got killed by another arrow.

mugg91:
The bug explained above seems to be fixed.
https://github.com/transhumandesign/kag-base/pull/1785 is probably the PR that fixed it.

There is now a different bug where the blob you switched from could die + make a death sound, but you can still continue on with the blob you switched to. I will close this as fixed.

I might open a separate issue for the new bug, but it is almost whatever.

PR #1970, opened 2024-03-22
[fixed] Change class death bug

mugg91:
## Status

## Description
```
[fixed] Change class death bug
```
Fixes https://github.com/transhumandesign/kag-base/issues/1968

When playing online, you could change class by clicking on a button at a respawn point while dying from taking damage.
In TDM, this caused a new blob to be created even though you were already rendered dead and removed from the match.

This PR adds a check to `StandardRespawnCommand.as` in `onRespawnCommand()` to see if the caller has tag `"dead"` and if he does, it doesn't proceed.

After applying this solution, the bug doesn't happen anymore.

## Steps to Test or Reproduce

Go to TDM, have the match start.
Be builder and place a ladder above your spawn.
Be low health.
Drop a spike from above you while standing on the ladder and quickly tap the button to change class.

If successful, you will be dead and won't be able to control any character. However, a knight or archer would still spawn and sit idle. 
In `on RespawnCommand()`, if you use `print("dead? "+caller.hasTag("dead"));`, the server-side console would print `dead? true`.

**After applying this PR, an idle knight or archer won't get spawned.**


Entities/Common/Respawning/StandardRespawnCommand.as
```diff
@@ -80,12 +80,12 @@ void onRespawnCommand(CBlob@ this, u8 cmd, CBitStream @params)
...
 		case SpawnCmd::changeClass:
 		{
-			if (getNet().isServer())
+			if (isServer())
 			{
 				// build menu for them
 				CBlob@ caller = getBlobByNetworkID(params.read_u16());
 
-				if (caller !is null && canChangeClass(this, caller))
+				if (caller !is null && !caller.hasTag("dead") && canChangeClass(this, caller))
```
Gingerbeard5773, line 88: It would be wiser to put this check into the canChangeClass bool- rather than in this line

Gingerbeard5773 (changes requested):

Entities/Common/Respawning/StandardRespawnCommand.as
mugg91: Resolved.

mugg91:
After testing, it looks like the issue was already fixed, probably by https://github.com/transhumandesign/kag-base/pull/1785
There is now a different bug of less significance where the blob you switched from could die and make a death sound, but you could still continue on with your new blob.

It seems there is no bug anymore where the class change is processed despite your death, probably thanks to the `if (caller is null) return;` check from the above PR.

PR #1973, opened 2024-03-24
[changed] Sandbox will allow faster building and mining

mugg91:
## Status

## Description

This PR makes changes to allow faster building and mining in Sandbox mode.
This makes sense to me because Sandbox mode should be all about building stuff.

Although `PlacementCommon.as` was changed to not check for rules tag `""faster building""` anymore, I added it back in and added the tag to `Sandbox_Rules.as`.

I noticed `BuilderLogic.as` would list animations `"chop_fast"` and `"strike_fast"` even though they aren't listed in `Builder.cfg`.
So I added them to `Builder.cfg`, reducing the animation time from 12 ticks to 8 ticks each.

If players would prefer the normal speed, maybe we can think about adding a setting to allow both.


Vam-Jam (approved):
Works in dedi server + localhost, does not seem to break CTF, WAR or tutorials.

PR #1974, opened 2024-03-24
[added] Water dripping effect for filled sponge and for runners who got out of water

mugg91:
## Status

## Description
```
[added] Water dripping effect
```

This PR adds a sprite script `WaterDrops.as` which is added to all runner classes.
It creates water particles and water drip sound if the blob has tag `"wet"`.

`RunnerDefault.as` will tag the runner `"wet"` if in water or if he has been in water in the past 5 seconds, otherwise untags it.

The effect only plays out if "faster graphics" is turned off.

The sounds have been extracted from here: https://pixabay.com/de/sound-effects/water-dripping-on-wateraif-14581/

Tested in offline and online, looks to be working well. The only thing I'm worrying is if the effect works correctly with many players in a server. Maybe checks for "is client", "is server", "is my player", etc. have to be added.

## Alternative solutions

As an alternative, instead of adding a sprite script to each .cfg file, it is probably possible to add it from within `RunnerDefault.as`.

## Steps to Test or Reproduce

Have "faster graphics" turned off.
Enter water and leave the water, notice water dripping off your character.

Throw a sponge in water and wait for it to be full.
Get the sponge out of water.
Notice the sponge will have water dripping off it.


mugg91:
Spotted an issue
* `WaterDrops.as` shouldn't be located in the Sponge directory. Should be in `Entities\Effects\Scripts`.

mugg91:
Added a commit with the following changes:
* `CLIENT_ONLY` added to `WaterDrops.as`, it works the same offline and online
* Don't make water drop particles when burning

Entities/Items/Sponge/Sponge.as
```diff
@@ -98,6 +98,20 @@ void onTick(CSprite@ this)
 {
 	u8 absorbed = this.getBlob().get_u8(ABSORBED_PROP);
 	spongeUpdateSprite(this, absorbed);
+	
+	// wetness
+	if (!v_fastrender)
+		{
```
Vam-Jam, line 104: indent here is wrong

mugg91:
I fixed the indent, but after copying over the files to test it on Build 4762, I found that the water dripping effect won't work for Sponge. Will look into it.

mugg91:
Originally the water dripping for Sponge lasted as long as it had sprite frame 3 (soaked), but since that sprite could last for a very long while, the water dripping lasting 15 seconds would look odd so I changed it. Now it works the same as for runners, only lasting about 5 seconds, and only if the sprite frame is 3 (soaked). No more changes planned now.



https://github.com/user-attachments/assets/0b56074f-586d-4346-a17e-a45b013aa4d1

Issue #1951, opened 2024-03-16
Stunned players still get to perform some actions related to buttons


mugg91:
## Description

These actions are not possible:

* Showing button at any tunnel (includes hall and outpost)
* Clicking on any buttons, including shop items, inventories and tent class menu

However, these actions are still possible while being stunned.

* Showing buttons for chest, present, class shops, quarters bed, lever, push button, coin slot, crate get in, storage storing all items, quarry put wood in
* Showing buttons for anything that has an inventory (storage, crate, hall, etc.)
* Showing buttons for anything that lets us change class (tdm_spawn, tent, hall, etc.)
* Hovering over any button and then letting go will execute the button's command (tunnel travel, change class at class shops or tdm_spawn)
* If holding `use` key, you can keep showing buttons even for tunnels.

Because of this, players can change class, open chests or use tunnel even when stunned.

## Suggestion

When getting stunned, make all currently shown buttons disappear.
Make it so you don't get to see any buttons (i.e. chest, present, etc. won't show buttons) while stunned.

The same logic that is used for tunnel travel button should be applied to all the other buttons.
Not sure where to look for the "mouse hovering over a button and then letting go to make it execute" logic, might be engine.
Not sure about inventory buttons which are engine-handled.

mugg91:
Stunned players can press 1, 2, 3, 4 ... in shop to buy items.

GELGELAT:
Suggestion don't touch it
why remove functions no one even complains about it but you want to remove

GELGELAT:
it's already part of the game high knowledge of the game and you just delete it

mugg91:
@GELGELAT 
I'm only pointing things out, I haven't created any change or fix yet.

So you are saying we should allow certain actions while stunned while not allowing specific things (like tunnel travel)?

We should allow buying items with number keys while not allowing buying items per mouse click? Why?

I'm sure these are things only a minority of players will know about anyway. You tigorsun are one of the best players, one of maybe 5 people who can decide games just by joining a team, doing a push, and then leaving afterwards (as I've seen you do the other day).

It's just like with bunnie who submitted PR to allow faster builder cooldowns. Why? Because bunnie gets to abuse the hell out of it. What about other players?

GELGELAT:
> [@GELGELAT](https://github.com/GELGELAT) I'm only pointing things out, I haven't created any change or fix yet.(Ok, not true, but it's not up to me if it gets into the game.)
> 
[...]

Issue #1028, opened 2021-04-14
Fix eating while stunned


benjamingorman:
**Description**

Pressing the food hotkey to eat while stunned does not work (as expected), however you can manually open your inventory, hover over a burger and close it in order to eat while stunned. This seems like a bug since the two control schemes behave differently.

**Reproduction**

1. Join any gamemode
2. Have somebody stun you with water (or any other type of stun)
3. Notice you cannot eat by pressing V while stunned
4. Notice you can eat by opening your inventory and selecting the food

**Expected Behavior**

I don't think access to inventory while stunned should be removed, since it would change other parts of the game, like being able to take your drill out while stunned to have it ready.

I propose to make it impossible to take food items out of your inventory while stunned.

Alternatively the item could go into your hands and the food logic could be changed so that food isn't consumed if you're stunned.

**Additional context**

Inability to eat while stunned would make water slightly stronger and there was some discussion in #development that we should perhaps accompany this change with a nerf to water, such as removing the ability to be chain-stunned by water arrows, or a flat price increase to water.

NotLaikaDawg:
It is completely an engine side issue, as your inventory does not get locked in any way while stunned, allowing you to cheese eating, using auto eat, by dropping them item so you consume it.
How some people dont think this is a glitch, or unintended, I may never know.

PR #1975, opened 2024-03-25
[fixed] You can't use buttons while stunned anymore

mugg91:
## Status

## Description
```
[fixed] that you could show "key_use" buttons while stunned
[fixed] that you could keep showing buttons, get stunned and then use them
```
Fixes https://github.com/transhumandesign/kag-base/issues/1951

### Preventing the player from showing new buttons

This PR adds a check to `canSeeButtons()` in `GenericButtonCommon.as` to see if the player is stunned (= knocked).
If the player is stunned, he won't be able to show any blob buttons by using "key_use".
Luckily this seems to work for inventory buttons, too. As far as I understand they are server-handled.
I wasn't able to access inventories after applying this fix.

### Removing already existing buttons

Inside `DoKnockedUpdate()` in `KnockedCommon.as`, I added:
```
if (this.isMyPlayer())
{
	this.ClearButtons();
	this.ClearMenus();
}
```
This causes buttons that already existed due to keeping "key_use" pressed to exit while the player is stunned.

Now the player can't manage inventories, change class or use tunnels while stunned anymore.

## Steps to Test or Reproduce

Be a knight, `/coins 999`, spawn a knightshop and buy a bunch of waterbombs.
Go to a chest, tent, storage or tunnel (when there are at least 2).
Throw a waterbomb above you.
Try to press or keep pressed the `key_use` key in order to show or keep showing buttons for above mentioned blobs, before or after getting hit by the waterbomb.
Notice you will be able to show and use buttons while stunned.
**After this PR, you won't be able to.**

### DIFF
--- a/Entities/Common/Scripts/GenericButtonCommon.as
+++ b/Entities/Common/Scripts/GenericButtonCommon.as
@@ -1,6 +1,12 @@
+
+#include "KnockedCommon.as";
+
 bool canSeeButtons(CBlob@ this, CBlob@ caller)
 {
-	if ((this is null || caller is null)) { return false; }
+	if (this is null || caller is null || isKnocked(caller))
+	{ 
+		return false; 
+	}
 
 	CInventory@ inv = this.getInventory();
 	return (
--- a/Entities/Characters/Scripts/KnockedCommon.as
+++ b/Entities/Characters/Scripts/KnockedCommon.as
@@ -70,7 +70,7 @@ u8 getKnockedRemaining(CBlob@ this)
 bool isKnocked(CBlob@ this)
 {
 	if (this.getPlayer() !is null && this.getPlayer().freeze)
-	{	
+	{
 		return true;
 	}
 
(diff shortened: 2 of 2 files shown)


mugg91:
I noticed this old issue https://github.com/transhumandesign/kag-base/issues/1028

This PR already fixes the issue but the inventory screen would open for 1 tick with each key_inventory press while stunned.
I added a commit that will make it so it doesn't open at all during stun.

Fixes the above-mentioned issue.

Gingerbeard5773 (approved):
Tested and works as intended.

Entities/Common/Scripts/GenericButtonCommon.as
```diff
@@ -1,6 +1,12 @@
+
+#include "KnockedCommon.as";
```
Vam-Jam, line 2: Push the include up one line

Vam-Jam (approved):
Works fine, will leave this open for discussion for now as it does close #1028, which some people may see as a feature that should not be removed (?)

Entities/Common/Scripts/GenericButtonCommon.as
mugg91: Fixed.

Issue #1925, opened 2024-02-25
Some blobs don't void out properly


mugg91:
## Description

* When chest falls into void it breaks
* When bomb falls into void it explodes
* When arrow collides with void, it is stuck on the void
Probably more things exist.

Suggestion is that these things should not happen and the blob should just fall into the black and then vanish there.

mugg91:
Any good way of knowing we are voiding out?
Or does the y coordinate have to be checked against the map height?

mugg91:
arrows stuck in the void is probably related to this https://github.com/transhumandesign/kag-base/issues/1940

PR #1976, opened 2024-03-25
[fixed] Blobs will void out properly

mugg91:
## Status

## Description
```
[fixed] Keg and Bomb would explode when touching the void
[fixed] Crate and chest would break when touching the void
[fixed] Water arrow, waterbomb and filled bucket would splash when touching the void
[added] sparkle fadeout sound for keg, bomb and water bomb
```
Fixes https://github.com/transhumandesign/kag-base/issues/1925

This PR does changes to some blobs to make them void out properly.

I think the void should be like a very deep "nothing", so blobs should not randomly break apart or explode when touching the void.

It doesn't seem like there is any conventional way of telling if a blob has voided out. The condition `this.getPosition().y > map.tilemapheight * map.tilesize` seems to work well so I added a new script `Void.as` with the function `isVoidedOut()` with that condition. The script was then added to some blobs to prevent unwanted stuff from happening when voided.

Since bomb's, keg's and waterbomb's `Sparkle.ogg` and `WaterSparkle.gg` "emit sound effect" would abruptly end when voiding, I added sound effects `"SparkleFade.ogg` and `WaterSparkleFade.ogg` which are played by those blobs via `sprite.PlaySound()` when voiding out to give the impression the blobs are actually falling down into the "nothing".

Tested in offline and online, no problems.

Although I know there is an issue with arrows touching and getting stuck on the void, I didn't encounter it in testing in offline and online. So I couldn't test if this PR fixes that.

## Steps to Test or Reproduce

Throw exploding waterbomb, bomb, keg or shoot waterarrow or throw waterbucket, crate or chest into a void.
Notice they will explode, splash or break apart.
**After this PR, they won't**


TerminalHash:
> Looks like some people in Discord didn't like this change. But nobody has made any arguments as to why. I'm just trying to improve the game here. Rather than posting thumb-downs or saying something sucks, I think it would help more if people give constructive criticism and stay polite. Thank you.

You're not fixing a bug, you're destroying a fairly useful feature. Maybe it's not realistic that kegs and bombs explode for a vanishing point in the void, maybe it's not have logic, but it solves some problems, like in TDM with rat-archers. There's a slight difference between fixing annoying bugs (which may constitute a skeleton to foul play) and changing quite a few useful features.

Vam-Jam:
Going to mark as discussion for now, the code seems fine but I wont test until people have decided on if this is something the game should receive or not

mehwaffle10:
I agree, this is a nerf. Especially in TDM, you can use it to bomb jump or kill camping archers. I wouldn't change this, it's niche enough that it doesn't impact anyone except hurting those who like to use it

mugg91:
Prior to making this PR, I didn't know bombs exploding in the void was a feature or a thing that people actively use. I never really see it in TDM, and if I do see it, it usually fails and the archer survives anyway. I just found it odd that crates break open or kegs/bombs explode when entering the void.

mugg91:
Unpopular PR. Also I'm sure adding `onVoidOut()` hook to the engine would make things easier, in case we do want Crates and Chests to not break apart.

Issue #1980, opened 2024-03-27
crate in TDM


GELGELAT:
**Current in-game behaviour.**
The tdm store has one empty cell.

**Proposed Change**
It would be cool if they added crate to that slot.

**Additional context**
It would diversify the game, you could do fun things and traps.

mugg91:
It's not trivial to do because crate requires 2 width and 1 height spaces. I'll try to change the logic to make it require only 1 by 1.

PR #1981, opened 2024-03-27
[added] Crate in TDM shop (without team color)

mugg91:
## Status

- **READY**

## Description

[added] Crate to TDM shop
~[changed] Keg in TDM shop has team color~
[fixed] Wrong default cost for burger in TDM shop

Fixes https://github.com/transhumandesign/kag-base/issues/1980

This PR adds Crate to TDM as a purchasable item.
Since the regular crate icon is 32x16 pixels and therefore would take 2 width and 1 height slots, I created a new icon `$crate_tdm` which is 16x16 pixels and therefore only requires 1 by 1 slots.

This icon is added in `MakeTradeMenu()` in `TDM_Trading.as`.
I wasn't sure if running `AddIconToken()` every round is bad so I added a check that the icon is only added if `(getIconTokenFilename("$keg_tdm$").length == 0)` is true, so it will only be added once and not every round. 

I fixed that `cost_burger` had a wrong default value. Now it has the same cost as in the cfg file.

I tried to give keg and crate team colored icons, but I didn't find a way, at least not without giving crate a 2 by 1 space.
For now it looks like this:

screen-24-03-29-13-44-11


GELGELAT:
what price did you give the box?

mugg91:
@GELGELAT 
50 coins. Should I change it something else?

GELGELAT:
I think its a good price

mugg91:
I settled for "no team colors" now.
I didn't find a way to make the shop buttons have team colored icons, at least not without giving crate a 2 by 1 space.
Updated the screenshot above.

mugg91:
There is a PR that makes items in TDM shop have team color, which I think is superior so I'm closing this.

Issue #1983, opened 2024-03-29
Make it possible to place trap block in Sandbox


mugg91:
## Current in-game behaviour.

Give builder trap block to place it Sandbox.
Either on the first page (normal tiles) or on a new page (misc tiles).

PR #1986, opened 2024-03-29
[added] Trap Block can be placed in Sandbox mode

mugg91:
## Status

## Description
```
[added] Trap block to Sandbox builder blocks
```
Fixes https://github.com/transhumandesign/kag-base/issues/1983

This PR changes `CommonBuilderBlocks.as` to make trap blocks placable once again, but only in Sandbox.
There was one free slot available on the first page of builder's block inventory, so it fits perfectly.
This might raise interest in someday adding it back to other gamemodes.


Vam-Jam (approved):
Works fine from testing (aside from trap block being kinda weird with its collision code)

PR #1987, opened 2024-03-31
[added] Crate in TDM shop (with trader's team color)

mugg91:
## Status

## Description

This is the same implementation as https://github.com/transhumandesign/kag-base/pull/1981 with the following changes:
* Removed the check for `GUI::hasIconName(...)` and applying `AddIconToken()` directly, with trader's team num.
`TDM_Ruins.as` does this, too when building the class switch menu, which would always be shown in the caller's team num.
* If you go to the blue shop and then to the red shop, then back to the blue shop, the blue shop's icons would have been overwritten with the red icons. So I'm using `"$icon_name" + trader.getTeamNum() + "$"` as icon names.

Now the tdm shop will always show items in the shop's team color. Of course items will still be given in the caller's team num.

screen-24-03-31-16-23-03

screen-24-03-31-16-23-01

screen-24-03-31-16-22-58


mugg91:
@Vam-Jam 
I added a commit that should fix the issue and icons should now show up correctly and the same both in offline and online.
I moved `AddIconToken()` instances from `TDM_Trading.as` to `TradingPost.as`.

Issue #1716, opened 2023-10-17
Please, add a fork for wires to the game


NikitaVNsv:
Please, add a fork for wires to the game
    B
A-+-A
    B
If is A is switch on and B is swith off, A is has signal, B hasn't signal
If is A is switch off and B is swith on, A is hasn't signal, B has signal
If is A is switch on and B is swith on, A is has signal, B has signal
If is A is switch off and B is swith off, A is hasn't signal, B hasn't signal

mugg91:
I'm rewriting MechanismsCommon.as to hopefully make it work.
Might also add a dual elbow.

mugg91:
Sorry, I had no luck finishing it. Kept having too many errors.
When changing functions or variables inside the `MapPowerGrid` class, it said "Shared type 'MapPowerGrid' doesn't match the original declaration in other module".
If someone helps me deal with this, I can continue working on the fork tile.

mugg91:
Basicly I would replace 
```
	array<u8> power_old;
	array<u8> power_new;
	array<u8> topo_in;      
	array<u8> topo_out;     
```
with
```
	array<array<IO_Element>> io_element_array;
```
and add
```
shared class IO_Element
{
	u8 input_topology;
	u8 output_topology;
	u8 power_new;
	u8 power_old;
	
	IO_Element(u8 input, u8 output, u8 power)
	{
		input_topology = input;
		output_topology = output;
		power_new = power;
		power_old = power;
	}
};
```

This would allow multiple input/output topologies and new/old powers in one component, since I would like to make fork work like a wire and not like a special tile.

asumagic:
> Sorry, I had no luck finishing it. Kept having too many errors. When changing functions or variables inside the `MapPowerChunk` class, it said "Shared type 'MapPowerChunk' doesn't match the original declaration in other module". If someone helps me deal with this, I can continue working on the fork tile.

I assume you've tried it, but restarting the game may be necessary when editing shared classes, are `rebuild` is going to have a hard time with it.

mugg91:
> I assume you've tried it, but restarting the game may be necessary when editing shared classes, are `rebuild` is going to have a hard time with it.

Nah, I didn't try. It seems that error stopped showing but now the game just doesn't load the map anymore and crashes.
I guess I will try without objects and will just use plain values to get the basics working.

PR #2000, opened 2024-04-08
[added] Dual Elbow & Fork wire components

mugg91:
## Status

- **IN DEVELOPMENT**
- May be possible to polish code-wise.
- Debug render (`g_debug = 1` in console) should visualise how much power is in the 2nd set and where it is going.

## Description
```
[added] Fork wire component
[added] Dual Elbow wire component
```
Fixes https://github.com/transhumandesign/kag-base/issues/1716

This PR changes logic in `MechanismsCommon.as` and all component script files to support a 2nd set of `input topology`, `output topology`, `power_new` and `power_old`. They have the same names but with a `2` appended to it at the end. The first set is refered to as "section 0" and the second set is refered to as "section 1".

The two new tiles make use of the 2nd set, whereas all the other components don't, making `Activate()`, `Deactivate()` and `Special()` return early from the function if it detects that it is called with any set that isn't "section 0".

Since the builder inventory menu couldn't fit both new component tiles on the 2nd page, I made it so it will be enlarged by an additional row when playing in Sandbox.

Tested offline, works as intended.

Although it is technically ready, it is probably possible to improve the code.
Initially, I tried implementing a class object that contains a whole set (see above), with the `MapPowerGrid@` receiving an array of such objects via `grid.setAll()`. Unfortunately the game kept crashing during loading each time so I resorted to implementing a second set of primitive variables instead.


mugg91:
In-game look - Yes, the Dual Elbow sprite is apparently off by 1pixel.
screen-24-04-09-00-08-34

Vam-Jam:
Setting as a draft as its been tagged as `IN DEVELOPMENT`

mugg91:
I'm wondering if this worth putting in the game. Also the implementation is more a "proof of concept". I'm not sure if it's a good thing to use "section 0" and "section 1".

Issue #1996, opened 2024-04-06
Cant put grain of tree


GELGELAT:
**Description**
You cant put grain of tree in when there's a saw or a store there.

**Reproduction**

https://github.com/transhumandesign/kag-base/assets/87004057/f0800d12-b103-40e5-b029-56af7a67078e

https://github.com/transhumandesign/kag-base/assets/87004057/f0d0147d-8556-44ee-8b15-8c4780440ab1

if we want to plant a seed when theres a shop or a saw there, we cant do it. 

I wish it could be done. Because its very inconvenient

mugg91:
It is worth noting that, by default, seeds do grow on top of Saw, but not on top of Building.

GELGELAT:
> It is worth noting that, by default, seeds do grow on top of Saw, but not on top of Building.

allow you to put the grain where the saw is?

mugg91:
I hate `PlacementCommon.as` but I will look into it.

PR #2001, opened 2024-04-08
[changed] Seed can be placed on non-static blobs

mugg91:
## Status

- But please look for odd cases.

## Description
```
[changed] Seed can now be placed on non-static blobs
[fixed] Seed can no longer be placed on another seed
```
Fixes https://github.com/transhumandesign/kag-base/issues/1996

Seed can now be placed on any blob unless it is static.
Therefore you can place seed on a Saw, Dinghy or other blobs that impose a no build area but not on a Building.
Also made it so you cannot place a Seed on another Seed.

Placing seed on a stuck arrow still works.

Briefly verified in offline.


Gingerbeard5773 (approved):
Tested and works as intended. Decent QOL for tree farms

Issue #1999, opened 2024-04-08
Icons turn grey in Builder's Block Inventory


mugg91:
## Description

When selecting any tile on the first page, everything is normal, but on the other three pages, icons turn grey after selecting them.
It seems this persists until you restart the game.

PR #2002, opened 2024-04-09
[fixed] Text wasn't aligned properly in Builder's block inventory

mugg91:
## Status

## Description

Single line text in builder's block inventory had an unnecessary linebreak.

## Before

before

## After
after

### DIFF
--- a/Entities/Characters/Builder/BuilderInventory.as
+++ b/Entities/Characters/Builder/BuilderInventory.as
@@ -110,7 +110,8 @@ void MakeBlocksMenu(CInventory@ this, const Vec2f &in INVENTORY_CE)
 			BuildBlock@ b = blocks[PAGE][i];
 			if (b is null) continue;
 			string block_desc = getTranslatedString(b.description);
-			CGridButton@ button = menu.AddButton(b.icon, "\n" + block_desc, Builder::make_block + i);
+			block_desc = block_desc.find("\n") != -1 ? "\n" + block_desc : block_desc;
+			CGridButton@ button = menu.AddButton(b.icon, block_desc, Builder::make_block + i);
 			if (button is null) continue;
 
 			button.selectOneOnClick = true;


mugg91:
Current build in-game look
screen-24-04-11-13-27-21
* Shows title and description mixed together with `\n` in two different places
* Typos and missing punctuation

----
I'm working on an improvement:
screen-24-04-11-13-23-31
* Title and description should be two separate strings and should be visibly separate on the hover text.
* Don't show description in bottom hover text.
* No `\n` inside descriptions.
* Start sentences with a capital letter and end them with a full stop.
* Working on adding components descriptions.
* Locale files didn't have `trap bridge` yet, adding that

mugg91:
Added a commit that makes the following changes (which is in line with the announced changes above):
* `BuilderInventory.as` changed to support more than 12 blocks per page while setting the minimum number of rows per page to 4. Changed some numbers to make it align with backpack inventory correctly. This change will come in handy when merging with any PRs that introduce more blocks.
*  `BuildBlock.as` and `BuilderInventory.as` changed to support a block title separate from its description.  Hover text for blocks will work as described in my previous post. I opted against changing the separation line for now.
* Descriptions to components added.
* Locale files changed to account for the changes and to add Trap Bridge entry.

### Todo:
* These Locale files still need to be done: Japanese, Polish, Portugese, Russian, Spanish, Turkish

mugg91:
Added a commit that fixes https://github.com/transhumandesign/kag-base/issues/1999 .

mugg91:
I think this would look best, but I don't think there is markup code within a string to make text get rendered bigger.

(Mockup: )
e

GoldenGuy:
Would have been great if you could figure out a way to make item icon height centered like this:
image

mugg91:
@GoldenGuy 
Yes, I would like to fix that but it's not trivial. 
These are inventory icons of the material blobs which `Requirements.as` ready directly by blob name.
Changing inventory icons would be a bad idea since itaalso shifts the icon up in other places we don't want (such as the inventory grid or the bottom hud).

`Requirements.as` would need to be changed to support different icons.

Or changes to the string markup could be done to allow setting the image in the location we want, for example:
`$icon_name$[align-top]`
`$icon_name$[align-center]`
And to allow changing font size. :p

Vam-Jam:
Setting as a draft for now as it was marked as `IN DEVELOPMENT`, fix up the merge requests/remaining issues and ill review it.

mugg91:
> Setting as a draft for now as it was marked as `IN DEVELOPMENT`, fix up the merge requests/remaining issues and ill review it.

Thanks, I might sooner or later.

Issue #2004, opened 2024-04-10
Components and devices have wrong frame when rejoining server


mugg91:
## Description

When building fresh

![screen-24-04-10-16-51-04](https://github.com/transhumandesign/kag-base/assets/14877281/8c0e9b27-696c-4dd6-b91b-f2436ad5f610)

After quitting and rejoining:

![screen-24-04-10-16-51-28](https://github.com/transhumandesign/kag-base/assets/14877281/50552560-ae0b-4a4d-bb9e-3ffca16038b6)

* Lamp has light, but no light color
* Pressure Plate has wrong frame (not pushed down)
* Lever has wrong frame (not active)
* Spiker has an additional spike on top?
* Most wire components have the wrong frame (not active)

I guess `onReceiveCreateData()` could be used to set the necessary state on rejoining.

PR #2007, opened 2024-04-11
[fixed] Wrong sprite frame on components and devices after rejoining server

mugg91:
## Status

## Description

[fixed] Some components and devices had the wrong frame after rejoining server
[dev] canBePickedUp() removed from component scripts and added to GenericComponent.as
[dev] this.Tag("builder always hit") removed from component scripts and added to GenericComponent.as
[dev] getNet().isServer() changed to isServer()
[dev] getNet().isClient() changed to isClient()
[changed] Spike and GenericSpike merged, Spike won't have a blood spritelayer but instead changes its frame (similar to https://github.com/transhumandesign/kag-base/pull/1854 )

Fixes https://github.com/transhumandesign/kag-base/issues/2004

This PR changes code in components to make their sprite frame update on rejoining the server.

Some code moved from component scripts to `GenericComponent.as` as a clean-up measure.  If there should be any component in the future that needs to be picked up or not be hit by builder, it can all be changed, but for now I think it is better to not repeat `canBePickedUp()` and `this.Tag("builder always hit")` a quadrillion times.
However, I didn't seek a solution for the repeated `onReceiveCreateData()` and `UpdateSprite()` instances because I didn't want to add a new script and add it to all the component blobs.
Anyway, this PR has more deletions than additions. :)

Please note, this PR - similarly to https://github.com/transhumandesign/kag-base/pull/1854 - merges `Spike.as` and `GenericSpike.as`. The linked PR implements that spike from spiker properly remembers its bloody state but at the time of this writing, it still has an issue that spike will be displayed on top of the spiker frame on rejoining. This PR fixes that.

Tested in online and offline, no problems encountered.

## Steps to Test or Reproduce

Join server.
Place a bunch of components and devices such as lamp, toggle, lever, spiker and pressure plate.
Have them in different states (active and inactive), then quit and rejoin server.
Notice they will have the wrong sprite frame (usually inactive, when it should have the active sprite etc.).
**After this PR, all components and devices will have the correct sprite frame after rejoining.**


Vam-Jam:
Merge conflicts just need to get sorted

PR #2008, opened 2024-04-12
[added] Two new components: Bell & Display

mugg91:
## Status

## Description
```
[added] Load component Bell
[added] Passive component Display
```
This PR adds two new components.
They will probably not see uses in TDM, but will possibly be enjoyed by Sandbox players.
Feel free to suggest changes.

## Bell
Will ring when powered.
Free sound taken from here: https://pixabay.com/sound-effects/ac-bel-105874/

Sprite ticking performance is good.
I checked with `for (int i = 0; i < 10000; i++)` loop around the code that sets offset and rotation of bell spritelayer and ringer spritelayer. With one Bell it is barely noticeable while with two, there is quite a bit of lag. 
This result is satisfactory, because it means it takes 10000 Bells before there is lag just from the ticking. In reality lag starts happening when there are like 300~400 blobs just existing.

bell

## Display
Will work like a junction that displays the current power number.
While it is possible to use `g_debug = 1` in console to display power numbers, I would like an in-game way of showing it.

It can display from `0` up to `16`, higher values will display `Er`, standing for "Error". In actual game-play, only `9` can be reached since power sources are coded to start with power `10`.

screen-24-04-12-10-38-55


mugg91:
@NoahTheLegend 
Thanks for your advice. Although I think the color contrast is ok. I'm afraid making it fully black will look odd although it would be possible to give it a dark grey outline.

That being said, I added a commit that makes improvements to how Bell looks and animates.

Vam-Jam (changes requested):
It seems to work from my testing (I can not test the audio since linux).

When placing the bell, the blob appears to be semi broken.



I'm unsure about the screen's colour, it does look out of place.

mugg91:
@Vam-Jam 
Do you mean the sprite for Bell looks odd when placing it?
I didn't change since since Lever also uses half its sprite when placing it.

Vam-Jam:
> @Vam-Jam Do you mean the sprite for Bell looks odd when placing it? I didn't change it back then since Lever also uses half its sprite when placing it.

Yea, if that's the case I'll revisit it this week.

Vam-Jam:
Closing as this is more suited for a sandbox mod

Issue #1528, opened 2022-12-02
Queue puts you as waiting to join "any team" when waiting to join a certain team


mugg91:
## Description

1) As spectator, click "join red team" or "join blue team".
2) Notice that sometimes you will get put as waiting to join "any team".

## Reproduction

I don't know reproduction steps yet.

mugg91:
It seems the problem is only graphical. 

1) It's 4 blue players, 6 red players, I queue as "waiting to join red team"
2) The button turns grey.
3) It still correctly joins me into red team when a slot becomes available.

AsuMagic:
@bunniewormy ping, I missed that issue.

mugg91:
In `SpectatorQueue.as`:
```
void onPlayerChangedTeam(CRules@ this, CPlayer@ player, u8 oldteam, u8 newteam)
{
	if (newteam == this.getSpectatorTeamNum())
	{
		SetupQueueGUI(this);
	}

	if (oldteam != this.getSpectatorTeamNum() && player.isMyPlayer()) 
	{
		hide = true;
		return;
	}

	if (this.get_bool(player.getUsername() + "_playsound"))
	{
		if (isClient() && player.isMyPlayer())
		{
			Sound::Play("AchievementUnlocked.ogg"); // TODO: different, distinct sound
		}

		this.set_bool(player.getUsername() + "_playsound", false);
	}

	client_selected = -99;
	RemoveFromQueue(player);
}
```
I think the line `client_selected = -99;` is the cause of the problem, making your button grey when someone before you switched teams. Maybe the issue is fixed by moving the line into the `if (oldteam != this.getSpectatorTeamNum() && player.isMyPlayer())` code block. But I'm currently unable to test it.

PR #2010, opened 2024-04-13
[fixed] Queue button doesn't turn grey when other player changes team

mugg91:
## Status

## Description
```
[fixed] Queue button doesn't turn grey when other player changes team
```
Fixes https://github.com/transhumandesign/kag-base/issues/1528

This PR changes code in `SpectatorQueue.as` to fix a bug where the queue button in the top right corner turns grey whenever a different player changes teams. The culprit is the line `client_selected = -99;` in `onPlayerChangedTeam()`.

Since the `newteam == this.getSpectatorTeamNum()` check and the `oldteam != this.getSpectatorTeamNum() && player.isMyPlayer()` check are basicly the same (player changes from red/blue team to spectator), I mixed both code blocks and also moved the line `client_selected = -99;` inside it. `SetupQueueGUI(this);` also should only run if `player.isMyPlayer()` is true.

I tested in a dedicated CTF server with bots. Although the queue logic doesn't work well with bots, when there are enough bots it is possible to queue for one team without instantly entering it, allowing me to test the bug scenario. When queuing for a team and having a bot player change its team, my button previously turned grey but after applying the above fix it doesn't.

The sound for successfully getting entered into a team will still play at the correct time.
It plays only when switching from spectator to red/blue via the queue.

|                      | Current build | After this PR |
| ------------- | ------------- | ------------- |
| Switch from spectator to red/blue via queue  | Sound plays  | Sound plays  |
| Switch from spectator to red/blue via team button while queuing  | Sound doesn't play | Sound doesn't play |
| Switch from red/blue to red/blue   | Sound doesn't play  | Sound doesn't play |
| Switch from red/blue to spectator  | Sound doesn't play  | Sound doesn't play  |

### DIFF
--- a/Rules/CommonScripts/SpectatorQueue.as
+++ b/Rules/CommonScripts/SpectatorQueue.as
@@ -438,14 +438,11 @@ void onPlayerLeave(CRules@ this, CPlayer@ player)
 
 void onPlayerChangedTeam(CRules@ this, CPlayer@ player, u8 oldteam, u8 newteam)
 {
-	if (newteam == this.getSpectatorTeamNum())
+	if (player.isMyPlayer() && newteam == this.getSpectatorTeamNum())
 	{
-		SetupQueueGUI(this);
-	}
-
-	if (oldteam != this.getSpectatorTeamNum() && player.isMyPlayer()) 
-	{
+		client_selected = -99;
 		hide = true;
+		SetupQueueGUI(this);
 		return;
 	}
 


mugg91:
I tested the issue and couldn't reproduce it anymore. It may have been fixed in https://github.com/transhumandesign/kag-base/pull/1805

mugg91:
I tested again and the issue doesn't happen after https://github.com/transhumandesign/kag-base/pull/1805 so I'm closing this.

PR #2014, opened 2024-04-15
[added] new source component: Solar Panel

mugg91:
## Status

- **IN DEVELOPMENT**
- Needs to be added to png loader and loader colors.
- Doesn't work on server when tested in "all mugg PRs merged", I will look for a fix

## Description
```
[added] Solar Panel
```

This PR adds Solar Panel as a new Source component.

It will provide power based on luminance of SColor fetched via `getMap().getColorLight(this.getPosition())`. 

Light must be above a set threshold otherwise power is 0 and tick frequency is 45. The threshold is set to 140.
If light is above the threshold, then power is `(light - LIGHT_THRESHOLD) / (250.0f - LIGHT_THRESHOLD) * power_source` and tick frequency is 3.
Therefore, Solar Panel provides power based on the light amount. If there is less light then it will provide less power.

Blob emitted light will work, so a Solar Panel can be powered by a Lamp.
However, there is an issue where blob emitted light will go through walls.
This is partly the reason why I introduced a threshold (see above).

I might change Lamp to give less light if it has less input power.

## Screenshots
screen-24-04-15-10-56-39
screen-24-04-15-10-56-57
screen-24-04-15-11-00-08
screen-24-04-15-10-59-35


Vam-Jam (changes requested):
Does not function correctly on a dedicated server, seems to just output at max power all the time regardless of the brightness

mugg91:
Hmm, it also doesn't work in offline. Not sure what went wrong, since it did work at one point.
I'm not submitting PRs unless I'm convinced they are complete.

I will take a look.

Vam-Jam:
Closing as this is better suited for mods

mugg91:
With the game featuring day & night, this makes sense to me to put in the game. Sad about the decision, but as explained, it is not possible to implement it anyway due to `getLuminance()` always returning 255 on server.

PR #2016, opened 2024-04-15
[changed] Lamp light is based on input power

mugg91:
## Status

## Description
```
[changed] Lamp light is based on input power
```
This PR changes Lamp to make it emit light based on its input power.
Less power means it emits less light.
Lamp isn't a "load component" anymore, but it doesn't make sense for it to be in the "passive components" directory so I kept it in the "load components" directory.

A power factor `(float(power_new + 1) / power_source)` is calculated which is set to not exceed `1`. 
Light radius will be directly affected by this power factor whereas SColor luminance, Red, Blue and Green are affected by it as well but with the factor having at least a minimum of `0.3f` to make light not drop too low.

screen-24-04-16-00-30-55

### DIFF
--- a/Entities/Structures/Components/Load/Lamp/Lamp.as
+++ b/Entities/Structures/Components/Load/Lamp/Lamp.as
@@ -2,6 +2,8 @@
 
 #include "MechanismsCommon.as";
 
+const u8 MAX_LIGHT_RADIUS = 128.0f;
+
 class Lamp : Component
 {
 	u16 id;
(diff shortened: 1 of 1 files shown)


NoahTheLegend:
You're adding more mechanisms and thats cool, but we should add arithmetic gates with a complex custom logic.

In short you may need one parent device for storing a value once it is different from its contents (previously saved value). That is implementation for average memory cell, then using that as a layout add divide/multiply and sum/subtraction gates which interact with nearby memory cells (so you do not have go out max signal range) taking value from side A, processing output to side B.
Making those more versatile to work with environment is preferable. Activating coin slot or sensor would add 1 / subtract 1 or multiply by 2 / divide by 2  etc corresponding to functionality of certain device 

this is quite raw idea and i was thinking of doing it by myself, although i will appreciate your feedback and more thoughts relative to this

Vam-Jam (changes requested):
Seems to break lighting on the lamp

Vanilla:
https://fs.vamist.dev/8f2742-5h5KOgJJruB0qVEJ.png

This PR:
https://fs.vamist.dev/8e8569-qOgPexmvrYxKElmS.png

Vam-Jam:
I suspect `SetLightRadius` is causing it to not work

Gingerbeard5773:
I personally don't agree with this change. I don't see any value that this would bring and instead would make the lamp more painful to use

mugg91:
It looks like this PR behaves correctly only in offline. In online it needs to be fixed.

> I personally don't agree with this change. I don't see any value that this would bring and instead would make the lamp more painful to use

This change aims to add more depth to the game. More depth makes the game feel more alive. 
It's just an idea, this PR is rather unimportant in comparison to some others.
I would like to try to fix the online issue before devs can decide what to do with it.

mugg91:
I submitted changes, the issue with the broken light in online should be fixed now. 
Client command is sent inside the Lamp's `Special()` now.
Code looks ugly lol, but it works.
Now you can do whatever.

Issue #2024, opened 2024-04-17
Archer can charge while sleeping


mugg91:
## Current

Sleeping in a quarters won't get rid of the charging.
If in "can legolas shoot" state, the bow spritelayer will be on top of the bed while sleeping.
After getting out of the bed, the charging continues.

## Suggestion

Sleeping should cancel the charging completely.

## Additional

Knight's attacks already get canceled, so Archer's should, too.

PR #2025, opened 2024-04-17
[fixed] Archer could continue stabbing or charging Arrow after getting attached

mugg91:
## Status

## Description
```
[fixed] Archer continued to charge Arrow when sleeping in a quarters
[fixed] Archer's stabbing state could continue after sleeping in a quarters
[fixed] Archer's cursor frame didn't reset when Archer got attached
```

Fixes https://github.com/transhumandesign/kag-base/issues/2024

This PR makes fixes to `ArcherLogic.as` so Archer won't be able to charge Arrows "in their sleep".
It was also possible to press `Action_2` to start stabbing a log, quickly start sleeping, and then Archer would fletch an arrow even when not pressing any keys.

When getting attached while charging, the cursor frame kept the charging meter. After this PR it will reset to indicate there is no charging.

Tested in offline and online.

### DIFF
--- a/Entities/Characters/Archer/ArcherLogic.as
+++ b/Entities/Characters/Archer/ArcherLogic.as
@@ -83,7 +83,8 @@ void ManageGrapple(CBlob@ this, ArcherInfo@ archer)
 		&& !archer.grappling
 		&& this.isOnGround()
 		&& !this.isKeyPressed(key_action1)
-		&& !this.wasKeyPressed(key_action1))
+		&& !this.wasKeyPressed(key_action1)
+		&& !this.isAttached())
 	{
 		Vec2f aimpos = this.getAimPos();
 		CBlob@[] blobs;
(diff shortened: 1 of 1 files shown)


Entities/Characters/Knight/KnightLogic.as
```diff
@@ -1613,25 +1568,8 @@ void onHitBlob(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@
...
-
-	u8 bomb_id;
-	if (!params.saferead_u8(bomb_id)) return;
-
-	string matname = bombTypeNames[bomb_id];
-	blob.set_u8("bomb type", bomb_id);
-
-	blob.SendCommand(blob.getCommandID("pick " + matname));
-}
-
```
Vam-Jam, line 1633: Could you explain why this was changed? I do not see it mentioned in the PR

Vam-Jam (changes requested):
I'd like a little bit more clarification on what the change to callback change does, as its not mentioned in the PR. Could you fix the merge conflict?

The mentioned features work fine, may need more testing

Entities/Characters/Knight/KnightLogic.as
mugg91: It's because I made this PR before all those bunnie changes were submitted and merged, as is the case with many of my other PRs. It would take more time and dedication to sort these file conflicts out which I'm not really interested in doing for now.

Vam-Jam:
Setting as a draft for now.

mugg91:
Fixed the file conflict and tested it briefly, works. Let me know if there are more issues.

Issue #1979, opened 2024-03-26
Dispenser and Spiker don't work when platform is placed at their exit


mugg91:
## Description

I think it is odd that Dispenser doesn't spit out an item when an away-facing platform is on it.
Maybe Dispenser and Spiker should ignore platforms.

PR #2026, opened 2024-04-20
[fixed] Spiker and Dispenser will work with away-facing Platforms at their exit

mugg91:
## Status

- **IN DEVELOPMENT**
- Spiker will not work correctly as it shoots spike through any solid blobs.

## Description
```
[fixed] Spiker will work when Platform is at its exit but only if it is not facing towards Spiker
[fixed] Dispenser will work when Platform is at its exit but only if it is not facing towards Dispenser
[changed] If Dispenser can't pop out items, don't make a Particle
[added] Dispenser will have a different sound effect if its exit was blocked
```
Fixes https://github.com/transhumandesign/kag-base/issues/1979

Previously, Spiker and Dispenser would not work if a Platform was placed at their exit. Since it looked odd that they wouldn't work even with away-facing platforms, I made this PR to make changes.

The `CollidesWithPlatform()` code from https://github.com/transhumandesign/kag-base/pull/1864 is used, this time separated in a new file `PlatformCommon.as` and added as an include in `Spiker.as` and `Dispenser.as`.
To make the ray correctly detect the platform, start position `position + offset * map.tilesize/2;` is used instead of `position + offset * 5;`

Previously, Dispenser would make a particle and sound without popping out item if its exit is blocked.
I changed it to not make the particle to better indicate that it is blocked.

Tested in offline. Should work the same in online.

screen-24-04-20-13-05-12
screen-24-04-20-13-31-16

Blocked by bridge
screen-24-04-20-13-07-59

but it works if it is opened (since it is non-solid)
screen-24-04-20-13-07-36


Vam-Jam:
This PR was marked as `IN DEVELOPMENT`, is it ready for me to test/are there any known issues?

Vam-Jam (approved):
Code looks sane and can confirm it works in dedi+client setup.

My one suggestion is you can place a platform on a spike that is active, either we should prevent platforms being placed or have the spike retract. This could for for another PR however.

mugg91:
This didn't work in the "all in one" PR I made but since @Vam-Jam said they confirmed it works, all is good.

Issue #1985, opened 2024-03-29
Change Quarters beer to an orange juice, when using kidssafe setting


mugg91:
## Current in-game behaviour.

Make quarters beer icon change into an orange juice icon or vice versa when toggling kidssafe on or off.

GELGELAT:
lol, yes

PR #2029, opened 2024-04-20
[added] Quarters will switch between Beer and orange juice when toggling kidssafe setting

mugg91:
## Status

- This PR does not take into account other PRs merged since build 4541.
- This PR needs testing.

## Description
```
[added] Quarters will switch between Beer and orange juice when toggling kidssafe setting
```
Fixes https://github.com/transhumandesign/kag-base/issues/1985

This PR changes `ShopCommon.as` to add `replaceShopItem()` which is run in `Quarters.as` only for client, replacing beer with juice or vice versa based on the player's `g_kidssafe` setting.

## Please note
I'm not sure if this will work for each player separately.

Please test:
* Have two human players in a server and build a quarters
* Human player 1 uses `g_kidssafe = false` and player 2 uses `g_kidssafe = false`
* Player 1 changes to `g_kidssafe = true`, quarters has juice.
* Does quarters have juice for player 2 as well? (If yes, then this PR can be burned)

There is possibly an alternative solution that might work better, but it takes time investigating. Quarters could add two shop_arrays, one with beer, one with juice and one or the other is shown to the player based on their `g_kidssafe` setting.
But that would also require non-trivial changes to `ShopCommon.as` and I'm not sure it will work (if the game can support two different shops in one blob, basicly).

Please note that this PR does not incorporate other PRs merged in since the current live build 4541.
I'm focussing on getting things to work based on the current live build first and foremost.

## Screenshot

screen-24-04-20-23-33-25


mugg91:
I think this PR is not good. Directly editing ShopItem is probably a bad idea.
Instead, if possible, two different shops should be set up with `g_kidssafe == true` presenting the user one shop button and `g_kidssafe == false` presenting the other shop button.

Issue #2032, opened 2024-04-22
Add /suicide chat command


mugg91:
## Description

Might be convenient if you are typing in chat and don't want to be bothered opening the menu and clicking with your mouse.

PR #2033, opened 2024-04-22
[added] Suicide chat command

mugg91:
## Status

## Description
```
[added] suicide chat command
```
Fixes https://github.com/transhumandesign/kag-base/issues/2032

This PR adds a suicide chat command.
The command works when typing in chat `/suicide` or `/die` and it kills you just like if you press the suicide button in menu.
I set it to not require `admin only` or `sv_test only`, so it can be used in any gamemode.

## Steps to Test or Reproduce

Type in `/suicide` or `/die`, notice your character dies just like if you press the "Suicide" button in the menu.

### DIFF
--- a/Rules/CommonScripts/ChatCommands/PlayerCommands.as
+++ b/Rules/CommonScripts/ChatCommands/PlayerCommands.as
@@ -1,4 +1,6 @@
-#include "ChatCommand.as"
+
+#include "ChatCommand.as";
+#include "Hitters.as";
 
 class ClassCommand : ChatCommand
 {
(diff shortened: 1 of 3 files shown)


Gingerbeard5773 (approved):
Tested and works as intended

Vam-Jam (approved):

Rules/CommonScripts/ChatCommands/PlayerCommands.as
```diff
@@ -207,3 +209,27 @@ class HealCommand : ChatCommand
...
+		if (isServer())
+		{
+			CBlob@ blob = player.getBlob();
+			if (blob is null)
+			{
+				server_AddToChat(getTranslatedString("You cannot kill yourself while dead or spectating"), ConsoleColour::ERROR, player);
+				return;
+			}
+
+			blob.server_Hit(blob, blob.getPosition(), blob.getVelocity(), 10.0f, Hitters::suicide);
```
Vam-Jam, line 232: On second thought, would it not be better to use blob.server_SetTimeToDie, at least then death will always happen

mugg91: Doesn't SetTimeToDie despawn the blob altogether? We just want to kill it on the battle field.

mugg91: It looks like for SetTimeToDie to work, you have to set a positive value above 0.
However, we don't want to kill the blob in the future. We want to kill it now.
Are there cases where server_Hit used like this will not result in the player's death?

Gingerbeard5773: only in mods where players have crazy amounts of health. You could probably just set the damage output to be double that of the blob's initial health or something. 10.0f should be plenty though so its whatever

Issue #2037, opened 2024-04-25
Bug in Seats.as


mugg91:
## Description

`Seats.as` has:
```
void onAttach(CBlob@ this, CBlob@ attached, AttachmentPoint @attachedPoint)
{
	if (attachedPoint.socket)
	{
		attached.Tag("seated");
		Sound::Play("GetInVehicle.ogg", attached.getPosition());

		if (this.getDamageOwnerPlayer() is null) {
			this.SetDamageOwnerPlayer(attached.getPlayer());
		}
	}
}

void onDetach(CBlob@ this, CBlob@ detached, AttachmentPoint@ attachedPoint)
{
	if (attachedPoint.socket)
	{
		detached.Untag("seated");

		if (!detached.getShape().isRotationsAllowed())
		{
			detached.setAngleDegrees(0.0f);
		}

		if (detached.getPlayer() is this.getDamageOwnerPlayer()) {
			this.SetDamageOwnerPlayer(null);
		}
	}
}
```

This code sets the sitting player as the damage owner and unsets him when he stops sitting.

However, when the sitting player dies, the code in `onDetach()` doesn't run and the player will still be the damage owner even when not sitting anymore.

When getting killed while sitting on a Bison, all kills the Bison does will be counted towards that player. Someone else sitting on the Bison will not override the damage owner due to the check for `if (this.getDamageOwnerPlayer() is null)`.

Also, when the damage owner player gets killed by the Bison, the game will say that the player killed himself.

I you do this bug with a catapult, - although not tested - it may be possible that someone else riding over enemies will count towards the player that previously sat in the catapult. I did verify that a catapult I'm no longer sitting in will attribute kills to me at least.

----

In a match, I tamed a Bison and got killed while riding.
After respawning, I saw kills getting attributed to me which is the reason I found this.
![222](https://github.com/transhumandesign/kag-base/assets/14877281/4c829833-67b7-4997-8944-1607c82ddd7e)

mugg91:
It looks like `onDetach()` does get run and successfully unsets the damage player.
But then I'm not unterstanding why the bug is happening ...

PR #2038, opened 2024-04-25
[fixed] Bison's or vehicle's damage owner did not get unset when sitting player died

mugg91:
## Status

## Description
```
[fixed] Bison's or vehicle's damage owner did not get unset when sitting player died
```
Fixes https://github.com/transhumandesign/kag-base/issues/2037

When playing online and dying while sitting on a catapult or a bison, the game correctly runs `onDetach()` in `Seats.as`.

However, 
```
		print("detached is null ? " + (detached is null));
		print("detached player is null ? " + (detached.getPlayer() is null));
		print("detached tag dead ? " + (detached.hasTag("dead")));
```
will yield false, false, false on client and false, true, true on server.
Because the player was already null on server, the damage player didn't get unset.
To fix the issue, this PR adds a check if the detached blob has the tag `"dead"`. When this condition is true we know the player must have died and we need to unset the damage owner.

Tested in online, works.

## Steps to Test or Reproduce

Play CTF with bots in a server.
Spawn bison and tame it.
Spawn a bomb and place it in your inventory, then sit on the bison to die while sitting.
Notice how any kills the bison gets now will be attributed to you.
**After this PR, kills are not attributed to you anymore.**

### DIFF
--- a/Entities/Vehicles/Common/Seats.as
+++ b/Entities/Vehicles/Common/Seats.as
@@ -67,7 +67,8 @@ void onDetach(CBlob@ this, CBlob@ detached, AttachmentPoint@ attachedPoint)
 			detached.setAngleDegrees(0.0f);
 		}
 
-		if (detached.getPlayer() is this.getDamageOwnerPlayer()) {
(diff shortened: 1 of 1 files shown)


Vam-Jam (approved):
LGTM, works in testing

Issue #1822, opened 2024-01-08
Bomb's onHitBlob (and likely other hooks) does not run sometimes on client


bunniewormy:
**Description**

In current vanilla, there's an issue with bomb jumps.
Bomb jumps sometimes feel "weaker" for no particular reason.
When a bomb explodes, it dies on server, then does server_Hit().

This means that the following hooks should run:

onHit in ShieldHit.as (knight was hit by bomb, blocked attack; add some velocity)
onHitBlob in Explosion.as (bomb has hit knight; add some velocity)

However, in certain conditions, onHitBlob - and most likely some other hooks - will NOT run on the client.

**Reproduction**

We add 2 prints to Explosion.as and ShieldHit.as:

ShieldHit.as
```
f32 onHit(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@ hitterBlob, u8 customData)
{
	[...]

	if (blockAttack(this, velocity, 0.0f) && this.hasTag("shielded"))
	{
		if (isExplosionHitter(customData)) //bomb jump
		{
			printf("onHit in ShieldHit.as " + getGameTime());
```

Explosion.as
```
void onHitBlob(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@ hitBlob, u8 customData)
{
	if (customData == Hitters::bomb || customData == Hitters::water)
	{
		{
			if (hitBlob.hasTag("player"))
			{
				printf("onhitblob " + hitBlob.getName() + ", velocity: " + velocity + " , length: " + velocity.Length() + ", gt: " + getGameTime());
			}
			hitBlob.AddForce(velocity);
		}
	}
}
```

Now, you need to keep bomb jumping on a dedicated server until the bomb jump will feel weaker for no reason.
To make it easier to reproduce, you should use some kind of lagswitch (managed to reproduce on Captains server with clumsy 50ms inbound lag and 5% drop)

The condition in which it happens is as follows:

For a good bomb jump, prints are like this (RCON is server-side prints):
```
[16:18:35] <RCON> ONHIT SHIELDHIT VELOCITY: (0, -1.8), gt: 25719
[16:18:35] <RCON> onhitblob knight, velocity: (79.36028, -208.3307) , length: 222.934, gt: 25719
[16:18:35] ONHIT SHIELDHIT VELOCITY: (0, -1.8), gt: 25719
[...]

PR #2062, opened 2024-05-04
[modified & fixed] better bombs (bomb explosion force)

bunniewormy:
## Status

## Description

1) bomb knockback force is now independent of distance between bomb and player. tested on captains, feels significantly better than vanilla bomb jumps, less weird ping-based timing which can make your bomb jump feel worse 
2) fixed the issue that bomb jumps from right-to-left were on average weaker than from left-to-right due to Maths::Round being dumb 
obraz
3) fixed https://github.com/transhumandesign/kag-base/issues/1822 by adding BombHit script and applying velocity from bomb explosions there instead of Explosion.as onHitBlob

### DIFF
--- a/Entities/Characters/Archer/Archer.cfg
+++ b/Entities/Characters/Archer/Archer.cfg
@@ -178,6 +178,7 @@ $name                                             = archer
 													FleshHitEffects.as;
 													PainSounds.as;
 													KnockBack.as;
+													BombHit.as;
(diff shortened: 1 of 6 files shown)


mugg91:
That's interesting, so the reason I had trouble bomb jumping from right to left wasn't just me being bad.

asumagic (approved):

PR #2063, opened 2024-05-04
[fixed] fix stun, when you jumping on trampoline

TerminalHash:
## Status

## Description

In the standard situation, when trying to jump on a trampoline for a while, the player gets a stun:
https://github.com/transhumandesign/kag-base/assets/64691722/5728a98d-23bc-4a13-99f9-62967c078928

With the fix, the player will no longer get stun from jumping on the trampoline (including from a lower height)
https://github.com/transhumandesign/kag-base/assets/64691722/22156bef-395f-4181-9d91-003d8e8e8a47

## Steps to Test or Reproduce
1. Build a small pillar of blocks upwards
2. Put the trampoline underneath
3. Climb the pole and jump from it onto the trampoline
4. ...
5. STUN!

### DIFF
--- a/Entities/Common/Movement/FallDamage.as
+++ b/Entities/Common/Movement/FallDamage.as
@@ -35,12 +35,12 @@ void onCollision(CBlob@ this, CBlob@ blob, bool solid, Vec2f normal, Vec2f point
 
 		// better check for trampolines
 		CBlob@[] blobs_around;
-		if (getMap().getBlobsInRadius(this.getPosition(), this.getRadius() * 4, blobs_around)) {
+		if (getMap().getBlobsInRadius(this.getPosition(), this.getRadius() * 4, blobs_around))
+		{
 			for (uint i = 0; i < blobs_around.length; i++)
 			{
 				CBlob@ b = blobs_around[i];
 
-				if (b is null) continue;
 				if (!b.hasTag("no falldamage")) continue;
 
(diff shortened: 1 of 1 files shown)


Entities/Common/Movement/FallDamage.as
```diff
@@ -33,24 +33,29 @@ void onCollision(CBlob@ this, CBlob@ blob, bool solid, Vec2f normal, Vec2f point
...
-					CBlob@ b = overlapping[i];
-
-					if (b.hasTag("no falldamage"))
-					{
-						return;
-					}
-				}
+				CBlob@ b = blobs_around[i];
+
+				if (b is null) continue;
```
Vam-Jam, line 43: You wont need a null check here, all the blobs returned by getBlobsInRadius are valid (and you are not storing it over multiple ticks)

Vam-Jam (changes requested):
Format the if statement to match, otherwise the code works fine and not found any bugs

TerminalHash:
done

Vam-Jam (dismissed):
lgtm

Vam-Jam (approved):

asumagic (approved):
This is very imperfect because it will cancel fall damage even in cases where it shouldn't (most notably when holding crouch while falling down), but I will merge this as is despite this. IMO it's less of a big deal than trampolines being basically broken half the time.

PR #2104, opened 2024-07-12
Victory by capturing one of enemy flags when the timer is activated

TerminalHash:
## Status

## Description
In our captains mod, Gruhsha CTF, we using vanilla match timer, i was surprised there was no code written for when one of the enemy's flags was captured and the timer activated.  I don't know if anyone else would play with the timer on, but why not?

Future plans are to write overtime conditions for when one of the flags is captured and match time has expired, but at another time.

## Steps to Test or Reproduce
1. Enable match timer in CTF.as by editing this line (editing ctf_vars.cfg is not working for this):
`s32 gameDurationMinutes = cfg.read_s32("game_time", -1);`
2. Cap one flag from enemy team
3. Wait, when timer will be ended

### DIFF
--- a/Rules/CTF/Scripts/CTF.as
+++ b/Rules/CTF/Scripts/CTF.as
@@ -40,8 +40,10 @@ void Config(CTFCore@ this)
 	{
 		this.gameDuration = (getTicksASecond() * 60 * gameDurationMinutes);
 	}
+
 	//how many players have to be in for the game to start
 	this.minimum_players_in_team = cfg.read_s32("minimum_players_in_team", 2);
+
 	//whether to scramble each game or not
 	this.scramble_teams = cfg.read_bool("scramble_teams", true);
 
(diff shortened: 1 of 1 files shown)


Vam-Jam (changes requested):
Feature works, lgtm, just fix the comment indent 

Anybody else wanting to test can do so by running `/rcon getRules().set_u32('game_end_timer', 30 * 60 * 1); getRules().set_bool('no timer', false);`

Rules/CTF/Scripts/CTF.as
```diff
@@ -600,6 +612,20 @@ shared class CTFCore : RulesCore
...
+			if (team_num == 0 && flags_red.length < flags_blue.length)				// BLUE WIN
+			{
+				team_wins_on_end = 0;
+			}
+			else if (team_num == 1 && flags_blue.length < flags_red.length) 		// RED WIN
+			{
+				team_wins_on_end = 1;
+			}
+			else if (flags_blue.length == flags_red.length)							// TIE
+			{
```
Vam-Jam, line 625: These comments line up wrong too

TerminalHash: yop, fixed this

Vam-Jam (approved):

TerminalHash:
Added overtime implementation, which we have on Gruhsha CTF.
For the vanilla game, you need to find a sound or several sounds that will be played when the time is extended, not to mention that it would be beneficial to discuss the significance of both the time added and the moment at which it should be added.

There is also an amusing feature: if both flags are selected, the time is added twice and the sound is played twice. This has not yet been corrected (due to lack of experience, I don't know how to do it better), but I decided to fix only the sound so that it is played exactly once.

GELGELAT:
Vanilla doesn't need artificial round limitation, on  pear it was needed to increase the number of rounds and to have a feeling that everyone is doing well

GELGELAT:
and adding one minute doesn't mean anything at all, day before yesterday there was a three-hour game. They stole 1 flag from us during the hour, then another one after 30 minutes and the last flag remained (there were 3 flags) and we defended for an hour and a half and stole their 3 flags in the last 10 minutes and won

GELGELAT:
so this timer will just kill the real competitive spirit and CTF mode will not be the same anymore

PR #2113, opened 2024-09-02
[fix] Vans_Ziraeth gold tiles adjusted to stop crashing

Vam-Jam:
Can not reproduce on linux (spamming `/rcon /loadmap Vans_Ziraeth`), will test windows at another point (or somebody else can verify it works).


Vam-Jam:
map fix credit: SonantDread

Vam-Jam:
Closing, engine patch should fix the issue instead

PR #2114, opened 2024-09-02
[fix] Tilepos on blob hits being set to 0,0 instead of an actual tilepos

### LINKED ISSUE #2082: Particles are spawned in the wrong location (Build 4762)
## Description

In the latest build 4762, particles are spawned in the top left corner.

For example, when hitting a wood log, the particles will not appear at the log, but at the top left corner of the map.
This did not happen in the previous build.
Happens both offline and online.

## 

Note; I will not work on more PRs until more of my existing PRs are merged into the game.

Vam-Jam:
Unsure what is actually going in the pickaxe code. Hitdata.tilepos only contains data when actually hitting a tile, but tilepos seems to contain the correct position when hitting a blob.

 Another fix could be editing line 610 and use b.getPosition() as suggested by Gingerbeard. 

Will do more testing later.

Credits to @Gingerbeard5773 

Closes #2082

### DIFF
--- a/Entities/Characters/Builder/BuilderLogic.as
+++ b/Entities/Characters/Builder/BuilderLogic.as
@@ -367,7 +367,7 @@ void Pickaxe(CBlob@ this)
 		{
 			CBitStream params;
 			params.write_u16(hitdata.blobID);
-			params.write_Vec2f(hitdata.tilepos);
+			params.write_Vec2f(tilepos);
 			this.SendCommand(this.getCommandID("pickaxe"), params);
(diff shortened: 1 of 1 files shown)


TerminalHash:
still appears on cave maps

asumagic:
> still appears on cave maps, tested on localhost and server

I feel like this is a separate issue of some sort? I'm not really seeing such issues

asumagic (approved):
Works for me (or at least is less broken)

PR #2132, opened 2024-09-27
[new] Show players in the browser

### LINKED ISSUE #143: Feature Pitch: Display players in each server in the server browser
The data for this is already stored in the API so someone only needs to add it. It could look similar to how I did it in my chrome extension:
image

Vam-Jam:
Closes #143 

Requires engine pr

### DIFF
--- a/GUI/guiBrowser.xml
+++ b/GUI/guiBrowser.xml
@@ -1202,7 +1202,7 @@
 				<attributes>
 					<int name="Id" value="10041" />
 					<string name="Caption" value="&quot;current players&quot;" />
-					<rect name="Rect" value="10, 380, 320, 220" />
+					<rect name="Rect" value="10, 350, 320, 450" />
 					<position name="MinSize" value="1, 1" />
 					<position name="MaxSize" value="0, 0" />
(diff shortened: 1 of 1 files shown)


GUI/guiBrowser.xml
```diff
@@ -1171,7 +1171,38 @@
...
+				</attributes>
+
+			</element>
+
+			<element type="staticText">
+
+				<attributes>
+					<int name="Id" value="10041" />
+					<string name="Caption" value="&quot;current players&quot;" />
+					<rect name="Rect" value="10, 380, 320, 220" />
```
asumagic, line 1205: Need to look into it but this makes it not render on staging. Swapping 220 and 380 makes it show "correctly", although alignment is weird (didn't have time to check vanilla right now to see if it was just like this already). Might be better to align to the bottom of the window or something if that space is unused?

Also letting this define the alignment instead of hardcoding it in engine.

Vam-Jam:
Fixed:

Vanilla: 

Staging:

asumagic (approved):

PR #2134, opened 2024-10-08
[new] Start of voting to kick logging

Vam-Jam:
Add's voting to kick logging that gets sent to bsion.

What the message looks like:

Known issues:
- Admin forcing to cancel does not call Pass() whatsoever, but forcing to pass does


Vam-Jam:
Added a new engine bind `getNet().sv_current_ip`

Vam-Jam:
PR also moves report code to the new engine bind, instead of having the client send `joined_ip` every report.

PR #1910, opened 2024-02-14
[changed] Fireplace and Egg will have collision with platforms and doors

mugg91:
## Status

- **READY**: this PR is (to the best of your knowledge) ready to be incorporated into the game.

## Description
```
[changed] Fireplace and Egg will have collision with static blobs (platform or door)
```
Fixes https://github.com/transhumandesign/kag-base/issues/1892

This PR does changes to the collision checks of Fireplace and Egg.
In particular, Fireplace and Egg can now stand on platforms and doors and will not fall through.

Tested in offline, works.

PR #1173, opened 2021-11-24
Add chicken emote; Allow players to sit on eggs to incubate them

benjamingorman:
## Status

- **READY**: this PR is (to the best of your knowledge) ready to be incorporated into the game.

## Description

image

This PR adds a new chicken emote (using one of the existing chicken sprite frames). It also adds the ability for players to crouch on eggs to incubate them. An incubated egg will hatch 3x faster than a regular egg. Whilst crouching on the egg you will automatically do the chicken emote. Your emote will be cleared if you leave.

## Steps to Test or Reproduce

I've tested this locally in sandbox.

- Open sandbox and spawn an egg
- Move over it as knight/builder/archer and crouch
- You should start incubating the egg. Your character will do the chicken emote, and the egg should hatch 3x faster than an un-incubated egg.

PR #2141, opened 2024-10-20
Chicken Breeding Rework

mehwaffle10:
Chicken breeding changes stemming from discussion on https://github.com/transhumandesign/kag-base/pull/1910. These have been tested on US captains for the past few weeks.

## Status

## Description

- Chickens no longer use a global variable for sound or egg laying, which led to chickens being unable to breed or make sounds if any chicken was making noise
- Moved shared variables to `ChickenCommon.as`
- Chickens now always breed even if there isn't a player nearby
- Chickens and eggs now ignore other chickens/eggs that are held or in inventories when counting nearby chickens/eggs
- Eggs now prevent hatched chickens from clipping into walls
- Eggs will not hatch a chicken inside of a static blob
- Eggs will not hatch a chicken in a 1-wide gap
- Eggs now reset their hatch time when detached
- Eggs no longer tick when held
- Eggs now always hatch even if they can't spawn a chicken
- Reduced egg hatch time to 20 seconds
- Reduced chicken/egg search radius to 10 blocks
- Chickens will not lay eggs if there are 3 other chickens/eggs nearby
- Eggs will not spawn a chicken when hatching if there are 4 chickens nearby

## Steps to Test or Reproduce

Chickens no longer use a global variable for sound or egg laying, which led to chickens being unable to breed or make sounds if any chicken was making noise

-Set up a chicken coop on one side of the map. Collide repeatedly with a different chicken ~20 blocks away (to make sure there is no overlap). Chickens in coop can still make `Pluck` sound and will continue to lay eggs.

Chickens now always breed even if there isn't a player nearby

- Set up a chicken coop on one side of the map. Walk to other side of map. Chickens will still breed.

Chickens and eggs now ignore other chickens/eggs that are held or in inventories when counting nearby chickens/eggs

- Place 3 chickens on the ground. Have a chicken/egg in your hand/inventory. Chickens will stay lay an egg. Continue with chicken/egg in hand/inventory. Egg will still hatch

Eggs now prevent hatched chickens from clipping into walls

- Place an egg directly against the side of a block or static blob. The chicken will not clip into the wall when hatched
image

Eggs will not hatch a chicken inside of a static blob

- Place an egg in a door or platform. It will not spawn a chicken when it hatches

Eggs will not hatch a chicken in a 1-wide gap

- Place an egg in a 1-wide gap. It will not spawn a chicken when it hatches
image

Eggs now reset their hatch time when detached

- Place an egg on the ground. Pick it up after 10 seconds. It will take ~20 seconds to hatch once dropped (delays occasionally due to tick timings. Egg only ticks every 4 seconds)

Eggs no longer tick when held

- Hold an egg. It will never hatch

Eggs now always hatch even if they can't spawn a chicken

- Place an egg in a 1-wide gap or in a coop with at least 4 chickens. It will not spawn a chicken when it hatches

Reduced egg hatch time to 20 seconds

- Place an egg on the ground. It will hatch in ~20 seconds (repeat of delay note above)

Reduced chicken/egg search radius to 10 blocks

- Create 2 chickens breeders at least 10 blocks apart, 1 that has 10 blocks between the chickens on the top and bottom and one that has 5. The former will continue to drop eggs while the latter will not once there are 2 chickens in the bottom (assuming you only put 2 chickens in the top)

Chickens will not lay eggs if there are 3 other chickens/eggs nearby

- Place 4 chickens in a coop. They will not lay eggs
- Place 2 chickens in a coop. They will lay 2 eggs total and no more

Eggs will not spawn a chicken when hatching if there are 4 chickens nearby

- Place 4 chickens and an egg in a coop. Wait until the egg hatches. There will still only be 4 chickens in the coop

If you added new features and want to explain how to use them, do so here.

Chicken coop:
- Will generate up to 4 chickens, but you need to keep 2 to continue production
image

Simple chicken breeder set up:
- 2 chickens in the top (optional 3rd for slightly faster production)
- Can take all 4 chickens from the collection chamber
image

- git branch used: chicken-breeding-rework
- gamemode if relevant: N/A
- specific objects or behaviours affected: chickens, eggs

Follow-Up Recommended Changes:
- Prevent `RegrowPlants.as` from spawning chickens, flowers, and grain. Random full healing or chickens isn't great for competitive gameplay
- Lower chicken health. Chicken shields against archers are stronger than they should be, considering 90% of the time it's an accident. Increased chicken health was also likely a band-aid fix to the next problem
- Change chicken team on pick up and prevent dealing damage to friendly chickens. Prevents teammates from accidentally killing all of your chickens in the coop or when you are holding a chicken. I'd recommend some sort of indicator for team color, I have a rudimentary one that's supposed to be a leg tag. This is also an issue for a lot of team based objects (bombs, boulders, etc)
image
- Potentially rework chickens further. Chickens being more accessible in their current state will exacerbate the trampoline sky tower meta that dominants EU CTF currently. In my mod, I've made chickens have a strong horizontal movement decay to nerf chicken and trampoline combo, and chickens will die if you take any damage while holding them so you can be shot down. On the flip side, I've made their hover stronger and added a double jump mechanic for them to give them more uses. I've also made chickens directly purchasable from the quarters for 50 coins instead of eggs. This incentivizes actually creating a breeder while also reducing the wait time (buying an egg and waiting 50 seconds is a horribly slow payoff when half the time a teammate kills the chicken)

### DIFF
--- a/Entities/Natural/Animals/Chicken/Chicken.as
+++ b/Entities/Natural/Animals/Chicken/Chicken.as
@@ -2,21 +2,18 @@
 //script for a chicken
 
 #include "AnimalConsts.as";
+#include "ChickenCommon.as";
 
 const u8 DEFAULT_PERSONALITY = SCARED_BIT;
-const int MAX_EGGS = 2; //maximum symultaneous eggs
-const int MAX_CHICKENS = 6;
-const f32 CHICKEN_LIMIT_RADIUS = 120.0f;
-
-int g_lastSoundPlayedTime = 0;
-int g_layEggInterval = 0;
+const string ALLOW_SOUND_TIME = "last sound time";
+const u8 SOUND_DELAY = getTicksASecond();
+const string EGG_INTERVAL = "egg interval";
 
 //sprite
 
 void onInit(CSprite@ this)
 {
 	this.ReloadSprites(0, 0); //always blue
-
 }
 
 void onTick(CSprite@ this)
@@ -82,24 +79,19 @@ void onInit(CBlob@ this)
 	this.set_f32(terr_rad_property, 75.0f);
 	this.set_u8(target_lose_random, 14);
 
-	//for shape
+	// for shape
 	this.getShape().SetRotationsAllowed(false);
 
-	//for flesh hit
+	// for flesh hit
 	this.set_f32("gib health", -0.0f);
 	this.Tag("flesh");
 
 	this.getShape().SetOffset(Vec2f(0, 6));
 
-	this.getCurrentScript().runFlags |= Script::tick_blob_in_proximity;
-	this.getCurrentScript().runProximityTag = "player";
-	this.getCurrentScript().runProximityRadius = 320.0f;
+	// breeding/sound
 
-	// attachment
-
-	//todo: some tag-based keys to take interference (doesn't work on net atm)
-	/*AttachmentPoint@ att = this.getAttachments().getAttachmentPointByName("PICKUP");
-	att.SetKeysToTake(key_action1);*/
+	this.set_u32(ALLOW_SOUND_TIME, 0);
+	this.set_u8(EGG_INTERVAL, 0);
 
 	// movement
 
@@ -142,6 +134,7 @@ void onTick(CBlob@ this)
 		}
 	}
 
+	u32 allow_sound_time = this.get_u32(ALLOW_SOUND_TIME);
 	if (this.isAttached())
 	{
 		AttachmentPoint@ att = this.getAttachmentPoint(0);   //only have one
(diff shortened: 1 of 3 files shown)


Vam-Jam (approved):
Code looks fine and works fine from testing. Will mark as discussion for others to chime in.

My only suggestion is, its not clear without reading the PR as to what will allow an egg to hatch. It would make sense to have eggs rotate slightly (back and forth)  when they are in a spot where the egg would make a chicken, so users have some sort of feedback as to whats good vs bad.

Vam-Jam:
Another idea, but you could also take a look at including https://github.com/transhumandesign/kag-base/pull/1173

mehwaffle10:
I like the idea, it helps with the issue of wanting a chicken immediately while still buying eggs from the quarters instead of chickens directly (meaning no need to redo localizations). Can also make it rotate faster if incubated, only concerns are making sure it's responsive if the tick rate is limited and that eggs will have to do the space check more than once

PR #2145, opened 2024-10-24
[fixed] pickup crash

### LINKED ISSUE #1342: Items shouldn't be picked through platforms

### LINKED ISSUE #2087: Items getting repeatedly picked up and thrown across the map
## Description
I see this sort of exploit done by other players on a regular basis in the current build 4762.
https://www.youtube.com/watch?v=VO8dEdzZHZM

(I was under the impression that @bunniewormy 's sanitizations would fix this but it didn't.)

Gingerbeard5773:
## Status

- **READY**

## Description

 * Fixes a client side crash that can occur when players use the pickup-menu.
It also fixes "random" crashes that people have.

* Fixes an issue where players could throw blobs that weren't attached to their player.

Cause of the crash:
Line 13 StandardPickup.as
```
CBlob@[] blobs;
this.set("pickup blobs", blobs);
CBlob@[] closestblobs;
this.set("closest blobs", closestblobs);
```

**This PR also changes how pickup deals with platforms.**
Players cannot pickup through platforms that are facing against them, but can pickup through platforms if they are facing with them. (this addition to this PR means that #1864 and #1342 can be closed)

resolves #1864, resolves #1342, resolves #2087

### DIFF
--- a/Entities/Common/Controls/StandardControls.as
+++ b/Entities/Common/Controls/StandardControls.as
@@ -70,6 +70,7 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
 
 		if (pickedup.isAttached()) return;
 
+		pickedup.setPosition(caller.getPosition());
 		caller.server_Pickup(pickedup);
 	}
 	else if (cmd == this.getCommandID("detach"))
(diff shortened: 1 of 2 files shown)


Gingerbeard5773:
After testing the changes on XDoodles Zombies server, I can confidently say that this PR will fix the 'random' crashes that occur on staging.

Vam-Jam (approved):

Issue #2162, opened 2024-11-16
Own arrows shouldn't get hit by own Saw when shooting upwards


mugg91:
## Current in-game behaviour.

When shooting upwards through your own Saw, your arrows shouldn't get destroyed.

Issue #2161, opened 2024-11-16
Saw should not destroy Trampoline of the same team


mugg91:
## Current in-game behaviour.

It would be cool and open up new strategies if Saw was allowed to bounce on Trampoline without destroying it. Also it makes sense that same team things shouldn't destroy each other.

PR #2163, opened 2024-11-17
[changed] Trampoline interaction with Saw & Arrows

mugg91:
## Status

## Description

```
[changed] Saw isn't destroyed by Trampoline unless if it falls into it from above
[changed] Own team arrows will not get eaten by Saw except if they are shot at it from above
```

Fixes https://github.com/transhumandesign/kag-base/issues/2161
Fixes https://github.com/transhumandesign/kag-base/issues/2162

This PR aims to open up more strategies by allowing same team Arrows to be shot through Saw from the side or below, and by allowing Saw to bounce on Trampoline without destroying the Trampoline. Enemy team arrows will still collide with and deal damage to Saw if shot from any side.

Unlike suggested in the issues linked above, I think it is ok to have Saw not destroy Trampoline regardless of the team.

Tested in offline and online, works as envisioned.

This PR is based on the current Build 4762.

## Comparison

**Before**

**After**

### DIFF
--- a/Entities/Industry/Saw/Saw.as
+++ b/Entities/Industry/Saw/Saw.as
@@ -188,8 +188,8 @@ bool canSaw(CBlob@ this, CBlob@ blob)
 		return false;
 	}
 
-	//flesh blobs & mines have to be fed into the saw part
-	if (blob.hasTag("flesh") || (name == "mine"))
+	//flesh blobs, mines, trampolines and arrows have to be fed into the saw part
+	if (blob.hasTag("flesh") || (name == "mine") || (name == "trampoline") || (name == "arrow"))
 	{
 		Vec2f pos = this.getPosition();
 		Vec2f bpos = blob.getPosition();


Vam-Jam (approved):
Works in dedicated, seems to be fine.

mehwaffle10:
To be clear, taking cover from archers behind enemy saws is a known and used strategy that this would be a nerf to. This also slightly increases the incentive for players to take team saws to use as one way shields, which some can be viewed as griefing and should not be encouraged without proper caution. On the flip side, this is a buff to archers who can get their hand on an enemy saw in their base against other archers but a nerf until they can actually pick it up and convert it.

mugg91:
@mehwaffle10 
> To be clear, taking cover from archers behind enemy saws is a known and used strategy that this would be a nerf to.

If taking cover from archers shooting from the bottom, then yes. This buffs Saw for the team it belongs to.

Now that you mention it though, "Saw shield" sounds lame.
We don't want archers hiding behind a Saw at the frontline, do we?
Even when Saw can possibly get destroyed quickly and isn't so easy to replace that it will get spammed at the frontline.
Should I remove that change from this PR?

mehwaffle10:
Less so from the bottom, more so running past a saw in a tree farm in the enemy base to take cover from archers. I would think it would be more so griefing than spam, since saws are expensive enough that the other classes can't readily acquire them without taking dropped materials

Issue #2164, opened 2024-11-17
Splashing a filled bucket applies force to the wrong team


mugg91:
## Description

When buying a filled bucket and splashing that, force is applied to enemies and you are standing still. (This is correct)

But when you pick up a filled bucket from the enemy team and splash that, the force will be applied to you and your teammates and any nearby enemy will not be affected. (Bug)

When picking up a filled bucket, the team it belongs to should change.

mugg91:
It might actually have to do with the "damage owner player".
The precise steps I did were this:
- Be team 0, buy filled bucket, put it down
- Switch to team 1, pick it up, use it

Issue #2143, opened 2024-10-22
Bucket sprite does not update (online)


mugg91:
## Description

In online, when bucket turns empty from splashing after colliding, the sprite will not update to empty on its own.
It will only update the sprite after picking up the bucket.

Issue #2142, opened 2024-10-22
Bucket can only splash once (offline)


mugg91:
## Description

In offline, a filled bucket will only splash once before turning empty.

Issue #2068, opened 2024-05-09
Filled bucket will not splash when destroyed with 1 splash remaining


mugg91:
## Description

1) Buy a filled bucket
2) Splash once
3) Change team and be knight
4) Slash the bucket
5) Notice it will just disappear

PR #2165, opened 2024-11-17
[fixed] 7 bugs related to filled bucket

mugg91:
## Status

- **READY**

## Description
```
[fixed] filled bucket will not splash when destroyed with 1 splash remaining
[fixed] Bucket could only splash once (offline)
[fixed] Bucket sprite did not update when used (online)
[fixed] Bucket sprite did not update when falling in water
[fixed] Bucket sprite did not update when hit by a splash hit
[fixed] Bucket sprite did not update after water count is depleted from spam decay hits
[fixed] Splashing a filled bucket applied force to the wrong team
```
Fixes https://github.com/transhumandesign/kag-base/issues/2068
Fixes https://github.com/transhumandesign/kag-base/issues/2143
Fixes https://github.com/transhumandesign/kag-base/issues/2142
Fixes https://github.com/transhumandesign/kag-base/issues/2164

This PR aims to fix bugs related to splashing a filled bucket.

|  | Current build 4762 | After this PR |
| -------------| -------------| -------------| 
| Using filled bucket in offline | Becomes empty after 1 use  | Can be correctly used 2 times  |
| Bucket collides and turns empty | Sprite doesn't update, still says it's full when it actually isn't | Sprite updates to empty |
| Using a filled bucket from the enemy team | Applies force to your team  | Correctly applies force to the enemy team |
| Destroying a filled bucket with 1 splash remaining  | Doesn't splash | Splashes  |

I renamed `TakeWaterCount()` to `DepleteWaterCount()` to make it clear what the function does.

The bug with destroying a filled bucket with 1 splash remaining happens because server depletes uses to 0 and syncs that to the client. So I check if health is equal to or less than 0 in `DepleteWaterCount()` to check if the bucket is being destroyed, then apply and Sync a new Tag `splash on destroy` so client can check for that and correctly splash. I couldn't come up with a different solution.

Some instances of `this.SendCommand(this.getCommandID("splash client"));` now check if we are playing in offline by checking for `if (!(isClient() && isServer()))`, so client doesn't splash twice at the same time.

Added
```
	   SetDamageToCarrier.as;
	   SetTeamToCarrier.as
```
to blob scripts in Bucket.cfg to fix the issue with applying force on the wrong team.

Setting the frame for the bucket after a splash is now done in `DoSplash()` instead of the now server-only `TakeWaterCount()` / `DepleteWaterCount()` function. As a result, only
```
	if (filled == 0)
	{
		filled = 0;
	}
```
remained, serving no purpose so I removed it.

Tested in offline and online, I witnessed no problems.

### DIFF
--- a/Entities/Items/Bucket/Bucket.as
+++ b/Entities/Items/Bucket/Bucket.as
@@ -61,8 +61,7 @@ void onTick(CBlob@ this)
 					!occupiedBlob.isKeyPressed(key_inventory)) // prevent splash when doing stuff with inventory
 				{
 					DoSplash(this);
-					if (!(isClient() && isServer()))
-						this.SendCommand(this.getCommandID("splash client"));
+					this.SendCommand(this.getCommandID("splash client"));
 					this.set_u8("water_delay", 30);
 				}
 			}
@@ -93,6 +92,11 @@ f32 onHit(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@ hitt
 			int id = this.getNetworkID();
 			this.setVelocity(this.getVelocity() + Vec2f(1,0).RotateBy((id * 933) % 360));
 			DepleteWaterCount(this);
+			
+			if (isClient())
+			{
+				SetFrame(this, this.get_u8("filled") > 0);
+			}
 		}
 	}
 
@@ -129,7 +133,7 @@ f32 onHit(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@ hitt
 
 void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
 {
-	if (cmd == this.getCommandID("splash client") && isClient())
+	if (cmd == this.getCommandID("splash client") && isClient() && !isServer())
 	{
 		DoSplash(this);
 	}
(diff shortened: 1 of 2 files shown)


Vam-Jam (dismissed):
Works in dedi + localhost, rejoining also works.

mugg91:
I'm considering changing the check in
```
void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
{
	if (cmd == this.getCommandID("splash client") && isClient())
...
```
to `&& isClient() && !isServer()` and removing the check at line 64 `if (!(isClient() && isServer()));`
to be in line with other PRs I submitted which use the same approach.
But it would require testing all the things again.

mugg91:
I made a commit that applies the change that I mentioned in my previous post.
I tested for all the bugs again in online and offline and it works just as before.

Also fixed a bug with spam decay hits.
When a spam decay hit happens, `DepleteWaterCount()` is run but if water counts turned to zero, the bucket sprite was not updated to empty. The commit I made fixes that so the bucket's sprite is updated after each decay hit.

mugg91:
Set to **IN DEVELOPMENT** while I try to fix more bugs (frame not updating when in water or when hit by splash).

asumagic (approved):

Issue #1222, opened 2022-04-06
Team platform won't always let you pass through


mugg91:
**Current in-game behaviour.**

Under some circumstances, the team platform will not let you pass through it even when you hold the DOWN key (S by default).

- This can happen if you climb down a ladder. You will stop at the team platform and cannot pass through until you press DOWN again. In CTF I https://i.imgur.com/3NN8BQw.jpeg

- If you stand on a solid wall and keep holding DOWN while walking on a team platform, you will not pass through until you press DOWN again.

**Proposed Change**

I would like it if team platforms let you pass through (= make it behave as if it wasn't there) always when you hold DOWN. The only exception being if you were stunned.

mugg91:
Ok. It just feels really wonky when navigating through such a passage. Costs time in CTF and feels odd.

The crouch mechanic only works if players start crouching on a solid block and then walk on a team platform. Like that self harm thing on the other issue, it certainly doesn't feel like intended or like something you could figure out easily. Players can go for it only if they specifically know that team platforms work that way.

I would prefer it if pressing S simply makes you go through team platforms, period. Crouching therefore being impossible on those platforms. If it has already been discussed and decided to be that way, then oh well...

PR #2168, opened 2024-11-18
[changed] crouching on Team Bridge improvements

mugg91:
## Status

## Description

Fixes https://github.com/transhumandesign/kag-base/issues/1222

This PR makes some adjustments to `RunnerMovement.as`. 
**Good:** 
* Can now go down ladder without getting interrupted by your own team bridge. 
* Can now crouch in team bridge and crate when holding `key_action3` before pressing down.
* Can now land on team bridge when holding `key_action3` when holding down.
* Can still crouch on a team bridge, by crouching while standing somewhere solid and then moving onto a team bridge, as before.

**Meh:**
* Knight with bombs can't crouch by holding `key_action3` without lighting a bomb, but can still use the previous method (crouching while standing, then moving onto a bridge).

|  | Current build 4762 | After this PR |
| -------------| -------------| -------------| 
| Holding down when falling down on your team bridge | Fall through | Fall through |
| Pressing down while standing on your team bridge | Fall through | Fall through |
| Going down a ladder while passing a team bridge | Get stuck standing on the bridge until you press down again (Annoying) | Pass through the team bridge without interruption. |
| Holding down when standing, then moving on a team bridge | Doesn't fall through, allowing you to crouch if you stand still. | Doesn't fall through, allowing you to crouch if you stand still. |
| Hold action3 and then down, on a team bridge | Fall through | Crouch |
| Falling down on a team bridge while holding action3 and down | Fall through | Land |

When standing on a crate, you can now hold `key_action3` and then press down in order to crouch, which was not possible before.

However, you cannot land on a crate when holding `key_action3` + `down`. You can land on a bridge, but you will fall through a crate. I haven ot figured out why. Seems not related to `RunnerMovement.as`.

I have tested extensively in offline, should work the same in online.

### DIFF
--- a/Entities/Characters/MovementScripts/RunnerMovement.as
+++ b/Entities/Characters/MovementScripts/RunnerMovement.as
@@ -31,6 +31,7 @@ void onTick(CMovement@ this)
 	const bool right	= blob.isKeyPressed(key_right);
 	const bool up		= blob.isKeyPressed(key_up);
 	const bool down		= blob.isKeyPressed(key_down);
+	const bool action3	= blob.isKeyPressed(key_action3);
 
 	const bool isknocked = isKnocked(blob);
 
@@ -137,6 +138,8 @@ void onTick(CMovement@ this)
 
 		moveVars.jumpCount = -1;
 		moveVars.fallCount = -1;
+
+		blob.set_u8("crouch_through", 3); // allows passing down through team bridges when going down a ladder
 
 		CleanUp(this, blob, moveVars);
 		return;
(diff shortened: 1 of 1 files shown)


Vam-Jam (commented):
Functionally, it works, I cant seem to find any issues with it.

I do think action3 might not be the best key for this, feels like it would be frustrating for builders or people who are activating bombs.

I'll leave it up to others to decide if this is something that should be added or not, as per https://github.com/transhumandesign/kag-base/issues/1222

mehwaffle10:
Action3 definitely feels like the wrong key. Maybe shift/modifier key. I don't know why you would need to crouch on a team bridge, crouching for an enemy will just open the bridge under your feet anyways. I don't think I've ever thought "man I wish I could crouch here". Also, for the last case, wouldn't you rather just *let go of the down key* to not pass through? What case does this handle that that wouldn't?

mugg91:
> I don't know why you would need to crouch on a team bridge

Because people apparently want this to be in the game. See https://github.com/transhumandesign/kag-base/issues/1222#issuecomment-1092927555

This PR makes it so you don't have to start holding down from a solid ground and then keep holding it for as long as you are on the bridge. Seems unintuitive.
Now you can more easily control when to crouch and crouching on crates is possible now, for whatever it's worth, and if it's just for style.

> Also, for the last case, wouldn't you rather just _let go of the down key_ to not pass through? What case does this handle that that wouldn't?

Yes, seems like you have a point. I was thinking there might be situations where you would want to hold down without passing a bridge but I couldn't think of a specific case.

Issue #2074, opened 2024-05-18
Arrows are harvested in mid air when shooting while hanging


mugg91:
## Description

https://github.com/transhumandesign/kag-base/assets/14877281/4ba465e3-5cd3-4f76-951a-b20249e35dc0

Played in online Sandbox

asumagic:
They don't actually collide with nothing, it's because of the arrow harvesting mechanism (from holding RMB) which doesn't actually check if the arrow is stuck on a tile or not.

mugg91:
Ok, I will rename the issue. I hope we can agree that harvesting should happen only when the arrow is stuck.

mugg91:
If you hang and shoot fire arrow, you will set yourself on fire.

PR #2169, opened 2024-11-18
[fixed] don't pick up your own shot arrows from mid air

mugg91:
## Status

## Description

Fixes https://github.com/transhumandesign/kag-base/issues/2074

This PR fixes a bug where you could harvest your own arrows in mid air, preventing you from shooting arrows at low charge while grappling from a ceiling. Arrows can now only be harvested if they are static.

This bug only happened in online. Tested and it doesn't happen anymore on a dedicated server.

Removed a duplicate `if (arrow !is null)` check.


Vam-Jam (approved):
Works

Issue #2170, opened 2024-11-20
Stones from catapult can damage chicken in hand


mugg91:
## Description

Stone shot from your team's catapult can damage the chicken you are holding in your hand.

mugg91:
It affects the following items:
* log
* chicken
* shark
* steak
* fishy
* enemy corpses
* enemy crates
* enemy sponge
* enemy bucket

Basicly anything that can be picked up and doesn't change ownership to your team.
I think items held by your team should not take damage even if the item is of a different team.
I will address this issue soon.

Issue #1267, opened 2022-05-25
Held items should not be damagable


Shylie:
# Current in-game behaviour.

Held items can be damaged.

# Proposed Change

Held items should not be able to be damaged.

# Additional Context

There are several instances where this would be particularly annoying.
There are probably other situations where this could be annoying, as I've only listed a couple.
- Fighting with a drill and the drill gets destroyed by the person you're fighting
- Holding a door to build with and it gets destroyed
- Bringing a dinghy with 30 mines in it to the frontline and it gets destroyed on your way

mugg91:
It would seem that you can climb on a ladder that an enemy builder is holding. But I'm not certain. That should also not be happening I guess.

mugg91:
Someone said they noticed this behavior:
You buy a boulder. Your teammates slash at the boulder while it is in your hand. The boulder gets damaged.
Needs testing to confirm this.

PR #2172, opened 2024-11-20
[fixed] Don't damage items held by teammates

mugg91:
## Status

- **ON HOLD**
- I have to test for keg and ballista bomb bolts.

## Description
```
[changed/fixed] Rocks from own catapult don't hurt item held by teammate
[changed/fixed] Ballista Bolt from own ballista doesn't hurt item held by teammate
[changed/fixed] Own Drill doesn't hurt item held by teammate
[changed/fixed] Own Bomb explosion doesn't hurt item held by teammate
```

Fixes https://github.com/transhumandesign/kag-base/issues/2170
Might fix some symptoms described here https://github.com/transhumandesign/kag-base/issues/1267

This PR makes it so own damage sources (bomb explosion, drill, ballista bolt, catapult rocks) don't hurt own held items (log, fishy, chicken, enemy corpse, enemy crate, etc.).

I removed this code from `Rock.as` and `BallistaBolt.as`:
```
	CBlob@ carrier = blob.getCarriedBlob();

	if (carrier !is null)
		if (carrier.hasTag("player")
		        && (this.getTeamNum() == carrier.getTeamNum() || blob.hasTag("temp blob")))
			return false;
```
This code doesn't seem to serve any purpose. It basicly means, if blue player A holds red player B, then player A doesn't take damage from red catapult rocks. The only time a player can hold another player is when they are dead, but then they get untagged `"player"` so this code never even runs. Let me know if it does serve any purpose, though.

## Reproduction

You can put a chicken in your catapult, then try to drill it or place a bomb or shoot with another catapult at it.
Prior to this PR, it takes damage, after this PR it doesn't.

### DIFF
--- a/Entities/Industry/Drill/Drill.as
+++ b/Entities/Industry/Drill/Drill.as
@@ -300,7 +300,20 @@ void onTick(CBlob@ this)
 								{
 									continue;
 								}
-
+								
+								// don't hit if carried by teammate
+								if (b.isAttached())
+								{
+									AttachmentPoint@ ap = b.getAttachments().getAttachmentPointByName("PICKUP");
+									
+									if (ap !is null)
+									{
+										CBlob@ occ = ap.getOccupied();
+										
+										if (occ is null || occ.getTeamNum() == this.getTeamNum()) 
+											continue;
+									}
+								}
 
 								if (isServer())
 								{
--- a/Entities/Items/Explosives/Explosion.as
+++ b/Entities/Items/Explosives/Explosion.as
@@ -294,10 +294,23 @@ void Explode(CBlob@ this, f32 radius, f32 damage)
 			if (hit_blob is this)
 				continue;
 
+			// don't hit if carried by teammate
+			if (hit_blob.isAttached())
+			{
+				AttachmentPoint@ ap = hit_blob.getAttachments().getAttachmentPointByName("PICKUP");
+				
+				if (ap !is null)
+				{
+					CBlob@ occ = ap.getOccupied();
+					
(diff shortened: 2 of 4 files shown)


mugg91:
I added some fixes to `Explosion.as` and added changes to `KnightLogic.as`.
Bomb and Keg will not hit chicken in teammate's hand or in team's catapult.
Bomb Ballista Bolts are also working correctly.
Knight's jabs and slashes won't deal damage to chicken if it is held by your teammate or catapult.

Feel free to re-review this.

mugg91:
Now all the relevant script files will use the same check function `isHeldByTeammate()`. I placed it in `Hitters.as` since it made most sense to put it there, since all the relevant script files include `Hitters.as` anyway.
I tested it for a while in offline.
No more issues known.

Ready for reviewing.

Entities/Common/Attacks/Hitters.as
```diff
@@ -78,3 +78,29 @@ bool isWaterHitter(u8 type)
...
+		{
+			for (uint i = 0; i < aps.length; i++)
+			{
+				AttachmentPoint@ ap = aps[i];
+
+				CBlob@ occ = ap.getOccupied();
+
+				//if (occ !is null)	print(ap.name + " " + occ.getName());
+
+				if (occ !is null && hitter_blob !is null && occ.getTeamNum() == hitter_blob.getTeamNum())
```
Vam-Jam, line 97: small nitpick, check hitter_blob at the same place you check held_blob, there's no reason to loop if thats null

Vam-Jam (approved):
Code wise, it works fine, but it needs testing in a real match just to make sure it does not break anything

PR #2175, opened 2024-11-20
[changed] Better Boulder map collision sounds

mugg91:
## Status

## Description

This PR improves sounds that Boulder makes when landing on the ground. Previously, it made strange wooden sounds which is unfitting for a stone item.

I kept Boulder as a medium weight item and introduced a new tag `"heavy sound"` which is added to Boulder in `onInit()` and checked for in `Stone.as`. If Boulder falls on the ground, one of two possible thud sounds will play which I added to the game, taken from Pixabay ( https://pixabay.com/de/sound-effects/thud-me-70412/ ). If it collides with a castle tile or wood tile, a different sound will play.

## Before

https://github.com/user-attachments/assets/48a7090e-a40c-4aaf-a1a4-cbadfb5b6984

## After

https://github.com/user-attachments/assets/ee11ec7a-a6ea-4af9-95b8-f5fd3041b889


Entities/Common/Fabric/Stone.as
```diff
@@ -41,13 +41,17 @@ void onCollision(CBlob@ this, CBlob@ blob, bool solid, Vec2f normal, Vec2f point
 		return;
 	}
 
-	if (!getNet().isServer())
+	if (!isServer())
```
Gingerbeard5773, line 44: this is a bug, none of the client only code below will be executed on dedicated server

Gingerbeard5773 (changes requested):

Issue #2027, opened 2024-04-20
Make TDM shop destructible


mugg91:
## Current in-game behaviour.

* You can hit the shop, and it makes a "destroyed" sound after a few hits, but nothing actually happens.

## Proposed Change

* Make it possible to actually destroy it
* Shop should gib into pieces and either the Trader should die or should be spawned as a single trader possible to kill.

Adds strategy.
You can prevent other players from purchasing.
If a player camps at the shop, other player might feel more inclined to attack.
Also it is fun to destroy something.

PR #2177, opened 2024-11-21
[added] TDM shop can be destroyed

mugg91:
## Status

## Description

Fixes https://github.com/transhumandesign/kag-base/issues/2027

This PR makes it so TradingPost of the other team can be destroyed (only in Sandbox and in TDM).
When destroyed, it spawns a trader of the same gender as the spritelayer that can be killed.

I made it so gender of the spritelayer depends on Network ID so it is the same for everyone and nobody will say 
"I killed the trader woman"
"Weird, it's a male trader for me"

Also fixed a check in `onHealthChange()`, from `if (oldHealth > 0.0f && this.getHealth() < 0.0f)` to `if (oldHealth > 0.0f && this.getHealth() <= 0.0f)`.

Tested in offline and online, no problems found.

https://github.com/user-attachments/assets/a3bf6c9d-39d8-4736-9ea2-9d4a8aaca3e2

## Remaining Issues

Some maps have blue trading post in the center, so only red team could damage it. Those maps would need to be changed to have a trading post that isn't team 0 or 1. Or we could allow hits to be processed regardless if it's the same team.
TDM shop might be destroyed too quickly, so health could be increased or attack's damage lowered.
Overall, I'm satisfied though.

### DIFF
--- a/Entities/Industry/Trading/TradingPost.as
+++ b/Entities/Industry/Trading/TradingPost.as
@@ -13,7 +13,9 @@ void onInit(CBlob@ this)
 	if (sprite !is null)
 	{
 		sprite.SetZ(-50.0f);   // push to background
-		string sex = traderRandom.NextRanged(2) == 0 ? "TraderMale.png" : "TraderFemale.png";
+		u8 trader_sex_num = this.getNetworkID() % 2;
+		string sex = (trader_sex_num == 0) ? "TraderMale.png" : "TraderFemale.png";
+		this.set_u8("trader sex num", trader_sex_num);
 		CSpriteLayer@ trader = sprite.addSpriteLayer("trader", sex, 16, 16, 0, 0);
 		trader.SetRelativeZ(20);
 		Animation@ stop = trader.addAnimation("stop", 1, false);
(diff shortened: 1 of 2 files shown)


Entities/Industry/Trading/TradingPost.as
```diff
@@ -110,18 +107,41 @@ f32 onHit(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@ hitt
...
+		return;
+	
+	if (oldHealth > 0.0f && this.getHealth() <= 0.0f)
 	{
-		MakeDustParticle(this.getPosition(), "Smoke.png");
-		this.getSprite().PlaySound("/BuildingExplosion");
+		// spawn trader that can be killed
+		if (this.exists("trader sex num"))
+		{
+			CBlob@ trader = server_CreateBlobNoInit("trader");
```
Vam-Jam, line 130: You should be calling trader.Init() + doing a null check

Clients can run this code and end up hitting a null error.

Vam-Jam (changes requested):

mehwaffle10:
I don't think this is a good change. I think it'd lead to more frustration than any interesting gameplay. The focus of TDM should be on fighting each other not on map destruction. Is there anything in particular this is meant to address?

mugg91:
> I don't think this is a good change. I think it'd lead to more frustration than any interesting gameplay. The focus of TDM should be on fighting each other not on map destruction. Is there anything in particular this is meant to address?

Just for fun, adding more aspects and things you can do to the game. 
I will implement more fixes and then it's up to the community if they want it or not.

GELGELAT:
like it

PR #2179, opened 2024-11-23
Command Quality of Life

mehwaffle10:
Quality of life improvements for the new command system. Allow syntax prior to the command rework. Tested on my balance mod since the command rework was added.

## Status

## Description

Address some common issues when using commands. Makes the following changes:
- Ignore character case when using commands
- Allow ! instead of / when using commands
- Default to the spawn command if no other command matches

## Steps to Test or Reproduce

1. Enable commands by setting sv_test to 1
2. Type !Keg to spawn a keg

git branch used: https://github.com/mehwaffle10/kag-base/tree/command-qol

### DIFF
--- a/Rules/CommonScripts/ChatCommandCommon.as
+++ b/Rules/CommonScripts/ChatCommandCommon.as
@@ -1,3 +1,5 @@
+const string[] prefixes = {"/", "!"};
+
 namespace ChatCommands
 {
 	ChatCommandManager@ getManager()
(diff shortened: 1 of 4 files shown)


Vam-Jam (approved):
Works, small request but for localhost servers, could you allow people to do !keg [count] to spawn multiple objects at once, e.g. !keg 20 will spawn 20 kegs

It's not needed for normal sandbox as it would be abused pretty heavily but for localhost, it would be extremely handy.

mugg91:
I think it would be better to have acceptable prefixes ("/" and "!") in one place rather than setting "/" and checking for "!" later.

asumagic (approved):

Issue #2066, opened 2024-05-05
Teammate builders should indicate if they have gold in their inventory


mugg91:
## Describe the feature

Teammate builders should indicate if they have gold in their inventory.
So teammates know who has the gold and so gold can't be hidden with bad intentions (grief).
Opponent builders should not indicate if they have gold.

PR #1761, opened 2023-12-15
Scripted playernames.

GoldenGuy:
## Status

- **READY**: this PR is (to the best of your knowledge) ready to be incorporated into the game. But can still be tweaked if something interesting is suggested.

Fixes #282

## Description

Moves player names and hearts ui from engine to scripts. Also added fade-in behavior.
Along with this pr there should be an engine pr to remove names from engine.

## Steps to Test or Reproduce

1) join server/sinlgeplayer
2) spawn bots
3) see this:

https://github.com/transhumandesign/kag-base/assets/26719582/44094844-4805-4258-a36c-19d332a0ee23

https://github.com/transhumandesign/kag-base/assets/26719582/49160f41-c2cd-4f9b-bf39-78aa22b06200

PR #2182, opened 2024-11-24
[added] Indicator that shows how much gold is in your team's inventories

mugg91:
## Status

## Description

```
[added] Indicator that shows how much gold is in friendly inventories
```

Fixes https://github.com/transhumandesign/kag-base/issues/2066

This PR adds a feature to indicate gold in your teammates' inventories as well as friendly Crates, Dinghies, Outposts, etc.

The blobs not showing gold are:
- you
- any blob that isn't from your team
- storage (because it already indicates if it has gold)

This is so you quickly know where your team's gold is and people can't grief by hiding gold in their inventory.

The gold indicator will be shown to the right of the health bar. It looks like this:

For players, the indicator will shift to the right the more hearts they have.

I have tested it in offline mostly, only a little in online but it looks to be working the same.

### DIFF
--- a/Entities/Materials/Construction/MaterialGold.as
+++ b/Entities/Materials/Construction/MaterialGold.as
@@ -28,7 +28,10 @@ void onRender(CSprite@ this)
 	if (localBlob is null) return;
 
 	CBlob@ invBlob = blob.getInventoryBlob();
-	if (invBlob is null || localBlob.getTeamNum() != invBlob.getTeamNum() || localBlob is invBlob) return;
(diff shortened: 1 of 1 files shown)


Vam-Jam:
You'll want to check that this pr works with this other PR that moves hearts script side, as it will be merged at some point.
https://github.com/transhumandesign/kag-base/pull/1761

mugg91:
I would like it if people discuss if this is something they want in the game, before I continue.
Have not looked at the code of the script-side player names PR yet.

Issue #2005, opened 2024-04-10
Kill feed should account for trampoline kills


mugg91:
## Current in-game behaviour.

Player A holds trampoline, player B bounces off it and dies (ground or void).
Kill feed should say player A killed player B by trampoline.
Not sure if easy to implement.
Just an idea..

PR #2184, opened 2024-11-24
[added] Shark, Bison & Trampoline to Kill feed

mugg91:
## Status

## Description
```
[added] Bison, Shark & Trampoline to kill feed
```

Fixes https://github.com/transhumandesign/kag-base/issues/2005

This is the same as this PR https://github.com/transhumandesign/kag-base/pull/1861 with the following adjustments:
By changing `RunnerDefault.as`, `TrampolineLogic.as` and `FallDamage.as`, Trampoline is now added to the kill feed.
It works like this:
When getting bounced on the trampoline, current game time is saved to the blob via `set_s16("trampoline last bounce")` and `blob.server_Hit(blob, blob.getPosition(), blob.getVelocity(), 0.0f, Hitters::fall_trampoline, true);` is applied, dealing no damage but telling the game that this is the most recent hit dealt to the player.

When the player falls in the void in the next 10 seconds, it counds as a trampoline death in the kill feed (same behavior as with any other attack source).

When the player dies by fall damage within the next 2 seconds, it counds as a trampoline death, otherwise as a normal fall death.

When the player takes any attack before taking the fall damage, `"trampoline last bounce"` is cleared.
Bouncing → getting hit by a spikes → falling down and dying will count as a normal fall death.
Only bouncing and directly falling to your death (within 2 seconds) will count as a trampoline death.

Tested in offline and online, no problems found.


Entities/Common/Movement/FallDamage.as
```diff
@@ -53,7 +53,15 @@ void onCollision(CBlob@ this, CBlob@ blob, bool solid, Vec2f normal, Vec2f point
 
 			if (damage > 0.1f)
 			{
-				this.server_Hit(this, point1, normal, damage, Hitters::fall);
+				if (this.exists("trampoline last bounce") && this.get_s16("trampoline last bounce") + 60 > getGameTime())
```
Vam-Jam, line 56: u32

Entities/Characters/Scripts/RunnerDefault.as
```diff
@@ -75,5 +75,9 @@ f32 onHit(CBlob@ this, Vec2f worldPoint, Vec2f velocity, f32 damage, CBlob@ hitt
 {
 	if (this.hasTag("invincible") && customData == 11)
 		this.Untag("invincible");
+
+	// when runner is hit, forget last trampoline bounce
+	this.set_s16("trampoline last bounce", -1);
```
Vam-Jam, line 80: u32

Entities/Structures/Trampoline/TrampolineLogic.as
```diff
@@ -138,6 +139,10 @@ void onCollision(CBlob@ this, CBlob@ blob, bool solid, Vec2f normal, Vec2f point
 			velocity.RotateBy(angle);
 
 			blob.setVelocity(velocity);
+			
+			blob.server_Hit(blob, blob.getPosition(), blob.getVelocity(), 0.0f, Hitters::fall_trampoline, true);
```
Vam-Jam, line 143: Not 100% sure about this, some onHit hooks may react differently based on parameters. It would need checking for edge cases.

Entities/Structures/Trampoline/TrampolineLogic.as
```diff
@@ -138,6 +139,10 @@ void onCollision(CBlob@ this, CBlob@ blob, bool solid, Vec2f normal, Vec2f point
...
+
+			blob.set_s16("trampoline last bounce", getGameTime());
```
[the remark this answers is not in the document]
mugg91: Yes, I keep mixing it up with network IDs which are usually saved in u16. Will fix.

Entities/Structures/Trampoline/TrampolineLogic.as
mugg91: I will check onHit() on every blob. If there is anything awry, we can still resort to using the PR without the Trampoline.

mugg91: The only thing I found is Bison will run `MadAt()` when touching your trampoline. Should be easy to fix.

mugg91: I found that watersplash will also cause Bison to become mad.
Should I add a check if damage is above 0? Or should I check for `Hitters::trampoline` specifically?

Vam-Jam: I would check if damage is above 0

mugg91:
I added a check to `Bison.as` in `onHit()` to see if the damage is above 0.
Water splash hit and Trampoline bounce hit won't enrage Bison anymore.
Fixes https://github.com/transhumandesign/kag-base/issues/2229

mugg91:
I moved `Bison.as` up in the blob script list in `Bison.cfg` to fix that damage in `onHit()` was always 0.
I'm not sure if there could be complications. Let me know if it's ok or what could be done better.

Issue #2183, opened 2024-11-24
Sometimes twice the fall damage is applied


mugg91:
## Description

On both client and server, I found that `onCollision()` in `FallDamage.as` could be run twice on the same tick, with unchanged velocity and old velocity, if getting bounced to the ground onto a corner like in the screenshot below. 

The bug happens because of the corner I'm standing on. It does not happen if it's two blocks.

Because of the bug, you can take the same damage twice (1 heart instead of 0.5 hearts). 

![screen-24-11-24-18-26-20](https://github.com/user-attachments/assets/a526d38e-e439-4bf2-896f-ecd1a56c2aa5)

I'm not sure why it happens.

PR #2185, opened 2024-11-24
[fixed] You no longer can take fall damage twice on the same tick

mugg91:
## Status

## Description

```
[fixed] You no longer can take fall damage twice on the same tick
```

Fixes https://github.com/transhumandesign/kag-base/issues/2183

When applying the fall damage hit in `FallDamage.as`, the game time is saved to the blob.
At the beginning of `onCollision()` the code drops out if it detects that it's the same tick again so it doesn't apply the hit twice.

I'm not sure why it happens in the first place. This is just a simple fix I came up with. Of course it could be worth looking for the root of the problem.

Tested in offline and online, works correctly. 

## Reproduction
Build a construction like in the screenshot from the issue linked above.
You have to be standing on a corner with the block being on your left or right side of the character.
When getting bounced down, you are taking two hits for 1 heart of total damage. **After this PR, it is only one hit for a total of 0.5 hearts damage**

### DIFF
--- a/Entities/Common/Movement/FallDamage.as
+++ b/Entities/Common/Movement/FallDamage.as
@@ -17,7 +17,7 @@ void onCollision(CBlob@ this, CBlob@ blob, bool solid, Vec2f normal, Vec2f point
 	if (!solid 
 		|| this.isInInventory()
 		|| this.hasTag("invincible")
-		|| (this.exists("last fall hit") && this.get_u16("last fall hit") == getGameTime()))
+		|| (this.exists("last fall hit") && this.get_u32("last fall hit") == getGameTime()))
 	{
 		return;
 	}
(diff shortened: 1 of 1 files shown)


Vam-Jam:
> I'm not sure why it happens in the first place

It's to do with box2d, each tile is made up of segments, on a corner, you can touch two segments, so the onCollision hook gets called twice since the game reports you touched two different segments at fast enough speeds.

Entities/Common/Movement/FallDamage.as
```diff
@@ -54,11 +57,14 @@ void onCollision(CBlob@ this, CBlob@ blob, bool solid, Vec2f normal, Vec2f point
 			if (damage > 0.1f)
 			{
 				this.server_Hit(this, point1, normal, damage, Hitters::fall);
+				this.set_u16("last fall hit", getGameTime()); // fixes bug of this code running twice in the same tick
```
Vam-Jam, line 60: getGameTime() returns a u32, not a u16.

Vam-Jam (changes requested):

asumagic (approved):
The approach seems sound enough to me. I was thinking of possible edge cases where it would be possible for fall damage to actually need to be applied twice within the same tick (e.g. fast collision with some tile, then sliding off and hitting something again?), but I'm not sure this even ever happens in practice.

Issue #2096, opened 2024-06-17
Ladder takes water damage when not inside water


mugg91:
## Description

![screen-24-06-17-09-49-16](https://github.com/transhumandesign/kag-base/assets/14877281/a8d4da03-2406-4023-b830-9f8bfe8dad08)

Issue #2077, opened 2024-05-27
Make water behave in less odd ways


mugg91:
## Current in-game behaviour.

Sometimes water looks odd after using sponge. For example, what looks like water is coming out of dirt - a waterfall - will look like a body of water floating in the air. It also sometimes looks odd that water spreads all over the map when it should just be falling down into the void.

## Suggestion

1) 
Add a water source tile that cannot be drained where water will always come out.

Example:
![screen-24-05-27-15-06-08](https://github.com/transhumandesign/kag-base/assets/14877281/6c62e4ec-7370-45d5-b197-bd6805211d31)

A sponge has been used on the left side, which looks odd because it looks like water coming out of thin air.
The topmost water tiles should be "water source tiles" where water will continue to come out and that cannot be drained.

----

2) 
Make water not spread sideways if there is no ground somewhere below the water tile.

Example:
![screen-24-05-27-15-03-01](https://github.com/transhumandesign/kag-base/assets/14877281/517963fd-628e-453a-8985-33a9e69c8930)

When placing the stone tile there, water would spread to the left. Since it falls down into the void, it should not spread to the left.

PR #2187, opened 2024-11-25
[fixed/changed] Ladders placed next to waterfalls won't decay

mugg91:
## Status

## Description
```
[fixed/changed] Ladder decays next to waterfalls even though it doesn't touch the water
[fixed] Ladder placed in a 1 tile water puddle didn't decay
```
Fixes https://github.com/transhumandesign/kag-base/issues/2096

`WaterDecaysDoors.as` says that water should be checked for the tile **above**, **left of** and **right of** ladders and wooden doors.

This PR changes that to check for water in the blob's tile position (**center**). This fixes that ladder placed inside in a 1 tile water puddle doesn't decay.
Also, **above**, **left** and **right** are only checked if shape consts `waterPasses` returns false, therefore allowing ladders to be placed to the left or right of waterfalls without decaying.

(It's worth noting that wooden doors cannot be placed next to waterfalls without water spreading on top of it.
Please consider these suggestions for improving the water behavior https://github.com/transhumandesign/kag-base/issues/2077 - the 2nd example suggests that placing anything solid (including a wooden door) shouldn't spread water on top of it if it's a waterfall leading into the void.)

Instead of fetching shape and shape consts, you could also simply tag Ladder blobs and check for that tag, but I didn't want to add clutter with tags when the problem could be solved without using tags.

### DIFF
--- a/Rules/CommonScripts/WaterDecaysDoors.as
+++ b/Rules/CommonScripts/WaterDecaysDoors.as
@@ -60,12 +60,25 @@ void onTick(CRules@ this)
 		}
 
 		Vec2f pos = b.getPosition();
-		Vec2f[] postocheck =
+		
+		CShape@ shape = b.getShape();
+		if (shape is null) continue;
+		
+		ShapeConsts@ consts = shape.getConsts();
+		if (consts is null) continue;
+
+		Vec2f[] postocheck;
+
+		postocheck.push_back(pos);
+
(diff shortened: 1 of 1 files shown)


Vam-Jam (approved):
Works from testing in localhost + dedi. Could not find any weirdness compared to vanilla, may need more testing.

Issue #1738, opened 2023-11-04
Swapping into other class issues


mugg91:
## Description

In offline, you can now use `/class blob` in order to change into any blob you want.

## Issues

1) If you are anything but Knight, Archer, Builder, Trader, you will not be able to zoom in or out.
2) As `bison` or `shark,` moving left will not update the sprite as left-facing.
3) As `fishy` or `shark`, you can move around on land. But you cannot move at all while in water.
4) If you despawn while being anything other than Knight, Archer, Builder, Trader, the game will respawn you once. Next time it happens you will get an exception error, though. This happens in Sandbox but not in Save the Princess.
5) If the blob you try to swap to doesn't exist, the game will despawn you and never respawn you again.

## Suggestions

Playing as princess, necromancer or bison is cool.
We could try to add support for online sandbox.
The game should prevent you from swapping into inanimate things.

PR #2189, opened 2024-11-26
[fixed] Don't allow changing to non-existing class

mugg91:
## Status

## Description

Fixes 5) in https://github.com/transhumandesign/kag-base/issues/1738 

`/class` will now check if the blob created matches the class name you entered, same method used as in the `/spawn` command, therefore preventing you from changing class to a blob that doesn't exist.


Vam-Jam:
I've submitted an engine patch to fix this.

Issue #2095, opened 2024-06-16
Sponge should not become dry inside water


mugg91:
## Description

Sponge will dry and turn into a "almost but not quite soaked" state when left alone inside water.
Sponge should only dry over time outside of water.

PR #2190, opened 2024-11-26
[fixed] Sponge will not dry in water

mugg91:
## Status

## Description

Fixes https://github.com/transhumandesign/kag-base/issues/2095

Fully soaked sponge could dry while floating in water if it ticks while its center is above water.
Logically a sponge that's on the water should not become dry.

This PR changes the check in `SpongeDryInAir.as` `onTick()` so it checks for the sponge's bottom position rather than using `this.isInWater()` (which seems equivalent to checking for its center position).

## Reproduction

When making a huge sea and placing a bunch of sponges inside, every few seconds a sponge would dry and drain some water for a short moment. After applying this PR, this doesn't happen.

### DIFF
--- a/Entities/Items/Sponge/SpongeDryInAir.as
+++ b/Entities/Items/Sponge/SpongeDryInAir.as
@@ -11,7 +11,10 @@ void onInit(CBlob@ this)
 
 void onTick(CBlob@ this)
 {
-    if (not this.isInWater())
+	Vec2f bottom_pos = this.getPosition() + Vec2f(0, this.getHeight() / 2);
+	CMap@ map = getMap();
+
+	if (!map.isInWater(bottom_pos))
     {
         u8 absorbed = this.get_u8(ABSORBED_PROP);
         absorbed = Maths::Max(0, absorbed - slow_frequency);


Entities/Items/Sponge/SpongeDryInAir.as
```diff
@@ -11,7 +11,10 @@ void onInit(CBlob@ this)
 
 void onTick(CBlob@ this)
 {
-    if (not this.isInWater())
+	Vec2f bottom_pos = this.getPosition() + Vec2f(0, this.getHeight() / 2);
+	CMap@ map = getMap();
+
+	if (!map.isInWater(bottom_pos))
```
Vam-Jam, line 17: Fix the indent.

While this might fix them drying in water, if water comes from the top or from the sides, this check will still report as it being in air.

I would instead do something like:
- Have we not removed water in the last 15(?) ticks
- Our velocity is 0 (e.g. we're on land) & not attached to anything

Vam-Jam (changes requested):

Entities/Items/Sponge/SpongeDryInAir.as
mugg91: Ok, I will do something.

mugg91:
I made fixes as requested.

Since `Sponge.as` already checks left, top, right and bottom of Sponge, I used that to save the last time any of those positions had water.

`SpongeDryInAir.as` will only dry sponge if no water was found in the last 15 ticks and if there is any absorbed amount to dry in the first place. 

I did not want to use velocity checking because I think sponge should dry in the air even when moving or bouncing on a trampoline.

PR #2191, opened 2024-11-27
[added] Arrows will ignite when shot through burning platforms

### LINKED ISSUE #1869: Make arrows catch fire when shot through a burning platform
## Current in-game behaviour.

When arrows are shot through a burning platform they are not set on fire.

## Proposed Change

When arrows are shot through a burning platform they are set on fire.

mugg91:
## Status

## Description
```
[added] Arrows will ignite when shot through burning platforms
```
Fixes https://github.com/transhumandesign/kag-base/issues/1869

Previously, arrows could be ignited if shot through world points that are on fire or if shot through a burning fireplace.
This PR adds that arrows are set on fire also when shot through a burning platform, even if the platform's background is stone.
I added this feature because it makes sense and might open up new strategies.

Tested and verified in online and offline.

## Reproduction

Place a wood block and some away-facing platforms on top. Place stone background next to that structure and on the wooden platforms.
Shoot at the wooden block with a fire arrow and then try to shoot normal arrows through the wooden platforms when they start to burn.
Notice they don't get set on fire. 
**After this PR, they will get set on fire.**


mugg91:
## Before

https://github.com/user-attachments/assets/5e456e91-84cd-4eba-8693-b082819bb86d

## After

https://github.com/user-attachments/assets/5cf35df4-8ea4-41af-bd5b-1266d78b04c5

(Played on dedicated server)

Issue #2216, opened 2024-12-01
Bomber and Airship conversion graphics are not shown


mugg91:
## Description

Although Bomber and Airship can be converted, the graphics for it are not shown.

Issue #2211, opened 2024-11-30
Airship and Bomber should not decay and convert at the same time


mugg91:
## Description

Airship and Bomber will decay if no teammate is close and an enemy is close.
However, they can also be converted which is odd.
Either allow converting without decaying, or don't allow converting.

I know we are talking about unused/old vehicles but I'm still reporting any bugs that I find.

Originally addressed in https://github.com/transhumandesign/kag-base/pull/1251

Issue #2206, opened 2024-11-30
Bomber and Airship should show healthbar


mugg91:
## Current in-game behaviour.

All vehicles, Longboat and Warboat show their healthbar on mouse hover, but airship and bomber don't.

https://github.com/user-attachments/assets/c9c3074a-4695-4614-b0ef-4cd9f7b1f24e

Originally addressed in https://github.com/transhumandesign/kag-base/pull/1251

PR #2217, opened 2024-12-01
[fixed] Airship & Bomber issues

mugg91:
## Status

## Description
```
[fixed] Team conversion bar for Airship and Bomber wasn't shown
[changed] Airship and Bomber will not decay if left with an enemy
[changed] Show healthbar for Airship and Bomber
```

Fixes https://github.com/transhumandesign/kag-base/issues/2216
Fixes https://github.com/transhumandesign/kag-base/issues/2211
Fixes https://github.com/transhumandesign/kag-base/issues/2206

Although Airship and Bomber can be converted, the graphics for it weren't shown. Now they are.

Although Airship and Bomber can be converted, they will decay if left with an enemy. Converting and decaying at the same time is stupid. So I removed the decaying when left with an enemy.

Although other vehicles and boats have a healthbar, Bomber and Airship didn't have one. Now they do.

### DIFF
--- a/Entities/Vehicles/Airships/Airship.cfg
+++ b/Entities/Vehicles/Airships/Airship.cfg
@@ -11,6 +11,8 @@ $sprite_factory                            = generic_sprite
 											 FireAnim.as;
 											 Airship.as;
 											 DynamicForegroundLayer.as;
+											 HealthBar.as;
+											 VehicleConvert.as;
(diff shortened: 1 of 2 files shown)


Vam-Jam (approved):
Works fine, I don't think this will change anything in vanilla as these two blobs don't typically get used.

Issue #2214, opened 2024-11-30
Keg should not collide with some blobs


mugg91:
## Description

1. Pick up a Keg
2. Walk through some hearts or Saw
3. Notice the hearts and Saw get moved around, while the Keg is in your hand.

While Keg should get hit by Arrows, sword slashes and other things to fall out of your hand, it should not shift things around that lie on the ground.

---
1. Spawn a catapult and a Keg
2. Load the Keg on the catapult
3. Try to throw other blobs through the Keg (Saw, hearts, etc.)
4. Notice the things you throw will collide with the Keg.

I think this should not happen.

PR #2218, opened 2024-12-01
[fixed] Attached Keg collision issus

mugg91:
## Status

## Description
```
[fixed] Keg in hand will not push blobs on the ground (Saw, Heart, ...) around anymore
[fixed] Blobs thrown at an attached Keg will not collide with it anymore 
        except for Boulder, Arrow, BallistaBolt and Rock
```

Fixes https://github.com/transhumandesign/kag-base/issues/2214

First I tried setting `this.getShape().getConsts().collideWhenAttached = true;` to `false` but then nothing would collide with attached Keg anymore. So I used  

```
{
	if (blob is null)
		return false;

	bool should_collide = this.isAttached() ? (blob.hasTag("projectile hits keg") && blob.getShape().vellen > 0.01f) : true;
	return should_collide;
}
```

If Keg is attached, only blobs that are moving and that have tag "projectile hits keg" can collide with Keg.
If Keg is not attached, then other blobs can collide with Keg as usual.

Boulder, Arrow, Ballista Bolt and catapult Rocks are given the new tag "projectile hits keg".
Other blobs that don't have this tag will not collide with an attached Keg.
So bucket or Saw thrown at Keg (which is loaded in a Catapult) will not collide with it anymore.

I considered using existing tag `"projectile"` but Rock and Boulder don't have this and blobs with that tag will set off mines so I couldn't just give Rock and Boulder that tag. So although I hate to do it I had to introduce the new tag "projectile hits keg".

I want only blobs to collide with an attached Keg that can actually hit it and therefore detach it.
Let me know if there are better approaches.

Tested in offline and online, works as envisioned.

For testing steps, see the linked issue above.

### DIFF
--- a/Entities/Items/Boulder/Boulder.as
+++ b/Entities/Items/Boulder/Boulder.as
@@ -13,6 +13,7 @@ void onInit(CBlob @ this)
 	this.set_u8("launch team", 255);
 	this.server_setTeamNum(-1);
 	this.Tag("medium weight");
+	this.Tag("projectile hits keg");
 
 	LimitedAttack_setup(this);
 
--- a/Entities/Items/Explosives/KegVoodoo.as
+++ b/Entities/Items/Explosives/KegVoodoo.as
@@ -242,3 +242,12 @@ void onThisAddToInventory(CBlob@ this, CBlob@ inventoryBlob)
 		this.doTickScripts = true;
 	}
 }
+
+bool doesCollideWithBlob(CBlob@ this, CBlob@ blob)
+{
+	if (blob is null)
+		return false;
+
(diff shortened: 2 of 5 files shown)


mugg91:
This video shows the issues that this PR fixes.

https://github.com/user-attachments/assets/9d6a50e3-38df-4aff-8848-350e81ac1e4d

mugg91:
It was requested that Keg should collide with Keg in hand. I will look into it.
Perhaps tag `projectile hits keg` should be added to Keg itself but I would change the logic so that it only actually collides if it's going to explode from the collision.

Vam-Jam (approved):
Works in offline & online, tested with two players

PR #2202, opened 2024-11-30
[fixed] Lever and Coinslot didn't change sprite & play sound when interacting with them

mugg91:
## Status

- **READY**: this PR is (to the best of your knowledge) ready to be incorporated into the game.

## Description

Fixes https://github.com/transhumandesign/kag-base/issues/2200 , fixes https://github.com/transhumandesign/kag-base/issues/2201

Lever and coinslot didn't make a sound and change their sprite when interacting with them.
Turns out the client command wasn't getting sent.

Works in offline and online now.

PR #2219, opened 2024-12-03
[fixed] mechanisms

### LINKED ISSUE #1274: Motion Sensor and Obstructor can float
# Description

A Motion Sensor or Obstructor (in its background state) can float indefinitely, if you first place it, then place background behind it, then destroy the supporting tile. The background will fall down but the blob itself will stay afloat.
You can touch and build around the floating tile and it will never fall down. It can still be destroyed though.

https://i.imgur.com/4PoaNws.png

This issue is not fixed by PR #1243 .

# Reproduction Steps

- Go to Sandbox.
- Place a ladder.
- Place a Motion Sensor or Obstructor on top of the ladder.
- Place a wooden background on the Motion Sensor or Obstructor.
- Destroy the ladder.
- Notice the background falling down, but the blobs themselves staying afloat.

### LINKED ISSUE #1874: Obstructor hit by Keg will become bugged

### LINKED ISSUE #1875: Some Components will leave blackness when destroyed

### LINKED ISSUE #2004: Components and devices have wrong frame when rejoining server

### LINKED ISSUE #2152: Spiker sometimes deals 2 hearts damage
## Description

Spiker triggered by motion sensor will sometimes deal 2 hearts of damage instead of 1.

Tested and verified in online Sandbox.

This is of significance in TDM since same traps exist there and one Spiker should not deal more than 1 damage per hit.

### LINKED ISSUE #2200: Coin Slot doesn't make coin sound
## Description

When entering 60 coins, it doesn't make a sound anymore. Both in online and offline.

### LINKED ISSUE #2201: Lever doesn't toggle sprite
## Description

When using a lever, it doesn't change its sprite, both in online and offline.

Gingerbeard5773:
## Status

- **READY**:

## Description

Mechanisms have been left behind with every update, causing various bugs.

* Fixes mechanism states & appearance not synchronizing to joining players on dedi.
* Fixes mechanism effects not appearing on client (lever & button).
* Fixes a problem where mechanism code was executed twice on localhost.
* Fixes a bug where players could not input items into the magazine due to their resolution.
* Improves the magazine so that it is easier to use.
* Fixed blood not appearing on the spiker.
* Improves the obstructor by giving it damage frames.
* Deletes two legacy files that become unnecessary with this PR. Don't scream at me for it because I genuinely believe it wont cause any problems for modders.

resolves #1274
resolves #1874
resolves #1875
resolves #2004
resolves #2152
resolves #2201
resolves #2200

Other PRs that are also closed by default due to this PR
#1879
#2007
#2202

Credits to @mugg91 , as some parts of this PR are from him.


Vam-Jam (changes requested):
Dispensers can not be placed in sandbox or by the map loader. Results in TDM maps breaking

Gingerbeard5773:
> Dispensers can not be placed in sandbox or by the map loader. Results in TDM maps breaking
resolved

Entities/Structures/Components/Load/Dispenser/Dispenser.as
```diff
@@ -23,87 +22,22 @@ class Dispenser : Component
...
-		1.0f,                               // scale
-		3,                                  // ticks per frame
-		0.0f,                               // gravity
-		false);                             // self lit
-
 		if (isServer())
 		{
+			CMap@ map = getMap();
 			CBlob@[] blobs;
 			getMap().getBlobsAtPosition((offset * -1) * 8 + position, @blobs);
```
Vam-Jam, line 29: use map instead of getMap() here

Vam-Jam (dismissed):
I played through a few TDM rounds, maps seem to be working correctly now, things worked when testing in sandbox. May need more testing but LGTM

Vam-Jam (approved):

asumagic (approved):
Also tested a bit manually in sandbox on dedi + tested TDM maps on dedi, saw nothing abnormal.

mugg91:
Spiker seems to behave normally now, in both offline and online.
The only thing I found is that Lever will already by activated when placed; previously it was in deactivated state when placed.

mugg91:
Just for future reference, this PR made it so that Obstructor and Sensor can't be built hanging from a row of blocks anymore.
It was more like a bug than a feature, so it is fine.

Issue #2209, opened 2024-11-30
Vehicle ammo count is not shown in offline


mugg91:
## Description

When attached in the gunner position of a vehicle (Catapult, mounted Bow), you can see the remaining ammo in online.
But in offline singleplayer, it will not be shown.

Originally mentioned and fixed in https://github.com/transhumandesign/kag-base/pull/1251

mugg91:
This seems to actually be a feature where ammo in singleplayer is infinite.
**VehicleCommon.as**
```
a.infinite_ammo = getRules().hasTag("singleplayer");
```
It would be nice to show an infinite symbol but other than that I guess there is no need to change this.

PR #2222, opened 2024-12-06
[changed] Infinite ammo count is shown in singleplayer

mugg91:
## Status

- **READY**

## Description

Fixes https://github.com/transhumandesign/kag-base/issues/2209

In offline singleplayer, when using a mounted bow or a catapult, you get to have infinite ammo.
But there is no ammo count drawn to indicate this.
This PR makes changes to `VehicleGUI.as` to show an infinite symbol.

I made some other changes to simplify the code a bit in `drawAmmoCount()` but let me know if it is good or not.

|   | Before | After |
| ----------- | -------------| -------------| 
| Bow in Save The Princess | BowBefore | BowAfter |
| Catapult in other gamemodes | CatapultBefore | CatapultAfter |


Vam-Jam (approved):
LGTM

Issue #2223, opened 2024-12-07
Make Mine prime when loaded in your Catapult


mugg91:
## Current in-game behaviour.

I think it would be more fun and make sense to have your Mine become primed in your catapult and then shoot that at the enemy team.

Issue #2224, opened 2024-12-08
Catapult doesn't shoot items a proper distance in offline


mugg91:
## Description
In offline, Catapult shoots items only a very short distance, compared to online.

PR #2225, opened 2024-12-08
[changed] Mine can prime in Catapult's mag

mugg91:
## Status

## Description
```
[changed] Mine can prime while loaded to a Catapult
```

Fixes https://github.com/transhumandesign/kag-base/issues/2223

This PR changes `Mine.as` so Mine can prime while attached to a Catapult's mag.

This allows that Mine can be shoot onto the battlefield while already primed which I think would be a fun, QoL addition to the game. Previously you could only shoot it in its unprimed state so the enemy could not be hit by it and steal it. 

Since this PR makes Mine to be collidable while attached, enemy can be killed when touching the primed mine while it is in the catapult. And it can be shot by enemy arrows which I think is a fair trade-off/risk.

`this.Untag(MINE_PRIMING);` is removed from `onAttach()` which looks to be required for this change to work. However, every aspect of Mine that I tested works as intended. Picking up Mine will not prime it in your hand and the timer will reset, just as before.

There were two checks for `this.isAttached()` in `onTick()` so I removed the 2nd one.

I tested in offline and online, works as intended.

Please note there is a bug that Catapult can't shoot items far away in offline. ( https://github.com/transhumandesign/kag-base/issues/2224 ) 
That will be fixed in a different PR.

### DIFF
--- a/Entities/Items/Explosives/Mine/Mine.as
+++ b/Entities/Items/Explosives/Mine/Mine.as
@@ -17,7 +17,13 @@ enum State
 
 void onInit(CBlob@ this)
 {
-	this.getShape().getVars().waterDragScale = 16.0f;
+	CShape@ shape = this.getShape();
+	
+	if (shape !is null)
+	{
+		shape.getVars().waterDragScale = 16.0f;
+		shape.getConsts().collideWhenAttached = true;
+	}
 
 	this.set_f32("explosive_radius", 32.0f);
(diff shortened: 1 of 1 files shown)


mugg91:
https://github.com/user-attachments/assets/1681d8bd-2b38-464e-903a-47eef5603d4a

Played on dedicated server

Vam-Jam (commented):
Works in offline and online, marking for discussion

PR #2226, opened 2024-12-08
[fixed] Catapult couldn't fire properly in offline

mugg91:
## Status

## Description

Fixes https://github.com/transhumandesign/kag-base/issues/2224

Catapult has a bug where it doesn't shoot blobs properly in offline.
It looks like this happens because the logic for the firing ( `v.onFire()` and `v.SetFireDelay()` ) is executed twice in offline.
This PR fixes that by adding a check for `!isServer()` in `Catapult.as`.

I tried adding the check to `Vehicle.as` at line 194
`else if (cmd == this.getCommandID("fire client") && isClient())`
--> `else if (cmd == this.getCommandID("fire client") && isClient() && !isServer())`
but it didn't fix the issue so I'm going ahead with the fix mentioned above.

After this PR, Catapult fires blobs properly and plays the firing sound in both offline and online.

### DIFF
--- a/Entities/Vehicles/Catapult/Catapult.as
+++ b/Entities/Vehicles/Catapult/Catapult.as
@@ -276,15 +276,19 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
 	}
 	else if (cmd == this.getCommandID("fire mag blob client") && isClient())
 	{
-		u16 id;
-		if (!params.saferead_u16(id)) return;
+		if (!isServer())
+		{
+			u16 id;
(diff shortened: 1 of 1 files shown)


mugg91:
## Before

https://github.com/user-attachments/assets/51d398ed-ee49-436e-8a67-f75b047f0245

## After

https://github.com/user-attachments/assets/2f6fe5a5-9d65-4f3c-915f-3372180fda54

(Played in offline)

Vam-Jam (approved):
Works in offline & online.

asumagic (approved):

PR #2232, opened 2024-12-11
[new] Holiday automagically changes

Vam-Jam:
Holiday is now automagically handled on start up & mid game. This is currently only enabled on staging, may try to back port to vanilla 

This PR adds the following:
- Christmas overlay
- Halloween overlay
- Holiday.as now includes `string GetCurrentHoliday()` which is read by the engine on startup.

Known issues:
- If holiday changes on map load, it takes another map load for it to work 
- Vanilla client uses Overlay files, will do an engine side PR if I don't backport to vanilla


Vam-Jam:
Taking out of draft as its ready for testing

Vam-Jam:
Made it so we reload the current map on holiday change, hacky fix for making assets reload properly until we get a better engine side way of reloading assets on the fly.

PR #2234, opened 2024-12-11
[added] Archer grapple collision sound

mugg91:
## Status

## Description

This PR is a redo of the grapple sound part of this PR https://github.com/transhumandesign/kag-base/pull/1880

This adds a sound for when the Archer's grapple collides with tiles and trees.

Tested, works the same in offline and online.

### DIFF
--- a/Entities/Characters/Archer/ArcherCommon.as
+++ b/Entities/Characters/Archer/ArcherCommon.as
@@ -121,6 +121,7 @@ bool ReceiveArrowState(CBlob@ this, CBitStream@ params)
 }
 
 const string grapple_sync_cmd = "grapple sync";
+const string grapple_sound_cmd = "grapple sound";
 
 void SyncGrapple(CBlob@ this)
 {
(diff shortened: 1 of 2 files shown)


mugg91:
https://github.com/user-attachments/assets/3315ef32-2f97-4008-8ab9-19bf9eea9029

I might implement sound when colliding with built blobs, but I don't think it's a problem for now.

If requested, I can also try to implement collision sounds depending on what we are colliding with (wood, dirt or stone), similar to https://github.com/transhumandesign/kag-base/pull/2175

mugg91:
> It also sounds too early cause it takes a bit for the force to accelerate you. I can only imagine people spamming it but I guess it's probably not worse than other sounds you can spam. I'd minimally change the sound used and make it quieter

It's just a collision sound, for the tip of the grapple. It's not a sound for when you move forward with the grapple.
About the spamming, I can change it so the sound only happens for your own grapple.
I tried a few of the existing sounds and found `hit_wood.ogg` fits best. It already uses only 0.5f volume.

PR #2235, opened 2024-12-13
[fixed] "on fire" state wasn't copied when changing class (new)

### LINKED ISSUE #1300: Changing class should copy "on fire" state

mugg91:
## Status

## Description

Fixes https://github.com/transhumandesign/kag-base/issues/1300

This is the same as https://github.com/transhumandesign/kag-base/pull/1905 except for some minor changes.
* A test print was left in, commented out, in `IsFlammable.as`
* removed or added some empty lines etc.

I redid this PR because solving a file conflict in the old PR with `StandardRespawnCommand.as` caused some confusing file change diffs.

Tested in offline and online, works.


mugg91:
The string `burn_hitter` doesn't really serve any purpose. If we remove it, we could save two instances of `#include "Hitters.as"` - in `StandardRespawnCommand.as` and `FireCommon.as`, cutting down on loading time.

In `IsFlammable.as`, this line
`blob.server_Hit(this, pos, Vec2f(0, 0), 0.25, this.get_u8(burn_hitter), true);` could change to `blob.server_Hit(this, pos, Vec2f(0, 0), 0.25, Hitters::burn, true);`

PR #2236, opened 2024-12-13
[fix] Map loader now spawns bombs

Vam-Jam:
TDM will now spawn bombs when a map uses the bombs colour. It no longer spawns them every 2000 ticks, instead it just does it once.


Vam-Jam:
Removed SpawnBombs from tdm, map loader now just spawns mat_bombs regardless of gamemode.

asumagic (approved):

PR #2243, opened 2024-12-15
[changed] Better blob spawn position logic

### LINKED ISSUE #2089: Flower and Grain Plant doesn't grow when spawned in Sandbox
**Description**

Flower and Grain Plant doesn't grow when spawned in Sandbox

mugg91:
## Status

## Description
```
[fixed] Static blobs won't spawn in the air or into the ground
[fixed] Flowers and grain plants will be spawned on the ground, allowing them to grow
```
Fixes https://github.com/transhumandesign/kag-base/issues/2089

This PR changes the blob spawn chat command.
Instead of spawning at position `pos + Vec2f(0, -5)`, it will spawn at `Vec2f_zero` at first.
After the blob is initialized, the blob's height can be fetched and blob's position will be set to `pos + Vec2f(0, getMap().tilesize) - Vec2f(0, height/2)`.
This will place them exactly on the ground the player is standing on, therefore fixing problems with blobs spawning one tile in the air (dummy, flowers, grain_plant) or into the ground (outpost, war_base).

I found that War_base and TDM_Ruins would spawn in position (0, 0) and I determined the cause is because they set their blob static in their init function.
Apparently it isn't possible to set the blob's position when the blob is static (at least for the tick it is set static? May need looking into).
To fix this, I removed `this.getShape().SetStatic(true);` for these blobs and added `AlignToTiles.as;` to their blob scripts list.
(`AlignToTiles.as` makes the blob static a tick later and is used by every other building.)

This change also makes it so spawning a Dinghy or Longboat won't somehow place you underneath the boat, getting you stuck.

Tested in offline and online, with several buildings and blobs, works as intended.


mugg91:
## Before

https://github.com/user-attachments/assets/5a175571-244b-4d09-a309-1c3052e28e9c

## After

https://github.com/user-attachments/assets/1e29d451-c2fe-4feb-863c-b5f6b677058b

mugg91:
I fixed the file conflict, but it could be better. We only need to find out what's the blob's height the first time, after that we could set height directly with `server_CreateBlob()`.

mugg91:
New commit submitted that will fetch height only the first time. Tested briefly in offline and it works (with 1 blob or 10 blobs).

Issue #2197, opened 2024-11-28
Dropped CTF flag orange arrow should point towards the origin location


mugg91:
## Description

A dropped CTF flag will have an orange arrow blinking and pointing to the left.
It should point to the side it was stolen from.

PR #2252, opened 2024-12-16
[fixed] CTF Flag return arrow will point towards its flag base

mugg91:
## Status

## Description
Fixes https://github.com/transhumandesign/kag-base/issues/2197

This PR makes the CTF Flag's return indicator point into the direction it was stolen from, rather than always pointing to the left.

### DIFF
--- a/Entities/Special/CTF/CTF_Flag.as
+++ b/Entities/Special/CTF/CTF_Flag.as
@@ -110,7 +110,7 @@ void onRender(CSprite@ this)
 		return;
 
 	CBlob@ blob = this.getBlob();
-	if (blob.isAttached())
+	if (blob is null || blob.isAttached())
 	{
 		//todo: render "go to here" gui
 		return;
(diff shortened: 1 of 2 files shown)


mugg91:
## Before

https://github.com/user-attachments/assets/9deba759-0403-43c4-b332-500ee8fdc31a

## After

https://github.com/user-attachments/assets/f8728f6a-7acb-446f-b9ac-54e48b49360f

mugg91:
Because it was declared as a "todo" in the script, I wrote code for showing an arrow pointing towards your nearest friendly flag base, when you are holding an enemy flag. The arrow works the same as the tunnel buttons pointing to where the tunnel is (8 possible directions). 

I would leave it out because I think it might create too much clutter on the screen. And finding your way back to your base should be part of gameplay. Just saying it's there, if there is need in the future. Or should I make a commit with the code commented out?

https://github.com/user-attachments/assets/11d4b05c-6f7a-4336-90a6-90daf2eba7a3

flagdirections

Entities/Special/CTF/CTF_Flag.as
```diff
@@ -152,14 +152,14 @@ void onRender(CSprite@ this)
 
 	if (getGameTime() % 15 > 10)
 	{
+		Vec2f base_position = blob.get_Vec2f("flag base position");
+		Vec2f flag_position = blob.getPosition();
```
Vam-Jam, line 156: use getInterpolatedPosition when rendering

mugg91: These positionings are not used for rendering, just for determining if we are on the left or right of the origin base.

asumagic:
To me, the `<`/`<<<` arrow on the progress bar is only intended to show the rate of returning.  
It doesn't really make sense to me for it to point towards where it should be returned, instead of pointing towards where the progress bar is depleting (i.e. left).

A similar convention exists for other things (sword and shield for vehicle capture) although it's not actually using arrows.

Since this doesn't really add much info I'm not sure it's worth adding.

PR #2256, opened 2024-12-18
[fixed] lever turning on immediately when placed

### LINKED ISSUE #2255: Lever will be activated when placed
## Description

After https://github.com/transhumandesign/kag-base/pull/2219 , Lever will be activated initially. It should be placed in deactivated state instead. Affects some TDM maps.

Gingerbeard5773:
resolves #2255


mugg91:
It works.

asumagic (approved):

PR #2258, opened 2024-12-19
[changed/fixed] Decay spam hits + [removed] DecayIfLeftAlone

### LINKED ISSUE #860: Fix DecayIfSpammed
DecayIfSpammed script is a good idea but is flawed in execution.

* The spam limit should be configurable. Currently > 4.
* Should not count attached blobs.
* DecayIfSpammed damages more than one of the duplicate blobs.
* The script probably shouldn't sample by doing (i*997)%length.

In order for DecayIfSpammed to only damage one we could probably tag one of them when it damages itself, and early out in the other loops if we find one with that tag. Then remove that tag before the next loop.

### LINKED ISSUE #1751: TDM minor issues

mugg91:
## Status

- **READY**

## Description
```
[fixed] DecayIfSpammed won't damage more blobs than necessary
[dev] spam limit can be set to blobs which will be checked for in DecayIfSpammed
[changed] Invincible blobs won't decay
[changed] Attached blobs won't decay
[changed] DecayIfFlipped and DecayIfSpammed will check for disallowDecaying() first
[removed] Script "DecayIfLeftAlone.as"
```

Fixes https://github.com/transhumandesign/kag-base/issues/1751
Fixes https://github.com/transhumandesign/kag-base/issues/860

`dissalowDecaying()` renamed to `disallowDecaying()`.

Decay scripts that use `disallowDecaying()` won't decay blobs if they are invincible or attached.

`disallowDecaying` added to `DecayIfFlipped.as` and `DecayIfSpammed.as`.

## DecayIfLeftAlone removal

Script `DecayIfLeftAlone.as` and all references to it removed as it is unused and I don't see it getting brought back in the future. It looks unpopular to me.

## DecayIfSpammed change

`DecayIfSpammed.as` was rewritten.
It checks if a spam limit has been set. Allows for custom spam limit that can be set to blobs via `set_u8("spam limit", number)`. Otherwise defaults to 5.

Blobs are kept track of via network IDs.
<s>Only the blob with the lowest net id will be hit, fixing the problem that several blobs were hit at the same time. More blobs will be hit if there are multiples of the spam limit:
* 4 buckets, with the limit being 4 - Nothing happens.
* 5~8 buckets, with the limit being 4 - The first bucket will be hit.
* 9~12 buckets, with the limit being 4 - The first two buckets will be hit.
* 13~16 buckets, with the limit being 4 - The first three buckets will be hit. etc.</s>
Oldest 5 (or whatever the spam limit is) buckets will stay and every bucket past that will decay.

This worked very well in testing.

## DecayQuantity change

`onAttach()` and `onThisAddToInventory()` are removed from `DecayQuantity.as` because the only thing they do is set `this.getCurrentScript().tickFrequency = (getRules().hasTag("quick decay") ? QUICK_FREQUENCY : FREQUENCY);` which is already done in `onInit()` so it looks like these hooks don't really fulfill a purpose.
If I'm mistaken and there is a purpose, please let me know and I can revert the change.


mugg91:
## Before

https://github.com/user-attachments/assets/91ffa653-f174-4720-bf7c-de6394be2b14

## After

https://github.com/user-attachments/assets/48e16b46-6d97-49a0-94cc-687f64a4e9cf



I noticed some things.
* It looks like the limit in the current live build is actually 5, with the 6th bucket triggering the decay. I will eventually change it accordingly in this PR.
* I think that decaying everything except the oldest 5 buckets would be better. Only decaying multiples of the limit is not a good counter measure against spamming.

Therefore setting this to __"In Development"__ for now.

mugg91:
Committed changes explained above. 
Ready for review now.

mugg91:
Updated to latest master build.

Issue #2254, opened 2024-12-17
Wall climb status should be reset when descending


mugg91:
## Current in-game behaviour.

Archer should gain his ability to wall climb back when descending past the "can still wall climb up" limit.

https://github.com/user-attachments/assets/eed8a793-c03a-4b8f-87c3-6557a0f08e95

Related issue https://github.com/transhumandesign/kag-base/issues/2118

Issue #2118, opened 2024-09-05
trampoline should set 'blob.isOnGround()' to true on tick when used


NoahTheLegend:
**Current in-game behaviour.**

some of in-game features like wallclimb force recharge won't work if you bounce off a trampo (you need either to step on ground, be in water/on ladder, or wall jump from opposite side), making it impossible to climb staircasely in sequence wallclimb>trampo jump>(failed) wallclimb.

**Proposed Change**

title

**Additional context**

make sure it won't break hacks in other code

https://github.com/user-attachments/assets/d982c42d-8d15-4f9b-975f-0172ffd18adf

PR #2259, opened 2024-12-20
[changed] wall climb state is reset if descending past slide speed or when bouncing on a trampoline

mugg91:
## Status

## Description

Fixes https://github.com/transhumandesign/kag-base/issues/2254
Fixes https://github.com/transhumandesign/kag-base/issues/2118

This PR resets your character's wall climb state when descending down beyond the slide-speed threshold, reducing annoying situations where the game prevents you from wall climbing if you have descended but not touched the ground yet, and then have made your way back up via grapple or trampoline.

This was done by splitting up the if-else codeblocks in `RunnerMovement.as` into:
```
	if (!blob.isOnCeiling() && !isknocked && !blob.isOnLadder()) 
	{
		const f32 slidespeed = 2.45f;
		bool key_pressed = (up || left || right || down);

		if (key_pressed)
		{

		( . . . )

		}
		else if (vel.y > slidespeed)
		{
			moveVars.wallrun_count = 0;
			moveVars.walljumped_side = Walljump::NONE;
		}
	}
```

Since this alone didn't fix the last situation in the video in issue https://github.com/transhumandesign/kag-base/issues/2118 , I changed `TrampolineLogic.as` to reset the player's wallclimb state when bounced.

Tested in offline, works as intended. Tested situations are:
* Archer descending but not touching the ground, then grappling up on the same wall again.
* Player wallclimbing and walljumping, then falling down and bouncing up by trampoline, then wallclimbing the same wall again.
* Knight walljumping off a left wall, then bouncing up, then trying to wall climb another left wall before touching the ground.
* 
After this PR, you are allowed to wall climb again in these situations.

### DIFF
--- a/Entities/Characters/MovementScripts/RunnerMovement.as
+++ b/Entities/Characters/MovementScripts/RunnerMovement.as
@@ -242,65 +242,52 @@ void onTick(CMovement@ this)
 			moveVars.walljumped_side = Walljump::RIGHT;
 	}
 
-	if (!blob.isOnCeiling() && !isknocked &&
-	        !blob.isOnLadder() && (up || left || right || down))  //key pressed
+	if (!blob.isOnCeiling() && !isknocked && !blob.isOnLadder()) 
 	{
-		//check solid tiles
-		const f32 ts = map.tilesize;
-		const f32 y_ts = ts * 0.2f;
-		const f32 x_ts = ts * 1.4f;
-
-		bool surface_left = map.isTileSolid(pos + Vec2f(-x_ts, y_ts - map.tilesize)) ||
-		                    map.isTileSolid(pos + Vec2f(-x_ts, y_ts));
-		if (!surface_left)
-		{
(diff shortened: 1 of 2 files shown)


azephiris:
Imo wall climbing (and other abilities) should use stamina.
In this case, that would mean using a timer to gain another jump every X.
Resetting the jump counter on particular situations seems risky and not future-proof.

mugg91:
I chose the condition for the state reset to be `vel.y > slidespeed` (slidespeed equals `2.45f`)
For speeds less than or equal to slide speed, nothing changes.
For speeds higher than slide speed, your wall climbing state is reset. But you will not be able to wall climb on the same wall again immediately because you are descending past the slide speed. (There are checks `vel.y < slidespeed` so logically there isn't a way to abuse the state reset - it only benefits you if you make your way back up through a legitimate method, via trampoline, grapple or bomb jump.

mugg91:
## Before

https://github.com/user-attachments/assets/e18084db-ae42-4204-8740-13cbfa5cae7d

## After

https://github.com/user-attachments/assets/1e645414-42bc-4141-a1c8-df6002bffc45

Issue #2106, opened 2024-07-13
Bombs aren't damaging same team wires


mugg91:
## Current in-game behaviour.

Wires that have the same color as your team aren't getting damaged by your bombs.
This is of significance in some TDM maps.
Wires should either always get destroyed by bombs or not, regardless the color.

Gingerbeard5773:
Since this issue is only relevant to TDM, the appropriate fix would be to make the components/wires spectator team

mugg91:
> Since this issue is only relevant to TDM, the appropriate fix would be to make the components/wires spectator team

Looks like this is going to be addressed in https://github.com/transhumandesign/kag-base/pull/2364

PR #2260, opened 2024-12-21
[fixed] Bombs couldn't damage same team wires (TDM)

mugg91:
## Status

## Description

Fixes https://github.com/transhumandesign/kag-base/issues/2106

Components are given the tag "is component" and TDM gamemode checks for the tag in `onBlobCreated()` to change components' team to neutral team (255). 

I made it so the sprite is reloaded in the previous team color, so we could use any team color we want. Some existing TDM maps use blue wires and I didn't want to change that to grey.

By setting team to 255, wires and components are now always damaged by bombs.
Previously, bombs wouldn't hurt wires of the same team.

I think this is much better than having to edit any map png files directly.

Tested in offline, works.

### DIFF
--- a/Rules/TDM/Scripts/TDM.as
+++ b/Rules/TDM/Scripts/TDM.as
@@ -831,3 +831,12 @@ void onInit(CRules@ this)
 {
 	Reset(this);
 }
+
+void onBlobCreated(CRules@ this, CBlob@ blob)
+{
+	if (blob is null)
+		return;
+
+	if (blob.hasTag("is component"))
+		blob.server_setTeamNum(255);
+}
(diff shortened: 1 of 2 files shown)


Rules/TDM/Scripts/TDM.as
```diff
@@ -831,3 +831,21 @@ void onInit(CRules@ this)
...
+
+	if (blob.hasTag("is component"))
+	{	
+		u8 team = blob.getTeamNum();
+		blob.server_setTeamNum(255);
+
+		CSprite@ s = blob.getSprite();
+		if (s !is null)
+		{
+			spr.ReloadSprites(team, 0);
```
asumagic, line 848:
```suggestion
			s.ReloadSprites(team, 0);
```

asumagic (changes requested):
This approach doesn't work for changing the component color online, whether on joining the server or loading a new map. I don't know if this specific approach can work due to that.

Issue #2262, opened 2024-12-22
Quarry doesn't show wood materials layer in online


mugg91:
## Description

When adding wood to a quarry, the wood layer is shown in offline, but not in online.

PR #2263, opened 2024-12-22
[fixed] Wood layer was missing when loading wood to a quarry in online

mugg91:
## Status

## Description

Fixes https://github.com/transhumandesign/kag-base/issues/2262

The command "add fuel" was only run on the server, so the spritelayer wasn't being set.

I added "add fuel client" command that is run client-side and synced `fuel_prop` in order to update the spritelayer.

In offline and online, it works correctly now. 
* Loading wood correctly set the spritelayer, the higher the amount, the more is shown visibly. 
* The wood will visibly decrease over time.


Vam-Jam (approved):
Worked offline, on dedi server & when rejoining.

asumagic (approved):

Issue #2207, opened 2024-11-30
Mounted Bow should be possible to detach from vehicles


mugg91:
## Description

https://github.com/user-attachments/assets/82bdca9f-2b41-4502-bed5-a74cc48c1b2b

Note: The Mounted Bow which is attached to Warboat should not be possible to detach.
Only allow detaching Mounted Bow which we attach ourselves.

Originally addressed in https://github.com/transhumandesign/kag-base/pull/1251

PR #2267, opened 2024-12-24
[changed] Mounted Bow can be detached from other vehicles

mugg91:
## Status

## Description

Fixes https://github.com/transhumandesign/kag-base/issues/2207

You are allowed to detach mounted bow that is attached to other vehicles (longboat, warboat, airship, bomber).
The button offset was misplaced for longboat and bomber and I fixed it by dividing the `ap.offset` by 2.
The mounted bow on warboat still cannot be detached after this change.

Tested in offline.

### DIFF
--- a/Entities/Vehicles/Common/VehicleAttachment.as
+++ b/Entities/Vehicles/Common/VehicleAttachment.as
@@ -25,10 +25,12 @@ void GetButtonsFor(CBlob@ this, CBlob@ caller)
 				CBlob@ occBlob = ap.getOccupied();
 				if (occBlob !is null) //detach button
 				{
-					if (this.isOnGround() && occBlob.getName() != "mounted_bow")	  // HACK:
+					if (this.isOnGround() || occBlob.getName() == "mounted_bow")	  // HACK:
 					{
+						Vec2f button_offset = ap.offset/2;
+						button_offset.RotateByDegrees(occBlob.getAngleDegrees());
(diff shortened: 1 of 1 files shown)


Entities/Vehicles/Common/VehicleAttachment.as
```diff
@@ -25,10 +25,12 @@ void GetButtonsFor(CBlob@ this, CBlob@ caller)
 				CBlob@ occBlob = ap.getOccupied();
 				if (occBlob !is null) //detach button
 				{
-					if (this.isOnGround() && occBlob.getName() != "mounted_bow")	  // HACK:
+					if (this.isOnGround() || occBlob.getName() == "mounted_bow")	  // HACK:
```
Vam-Jam, line 28: Pr works fine, but IMO just remove the getName check, does not make sense to have the bow as the only thing you can drop while flying/moving.

Entities/Vehicles/Common/VehicleAttachment.as
```diff
@@ -60,7 +62,7 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
 						// range check
 						if (this.getDistanceTo(b) > 64.0f) return;
 
-						if (this.isOnGround() && occBlob.getName() != "mounted_bow")
+						if (this.isOnGround() || occBlob.getName() == "mounted_bow")
```
Vam-Jam, line 65: (same here)

Vam-Jam (changes requested):

PR #2271, opened 2024-12-25
[fixed] Blood on Saw/Spikes/Spiker will be toggled properly on changing g_kidssafe

mugg91:
## Status

## Description
```
[fixed] Spikes, Spiker and Saw will properly remember if they are bloody or not and will instantly update their spritelayer when changing g_kidssafe from the menu
[changed] Spikes don't continuously update their animation frame in onTick() via calling onHealthChange()
[added] Water splash from bucket, water bomb or water arrow will remove the blood 
[added] Spiker will show blood on its tip even in closed state
```
Fixes https://github.com/transhumandesign/kag-base/issues/1811

This PR is a revive of https://github.com/transhumandesign/kag-base/pull/1854 with changes.
* Fixed file conflicts and did some careful changes in order to support the bloody toggle and the clean sparkle effect.
* The "no flesh gibs when kidssafe is on" part has been split into a separate PR [].
* In `SplashWater.as`, the check for the invincible tag is moved further down.
* `CleanSparkles.as` won't produce sparkles if the blob has become bloody again.

Now the bloody toggling works just like it used to in the old PR.


mugg91:
I'm having trouble fixing Saw in this PR.

This change from April 27 https://github.com/transhumandesign/kag-base/commit/377a47b0f5c023a287b4c4e229b09ac8f183e0b3 makes it so Saw only tags "bloody" on server and syncs it.
This works the first time, but when using water splash to remove the "bloody" tag and then blending something again, the tag is not synced to the client anymore. What did the above change fix? I might have to revert it if I can't find another solution.

mugg91:
I tried for 2 hours and can't fix the spiker issue (with the spritelayer). It's in the older PR and vanilla, too.

mugg91:
I tried for another 2 hours. Found that `onSetStatic()` is run when rejoining, which is kind of stupid but it explains why the attachment point offset gets set to the "hidden" positioning. By checking if there already is a spike attached to spiker, I can set the correct offset:
```
	AttachmentPoint@ mechanism = this.getAttachments().getAttachmentPointByName("MECHANISM");
	CBlob@ occ = mechanism.getOccupied();
	if (occ is null)
		mechanism.offset = Vec2f(0, -3); // no spike attached yet, set to "hidden" offset
	else
		mechanism.offset = Vec2f(0, occ.get_u8("state") == 0 ? -3 : -7); // spike is attached, set offset depending on current state
```

Did some other miscellaneous changes. 
* Saving the spiker netid to spike is no longer needed. 
* I removed the background spritelayer on spiker and adjusted the possible offset positionings for Spike. 
* I removed that spike would get set static - it's already attached so it's not like it's going to move. 
* Spike now correctly checks if damage is above 0 and we aren't in water, before applying the "bloody" tag, just like the old PR.

I tested with Spiker in various settings and it looks all correct now, even after rejoining. (Phew...)

mugg91:
## Before

https://github.com/user-attachments/assets/3dd71103-9c9f-4be9-8189-fa306ce20f8d

## After

https://github.com/user-attachments/assets/beb787f3-af1d-4ab9-88e9-f49bd5021797

mehwaffle10:
I'm not a fan of the sound or particle effects for this

mugg91:
> I'm not a fan of the sound or particle effects for this

It only happens in gameplay if a water splash is done near a bloody thing. Should it be different or should it not be present in the game? I'm just aiming to add more life to the game. Let me know how I could change it.

Vam-Jam:
The kid_safe part is fine, but the water cleanup / particle changes should be a separate PR. IMO I would get rid of the clean up sound and tweak the wash particle to have a shorter duration & smaller in size

Issue #1309, opened 2022-07-06
Bugs when spawning a new music


mugg91:
## Description

When spawning a new music via chat command there are problems.
These problems happen in offline and online and are not fixed by rejoining a server.

1) Spawning a challenge_music will cause music to play for 2 seconds and then stops suddenly, repeatedly.
This happens both offline and online and isn't fixed by rejoining the server. The music stays bugged forever.
2) Spawning a ctf_music on top of pre-existing other music will cause the ctf_music to play on top of it. "KAGWorld1-14outro" will be used even though the match timer is running.
This happens only in offline.
In online, war_music, music and ctf_music will correctly replace each other.
3) You would think tdm_music would work but it is named simply "music".

Tested in offline and online Sandbox. Other game modes may differ.

## Reproduction

Go to offline Sandbox.
!ctf_music
Notice two musics playing on top of each other.

---

Go to online Sandbox EU server or offline Sandbox.
!challenge_music
Notice music getting bugged.

---

Go to online Sandbox EU server or offline Sandbox.
!tdm_music
Notice chat saying it doesn't exist.

## Expected Behavior

Spawning a ctf_music should behave the same in offline and online, correctly replacing the existing music and not playing the outro when the match timer is still running.

Spawning a challenge_music should not wreck the music but replace the existing music and keep playing as normal.

TDMMusic should be configured as "tdm_music" instead of "music" for consistency reasons.

mugg91:
Instead of allowing spawning musics on top of each other it could be a possible solution to "not allow it":
- Don't spawn the music if there is already one playing, or 
- put all types of musics on the blacklist in chat commands.

PR #2272, opened 2024-12-26
[fixed] Bugs when spawning a music

mugg91:
## Status

## Description

Fixes https://github.com/transhumandesign/kag-base/issues/1309

After this PR, you cannot spawn a `ctf_music`, `music`, `war_music` or `challenge_music` if one of these already exists.

Fixes bugs that occur when multiple music blobs exist (music getting cut off over and over).

Tested in offline and online, found no problems. 
When a CTF match ends and the new map loads, music will correctly start playing again.

## Alternative Solutions

This Pr might not be a good solution but I only aimed to fix the issue. There could be other possible solutions.

a) Replacing the existing music
It might be possible to write logic for replacing the existing music but it requires more time and research.

b) Blacklisting
Alternative to this PR would be to just put music blobs on the server's black list.

c) Moving music code to a rules script
Why are we using blobs?

### DIFF
--- a/Entities/Meta/CTFMusic.as
+++ b/Entities/Meta/CTFMusic.as
@@ -2,6 +2,8 @@
 
 #define CLIENT_ONLY
 
+#include "MusicCommon.as";
+
 const string[] classList = {"knight", "archer", "builder"};
 const string[] blobList = {"knight", "archer", "builder", "ballista", "tunnel", "keg"};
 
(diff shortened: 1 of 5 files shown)


Vam-Jam:
I'd rather blacklist spawning these blobs in sandbox servers (ignoring localhost). I forgot the music was even a blob to begin with.

mugg91:
@Vam-Jam 
If you add the music blobs to the blacklist, then this PR can be dismissed.

PR #2275, opened 2024-12-28
[fixed] spiker Z layer for vanilla

### LINKED ISSUE #2274: Spike on Spiker has the wrong positioning when rejoining the server
## Description

So I think this issue is noteworthy that it warrants an entry.

This issue exists in 
the current live build, 
in the current dev build with the mechanism PR by @Gingerbeard5773 https://github.com/transhumandesign/kag-base/pull/2219
and in the obsolete bloody blobs PR by me https://github.com/transhumandesign/kag-base/pull/1854

Because `onSetStatic()` in `Spiker.as` gets run after rejoining, the attachment point offset on spiker gets set to `Vec2f(0,-3);`. 
This code makes sense when first placing the spiker but when rejoining, it could set the offset to this Vec2f even when there already is a spike attached, so it looks incorrect.

It can be fixed by first checking if there already is a spike attached and setting the offset accordingly.

Gingerbeard5773:
resolves #2274

### DIFF
--- a/Entities/Structures/Components/Load/Spiker/Spiker.as
+++ b/Entities/Structures/Components/Load/Spiker/Spiker.as
@@ -135,7 +135,7 @@ void onSetStatic(CBlob@ this, const bool isStatic)
 	CSpriteLayer@ layer = sprite.addSpriteLayer("background", "Spiker.png", 8, 16);
 	layer.addAnimation("default", 0, false);
 	layer.animation.AddFrame(4);
-	layer.SetRelativeZ(-10);
+	layer.SetRelativeZ(-15);
 	layer.SetFacingLeft(false);
 }
 


Gingerbeard5773:
@mugg91 you can change up your pr to incorporate this change, i will close this pr in response

mugg91:
Omg, it works.
Ok, I will revert all that stuff I did in https://github.com/transhumandesign/kag-base/pull/2271 but it might take me a day or two.

mugg91:
> Omg, it works. Ok, I will revert all that stuff I did in #2271 but it might take me a day or two.

I tried this solution in https://github.com/transhumandesign/kag-base/pull/2271 but I noticed a problem.
When spiker is ejected and bloody, then rejoining, then retracting the spiker, its tip wouldn't be bloody (a feature this PR added).
It's likely because when Spike is created, `onReceiveCreateData()` is called, which calls `UpdateSprite()`, which tries to update Spiker by fetching its net id from Spike. But it doesn't go through with the update, maybe because Spiker isn't created yet.

I think dealing with this problem would introduce more code to the point where I think the solution I worked out in https://github.com/transhumandesign/kag-base/pull/2271 is more satisfying. Instead of updating blood in Spike and Spiker, this PR only updates blood on Spike's sprite, period.

Vam-Jam (approved):

PR #2276, opened 2024-12-28
[modified] Pickup menu overhaul

Gingerbeard5773:
## Status

- **READY**

## Description

The pickup menu cannot be used with all blobs in KAG, making it useless in certain scenarios in game. It is also completely useless for mods that have custom items.
This PR rewrites the pickup menu so that it can be used for all blobs in base KAG, as well as mods.

Demonstration of the new menu:

https://github.com/user-attachments/assets/d986376e-755e-4956-8759-3c8af76ed7d4

* Only one selection on the menu can be displayed for any type of blob, eliminating bloat that could potentially occur on the menu.

### DIFF
--- a/Entities/Common/Controls/StandardPickup.as
+++ b/Entities/Common/Controls/StandardPickup.as
@@ -24,49 +24,6 @@ void onInit(CBlob@ this)
 	if (menu.entries.length == 0)
 	{
 		menu.option_notice = "Pickup";
-
-		// knight stuff
-		menu.add_entry(PickupWheelMenuEntry("Keg", "$keg$", "keg"));
-
-		const PickupWheelOption[] bomb_options = {PickupWheelOption("bomb", 1), PickupWheelOption("mat_bombs", 0)};
-		menu.add_entry(PickupWheelMenuEntry("Bomb", "$mat_bombs$", bomb_options, Vec2f(0, -8.0f)));
-
-		const PickupWheelOption[] waterbomb_options = {PickupWheelOption("waterbomb", 1), PickupWheelOption("mat_waterbombs", 0)};
-		menu.add_entry(PickupWheelMenuEntry("Water Bomb", "$mat_waterbombs$", waterbomb_options, Vec2f(0, -6.0f)));
-
-		menu.add_entry(PickupWheelMenuEntry("Mine", "$mine$", "mine"));
-
-		// archer stuff
-		menu.add_entry(PickupWheelMenuEntry("Arrows", "$mat_arrows$", "mat_arrows", Vec2f(0, -8.0f)));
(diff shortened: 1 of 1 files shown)


TerminalHash:
picked that pr as additional mod setting for testing on Gruhsha

mugg91:
I like the existing pickup menu because I can quickly pick up seeds, which will always be in the bottom like in the screenshot.



This PR will change positionings every time so it will be difficult to use the menu quickly.

I think what would be really good is if players could customize their own pickup menu, by setting what items they want and in what positionings.

mugg91:
One other thing that might be a problem is since the selection changes based on what's in reach, items getting inside or outside of reach could possibly mess up your selection.

In this demonstration, it's impossible to pick up your desired item.

https://github.com/user-attachments/assets/287cd67f-e803-4b63-8a9d-b3b4bb8b7ad7

Gingerbeard5773:
Current state of the changes. The menu has been redone again, this time using a 'hash map' of sorts to set the items to their own special spot in the wheel. This is a great new method and appears to fix 95% of the previous problems that people have mentioned. The other 5% of the time- the hash method can have collisions, which may move items on the menu. But this doesn't seem to happen almost at all on vanilla, and isn't very jarring when it does occur either.

https://github.com/user-attachments/assets/0d3e25bd-058f-4398-a298-1a99000aa393

mugg91:
It seems better, although I would like seed to be in the same spot as before.
When there is a bomb and a mat_bombs, there will be two identical icons.

Gingerbeard5773:
> It seems better, although I would like seed to be in the same spot as before.
> Items seem to be aligned all over the place whereas previously it was sorted more or less.

This is because the blobs are put into the wheel by hash, which is what allows the menu to have all possible blobs displayed

> When there is a bomb and a mat_bombs, there will be two identical icons (same for waterbombs).

The icons are taken from the inventory_name first and then it goes for the normal name, so this means that the game already made an icon for the inventory_name without my knowledge. I'll mess around to get it looking correct. it seems like kegs have a weird icon as well

Gingerbeard5773:
PR seems to be complete. Looking for testers and reviewers. This current menu can be tested on the Zombie Fortess server and also grusha(?)
This menu is nearly the same as existing base menu aside from blob compatibility and the different layout caused by hashing.

Gingerbeard5773:
Gonna assume that this will only be merged if the menu layout is the same. So ill add in a system that allows for blobs to have a customized position in the menu.

Vam-Jam (approved):
Seems to work fine locally and on dedi, an upgrade from the current version. I don't see any obvious issues from looking at the code either.

I'm going to mark it as 'testing needed', so hopefully some other mods/servers pick it up and try it out.

asumagic:
@TerminalHash I see that was merged on Gruhsha, any feedback?

TerminalHash:
dont have issues with last changes, works perfectly for us

Issue #2238, opened 2024-12-13
this.getOverlapping(@overlapping) doesn't fetch overlapping blob


mugg91:
## Description

After a blob detaches from something, `this.getOverlapping(@overlapping)` won't fetch the thing it was attached to even if it is overlapping.

I placed this test code in `ArcherLogic.as` in `onTick()`.
It is supposed to attach archer to a quarters when the archer is below max health.
```
	//TEST CODE
	print("!attached : " + (!this.isAttached()) + " lowhp: " + this.getHealth() + " < " + this.getInitialHealth()); 
	if (!this.isAttached() && this.getHealth() < this.getInitialHealth())
	{
		CBlob@[] overlapping;
		if (this.getOverlapping(@overlapping))
		{
			print("A");
			for (uint i = 0; i < overlapping.length; i++)
			{
				print("B");
				CBlob@ b = overlapping[i];
				if (b is null || b.getName() != "quarters")
					continue;
				print("D");
				b.server_AttachTo(this, "BED");
			}
		}
	}
```
After the archer is done healing and he gets damaged again, he won't attach to the quarters again until he is moved out of the quarters (so he isn't overlapping) and then back in.

Vam-Jam:
I'm pretty sure this is fixed in an experimental Box2d v3 branch, the reason for this in v2 is because box2d BeginContact isnt called again until you leave the shape and touch it again.

This will most likely not be fixed until then.

PR #2284, opened 2025-01-02
[changed] Allow 1 seed to grow when 2 or more are overlapping

### LINKED ISSUE #2283: Two seeds overlapping is not obvious
## Current in-game behaviour.

When two seeds are on the same tile, none of them will grow.

This can easily happen in normal gameplay and not every player knows this.

## Suggestion

1) Add indicator above seeds telling that nothing will grow on that tile.

or

2) Allow one seed to grow, with the other seed staying on the ground, visible in front.

mugg91:
## Status

- **IN DEVELOPMENT**
- Still issues with several seeds growing on the same tile.

## Description

Fixes [Issue #2283](https://github.com/transhumandesign/kag-base/issues/2283)

When two or more seeds are on the ground at the same tile position, no seeds would grow.
This PR changes that so the seed with the lowest net id can grow.

`this.getOverlapping(@overlapping)` was replaced by `map.getBlobsInRadius()` due to the previous logic sometimes failing, see [PR #2238](https://github.com/transhumandesign/kag-base/issues/2238) .

<s>Tested in offline, works as intended.</s>
After further testing, I found that several seeds could grow on the same tile. Will fix soon.

### DIFF
--- a/Entities/Natural/Scripts/canGrow.as
+++ b/Entities/Natural/Scripts/canGrow.as
@@ -1,17 +1,41 @@
 //common "can a plant grow at this tile" code
 
-bool isNotTouchingOthers(CBlob@ this)
+bool isNotBlockedByOthers(CBlob@ this)
 {
-	CBlob@[] overlapping;
+	CBlob@[] blobsInRadius;
+	CMap@ map = getMap();
 
-	if (this.getOverlapping(@overlapping))
+	if (map.getBlobsInRadius(this.getPosition(), map.tilesize / 4, @blobsInRadius))
 	{
-		for (uint i = 0; i < overlapping.length; i++)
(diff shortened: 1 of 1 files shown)


mugg91:
I added some fixes:
- I'm not using `map.getBlobsInRadius()` anymore since it seems to have problems fetching the correct blobs. Reverted back to using `this.getOverlapping()`.
- Correct check if tree is overlapping `(blob.getName().find("tree") != -1)` rather than `(blob.getName().find("tree") == -1)`
- Added early-out `if (blob is null) continue;` to the loop.

Seeds growing inside trees or seeds not growing when there is nothing on the same tile should not be happening anymore. Briefly tested in offline and online.

PR #2287, opened 2025-01-02
[fixed] Invincibility is lost after setting the position when tunnel travelling, instead of before

### LINKED ISSUE #2281: Tunnel travel should remove invincibility after traveling, not before
## Description

When enemy is slashing at your tunnel and you respawn and travel, you will get hit by the slash.

It should remove the respawn invincibility after travelling, not before.

mugg91:
## Status

## Description

Fixes https://github.com/transhumandesign/kag-base/issues/2281

This changes `TunnelTravel.as`.
Previously, invincible tag was removed and then the players's position got set to the destination tunnel.
This caused an issue were you could get hit from attacks that happened at the origin tunnel.

This PR changes that so you get set to the destination tunnel's position and then another command is sent that untags invincible.
I tested with 50~250 ping on dedicated server and got killed by the saw at the destination tunnel every time.

<s>I couldn't verify the fix itself yet though. If I do, I will update here.</s>
I verified the fix, see the videos in my next post.

### DIFF
--- a/Entities/Industry/Tunnel/TunnelTravel.as
+++ b/Entities/Industry/Tunnel/TunnelTravel.as
@@ -31,6 +31,7 @@ void onInit(CBlob@ this)
 {
 	this.addCommandID("travel to");
 	this.addCommandID("travel to client");
+	this.addCommandID("travel finished");
 	this.Tag("travel tunnel");
 
 	int team = this.getTeamNum();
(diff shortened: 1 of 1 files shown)


mugg91:
I tested on dedicated server with a modified knight bot who keeps slashing at my tunnel. 

## Before (Bug)

(At 0:16)

https://github.com/user-attachments/assets/e23823b3-05cd-4632-ae71-2ebb53aa3ed2

## After (Fix)

https://github.com/user-attachments/assets/fe389328-b29f-4286-bb1f-fc06d5df86fe

Vam-Jam:
Going to mark this as testing needed, it feels like a sane change in theory but I'm wary that it might break/buff tunnels in an unexpected way

Issue #2285, opened 2025-01-02
Respawn invincibility flashing will continue even if invincibility ends prematurely


mugg91:
## Description

When respawning and travelling via tunnel (which ends the invincibility), the player's invincibility flashing will still continue, both online and offline.

PR #2288, opened 2025-01-02
[fixed] Immunity flashing will stop when losing invincibility

mugg91:
## Status

## Description

Fixes https://github.com/transhumandesign/kag-base/issues/2285

## Before

https://github.com/user-attachments/assets/0606834b-2df5-48a4-a620-654fc6835977

## After

https://github.com/user-attachments/assets/f675cd46-6116-44e6-90e2-4cafed0ab9e9

### DIFF
--- a/Entities/Characters/Scripts/UnSpawnImmunity.as
+++ b/Entities/Characters/Scripts/UnSpawnImmunity.as
@@ -12,6 +12,12 @@ void onInit(CBlob@ this)
 
 void onTick(CBlob@ this)
 {
+	if (!this.hasTag("invincible"))
+	{
+		EndInvincibility(this);
+		return;
+	}
+
 	bool immunity = false;
 
 	float time_modifier = (this.isKeyPressed(key_action1) ? 0.75f : 1.0f);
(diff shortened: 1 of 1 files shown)


Entities/Characters/Scripts/UnSpawnImmunity.as
```diff
@@ -33,11 +39,16 @@ void onTick(CBlob@ this)
...
-		this.getSprite().setRenderStyle(RenderStyle::normal);
+		EndInvincibility(this);
 	}
 }
+
+void EndInvincibility(CBlob@ this)
+{
+	this.Untag("invincible");
+	this.Tag("invincibility done");
+	this.Sync("invincibility done", true);
```
asumagic, line 50: Somewhat wasteful as this will end up syncing this property every time ontick

asumagic (changes requested):

Entities/Characters/Scripts/UnSpawnImmunity.as
mugg91: I just moved the existing code. I'm not sure we really need this tag in the first place, there might be a way to replace it with a different check.

asumagic (approved):

PR #2289, opened 2025-01-03
[added] Bush wiggle and rustling sound

mugg91:
## Status

## Description

This adds the bush wiggling and rustling from this https://github.com/transhumandesign/kag-base/pull/1393
with the following additional changes:
* Check `if (!isClient()) return;` in `onTick()`
* Check `!this.hasTag("is animated")` in `onCollision()` 

This makes Bush be animated and do a rustle sound when a player collides with it (when `v_fastrender` is false) .

https://github.com/user-attachments/assets/7dbc3416-ee3f-42e9-a6ec-488abb488f5b


mehwaffle10:
Not a huge fan of the sound or the animation tbh. Bushes are meant to be background and there are implications when hiding mines

TerminalHash:
we playing with cherry-picked https://github.com/transhumandesign/kag-base/pull/1393 on Gruhsha CTF almost one year, animated bushes and trees looks more good than static vanilla, tbh. idk what other players think, but the only complaints were about the sound of the bushes and the switch was made on that one.

mugg91:
I will try to think about a solution to deal with mines getting exposed.
Perhaps there can be a static bush and the animation in front of that or something.

Issue #2293, opened 2025-01-04
Wrong "get out" button on crate after rejoining


mugg91:
## Description

When putting a player corpse in a crate, you can take it out again like a normal item.
But when leaving and rejoining, it will show a "get out" button as if the player inside was alive.

mugg91:
Character death untags "player", and tags "dead" and "dead player".
When rejoining, the dead character will somehow have regained the "player" tag, but still also have "dead" and "dead player".
The check in `Crate.as` `getPlayerInside()` can be changed accordingly. 

There will be a normal inventory button, but it will not present an inventory for some strange reason,
even though `isInventoryAccessible()` returns true and `GetButtonsFor()` doesn't interfere in any way,
as long as the dead character is inside.

Maybe the engine side has some additional failsafes that disables the inventoryi f it thinks an alive player is inside?

mugg91:
I figured it out just after my previous post.
It looks like I was right. When the character doesn't have the "player" tag, the inventory can be accessed.

It requires changing 
```
this.tag("player")
```
to
```
	if (!this.hasTag("dead"))
		this.Tag("player");
```
in all runner blob init functions.
I will fix it soon.

PR #2295, opened 2025-01-04
[fixed] characters don't regain tag "player" on rejoin

mugg91:
## Status

## Description

Fixes https://github.com/transhumandesign/kag-base/issues/2293

For a character that dies, `"player"` gets untagged, and `"dead"` and `"dead player"` get tagged.

But when rejoining, the dead player will get `"player"` tag back because `onInit(CBlob@ this)` gets run for every character on rejoin which applies the tag. This causes a bug where a crate that has a dead character inside will - after rejoining - show a "get out" button instead of the normal inventory button.

I added 
```
	if (this.hasTag("dead"))
		this.Untag("player");
```
to `onInit()` in `RunnerDeath.as`.
Because `RunnerDeath.as` is at the bottom in the blob scripts list for all the runners, this code effectively removes the "player" tag on rejoin. This is the simplest solution. Alternatively, you could also not tag "player" if tag "dead" is present, in each runner's blob init function.

Tested on dedicated server, it works and the crate bug doesn't happen anymore.

I think there could be more bugs like this in the game. I will search for more.

### DIFF
--- a/Entities/Characters/Scripts/RunnerDeath.as
+++ b/Entities/Characters/Scripts/RunnerDeath.as
@@ -9,6 +9,9 @@ const f32 HEAVY_CARRIED_BLOB_VEL_SCALE = 0.6;
 
 void onInit(CBlob@ this)
 {
+	if (this.hasTag("dead"))
+		this.Untag("player");
+
 	this.set_f32("hit dmg modifier", 0.0f);
 	this.set_f32("hit dmg modifier", 0.0f);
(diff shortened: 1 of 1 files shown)


Vam-Jam (approved):

asumagic:
Hm, interesting find. I think the fundamental issue is that the tag applies unconditionally `onInit`, except that a server will sync those tags to clients when they log in anyway. I don't know if we have a convenient/short facility for detecting this, which would be more robust than relying on the `onInit` order.

Issue #2294, opened 2025-01-04
Coinslot problem on Biurza_Wealth_Drain


mugg91:
## Current in-game behaviour.

On the map `Biurza_Wealth_Drain` the coinslot in the middle is overlapping with the chest that comes out when using the coinslot.
This can lead to a problem where you want to open the chest but accidentally use the coinslot again.
Perhaps coinslot should only be usable once in TDM.

PR #2297, opened 2025-01-05
[changed] Coinslot can only be used once in TDM

mugg91:
## Status

## Description

Fixes https://github.com/transhumandesign/kag-base/issues/2294

This makes it so that you can pay for coinslot only once in TDM.
All maps that use Coinslot only benefit you once - there is no point in paying multiple times.
This aims to fix an issue with accidentally paying for coinslot in situations where you don't want to.

If you try to use the coin button when the coinslot has already been paid, there will be a "NoAmmo" sound.

Tested in online, works as intended.

### DIFF
--- a/Entities/Structures/Components/Source/CoinSlot/CoinSlot.as
+++ b/Entities/Structures/Components/Source/CoinSlot/CoinSlot.as
@@ -34,6 +34,7 @@ void onInit(CBlob@ this)
 
 	this.addCommandID("server_activate");
 	this.addCommandID("client_activate");
+	this.addCommandID("client_fail");
 
 	AddIconToken("$insert_coin$", "InteractionIcons.png", Vec2f(32, 32), 26);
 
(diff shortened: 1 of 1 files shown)


mehwaffle10:
Aren't there cases where you may want to be able to buy a mechanism multiples times? To activate a trap or slot machine etc?

Vam-Jam:
I'm closing this for now, I'd rather keep mechanisms the same across game-modes. I also don't think using the coin slot twice is a big issue to begin with.

mugg91:
> Aren't there cases where you may want to be able to buy a mechanism multiples times? To activate a trap or slot machine etc?

There are only 2 or 3 maps in TDM that use coinslots and they only lead to chests falling down.
There is no benefit in activating coinslots more than once in the existing TDM maps, unless new maps are added that would utilize coinslot in a different way (but for traps, just using a push button would probably make more sense).

Coinslot was only changed for TDM, other gamemodes such as sandbox would have worked the same as before.

With this being closed, what would be an alternative? 
Disabling coinslot for a second after use, before it can be used again? (Although there would be no benefit in using it again whatsoever, as I pointed out)

PR #2303, opened 2025-01-07
Fix tree segments being one-sided/having broken visual randomness

asumagic:
[dev] also made all references to TreeSegment[] into a TreeSegment@[] to avoid incorrect copy construction

## Status

## Steps to Test or Reproduce

Observe trees, before this PR all background leaves for the pine would be to the right.  
This is because the TreeSegment copy constructor is called but doesn't properly copy the random seed. Instead, make it so that the list is a list of handles, so the copy constructor doesn't need to get involved in the first place.

Should make sure trees are still synced between players.

There were visual tweaks to make it so that the leaves are always shown on both sides of the pine tree. The random offsets were edited for visual style reasons.

Need to test on vanilla too

Sample screenshot:

image

### DIFF
--- a/Entities/Natural/Trees/BushyTreeLogic.as
+++ b/Entities/Natural/Trees/BushyTreeLogic.as
@@ -59,7 +59,7 @@ void GrowSprite(CSprite@ this, TreeVars@ vars)
 		this.animation.frame = 1;
 	}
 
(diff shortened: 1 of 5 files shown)


asumagic:
Tested on net, localhost, and tested in vanilla, seems to work

PR #2304, opened 2025-01-07
[fix] Knight now calls RunStateMachine sooner

Vam-Jam:
Knight no longer processes an older state, this *should* not change anything aside from fixing a 1 tick delay with shields.

This also means `knight_clear_actor_limits` does not get called on the first tick you jab/hold sword, not entirely sure if this could result in anything different.

Needs testing in mp.

### DIFF
--- a/Entities/Characters/Knight/KnightLogic.as
+++ b/Entities/Characters/Knight/KnightLogic.as
@@ -241,6 +241,11 @@ void onTick(CBlob@ this)
 		return;
 	}
 
+	if (!knocked)
+	{
+		RunStateMachine(this, knight, moveVars);
+	}
+
 	Vec2f pos = this.getPosition();
 	Vec2f vel = this.getVelocity();
 	Vec2f aimpos = this.getAimPos();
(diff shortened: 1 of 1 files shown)


Vam-Jam:
New commit changes the following:
- Client sync logic moved to 'HandleSyncedState' func, fixes a bug where RunStateMachine could end up looping in on itself forever when trying to debug sync states.

- Shield state is now synced, previously the server would only sync the sword being drawn.

Vam-Jam:
New commit fixes knight.state being stuck at `KnightState::Normal` on join, resulting in actions not being synced until knocked.

TerminalHash (commented):
works good for us on Gruhsha, but without testing more complex and ping-dependent mechanics. It would be nice if specifically for the test of these mechanics gathered good players to avoid possible complaints about lost “mechanics” (which are more likely bugs due to the peculiarities of the implementation, imho).

While playing EU Captains, im compared the feel, there are a lot of moments where the client shows a fake shield is missing, which I think has a positive effect on the game.

asumagic:
Putting as discussion for now to think about later because it does change gameplay timings

PR #2308, opened 2025-01-11
[added] changing class at base copies your current emote

mugg91:
## Status

## Description

When changing class at a tent, your current emote will be copied.


Vam-Jam (approved):
Works locally and on dedi server.

Issue #2310, opened 2025-01-12
stone blobs can take fire damage


mugg91:
## Description

1. Place a stone door or an obstructor (solid; current dev build) <s>, place some wood back tiles in front of it.</s>
2. Change team
3. Shoot fire arrow at it, let the fire arrow expire
4. Damage is applied to the stone blob. (Makes a stone damage sound)

I think `shape.getConsts().isFlammable;` / `blob.isFlammable()` should be checked before processing fire hits.

PR #2311, opened 2025-01-12
[fixed] Stone blobs will no longer take damage from fire

mugg91:
## Status

## Description

Fixes https://github.com/transhumandesign/kag-base/issues/2310

Stone doors, obstructors(solid, in the current dev build) and possibly other stone blobs could take fire damage, if an adjacent wood back tile starts burning.

This PR fixing it by changing `StoneStructorHit.as` to set fire damage to 0.
With the damage being 0, `Stone.as` won't do a stone hit sound either.

Tested in online, works.

### DIFF
--- a/Entities/Structures/Common/StoneStructureHit.as
+++ b/Entities/Structures/Common/StoneStructureHit.as
@@ -1,6 +1,5 @@
 //scale the damage:
-//      knights cant damage
-//      arrows cant damage
+//      knights, arrows and fire doesn't deal damage
 
 #include "Hitters.as";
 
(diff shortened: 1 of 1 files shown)


mugg91:
## Before

https://github.com/user-attachments/assets/bc9104f5-68a2-4298-bfb8-ad552bfe1f5b

## After

https://github.com/user-attachments/assets/1f0348a8-0f01-45bb-858d-9b7ac6767996

The fire cross is odd, I might change it later on (should only do fire cross if the fire arrow is picked or stepped on twice).

asumagic (approved):

PR #2319, opened 2025-01-21
Lighting experiments

asumagic:
- Blobs on fire now emit red-ish light particles every few ticks.
- Fireplaces now have a wider radius but a much more red-orange tint.
- Fire arrows have reworked particles, and also emit light particles on their flight.
- Trap blocks (only relevant to specific gamemodes) now let light pass when open. This was possible before, but the legacy lighting engine was very bad at updating them properly.
- Tiles on fire are now tracked via a `FireTracker.as` script, which listens to flood layer updates. Tiles on fire emit light particles.
- Added an optional `ShadowmapDebug.as` file for messing around the shadowmap configuration.
- Added commented out code to adjust the shadowmap on TDM, which allows to apply _some_ shadowing to TDM, but it seems like players don't want it.
- Adjusted light texture to be softer.
- Necromancer magic orbs emit blue-purple light particles.


asumagic:
Note that this requires staging patches that hasn't made it to the the staging builds.

asumagic:
Will merge for now and see what people think of it in the test build.

PR #2322, opened 2025-01-26
Custom Head Syncing

Vam-Jam:
## Status

## Description
Rework of https://github.com/transhumandesign/kag-base/pull/1051, still in WIP.

Planned features are as follows:
- [x] `CustomHead.png` is sync to the server and other clients on join. 
  - [x] Fix up the current sync bugs
  - [x] Custom head toggle
  - [x] Only the following users can use it:
    - [x] patreon supporters (Round Table)
    - [x] accolades with the custom head flag
    - [x] super admins & isDev flag
    - [x] permanent head users
- [x] Settings menu changes:
  - [x] Patreon tab
  - [x] Allow you to see your custom head without having to join a game 
  - [x] Reload png button (?)
  - [x] Toggle on/off
  - [x] Link to patreon (if you don't have access to custom heads)
  - [x] Link to the directory where `CustomHead.png` is located
 - [ ] Fix up other settings menu width


furai:
Uh, this feature shouts security issues.

Vam-Jam:
I don't think there's much potential for security issues right now. There's room for improvement but I don't think its a deal breaker.

The png is split up into 'R G B' values for every pixel before being sent over the network, instead of sending the file as is. 

If the png is malformed and Irrlicht can't read it, then the texture would never get sent in the first place (and or the server texture creation would fail).

If somebody tries to make a 4000x4000 png, the only thing that is sent is the first 16x48 pixels. (Client checks the width/height before syncing & server only reads the first (16x48x3)+1 packets, ENET would have a size limit as well AFAIK)

If we want more validation/verification, that can be handled engine side on the server.

PR #2116, opened 2024-09-04
[fixed] localhost building

Gingerbeard5773:
## Status

- **READY**

## Description

* Fixes a bug related to building two structures at the same time on localhost.
* Fixes a bug where pages wouldn't change in localhost.
* Improves mod support by removing hack code.

## Steps to Test or Reproduce

On localhost, build a building.
If using the old build: you will notice that the building is made twice and one of the buildings will be destroyed.
If using the fixed build: no duplicate building is produced.

Netcode was altered so make sure to test on dedicated server. 

It should be noted that because important builder files were modified, this PR may break some mods- though it should be easy to update those mods to fix any issue.

resolves #2137, resolves #2081, resolves #2196

PR #2333, opened 2025-05-11
Build Height Changes

mehwaffle10:
## Status

## Description
Changes behavior for interacting with the top of the map. Towers at max build height essentially get an artificial buff in how hard they are to get over. This is extra noticeable with the incoming general cut to all map heights with my map cycle update.

## Steps to Test or Reproduce

Adds new `no blobs` and `no solids` sectors to the top of the map. The `no blobs` sector prevents building all static blobs besides shops. The `no solids` sector prevents building any solid blocks. You can still spike drop in these new sectors, as long as the spike will fall.

Original `no build`:
image

New `no solids`:
image

New `no blobs`:
image

These have the following net effect:
1. You can always build shops at the top of the map
2. You can always put solid blocks in front of doors on the top of the map, unless you have vertical ladders or an exposed shop in the way

You can no longer place backwall (or any block) in the top 3 blocks of the map. This avoids an issue with invulnerable backwall at the 3rd layer at the top of the map where it could be built inside the `no build` zone.

Blobs are no longer limited to the top of the map when it is not a cave map (defined as any blocks at the top of the map in `KAG.as`). This allows players to get over the top of towers much easier rather than being bottle-necked at build height. The border color for the top of the map is lighter when it is not a cave map.
image

Cave maps are dark and prevent blobs from passing through as normal. Crates can now deploy at any height:
image

These changes have been used on US captains for over a year. They are pretty universally well received.

### DIFF
--- a/Entities/Characters/Archer/ArcherAnim.as
+++ b/Entities/Characters/Archer/ArcherAnim.as
@@ -536,7 +536,9 @@ void doRopeUpdate(CSprite@ this, CBlob@ blob, ArcherInfo@ archer)
 		return;
 	}
 
-	Vec2f adjusted_pos = Vec2f(archer.grapple_pos.x, Maths::Max(0.0, archer.grapple_pos.y));
+	// Only limit grapple to top of map if it's a cave map
+	CRules@ rules = getRules();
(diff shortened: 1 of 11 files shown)


Vam-Jam:
This PR has conflicts with the following past PRs:
https://github.com/transhumandesign/kag-base/pull/2116
https://github.com/transhumandesign/kag-base/pull/2001

Aside from that, I'll mark it as testing needed as its a balance change

GELGELAT:
nice!

PR #2338, opened 2025-06-07
[fixed] magazines wouldn't spawn food

SonantDread:
## Status

## Description

XORRandom was treated like it was inclusive, preventing food spawning (mainly in TDM).

## Steps to Test or Reproduce

1) Load into a map with magazines, such as Noah_Subterranean
2) Food should spawn from a magazine. If not, try to restart the map a few times.

### DIFF
--- a/Scripts/MapLoaders/BasePNGLoader.as
+++ b/Scripts/MapLoaders/BasePNGLoader.as
@@ -183,7 +183,7 @@ class PNGLoader
 				string name;
 				if(alpha == items.length) // random
 				{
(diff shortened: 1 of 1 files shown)


mugg91:
Why was this closed?

Vam-Jam:
Food was disabled intentionally in the past, mostly to prevent RNG from giving one side a bigger advantage.

Maybe it's worth revisiting if we ever build out a way to allow maps to have their own settings, but for now it's fine as is.

PR #2344, opened 2025-06-22
add balloon into vehicle shop

TerminalHash:
## Status

## Description
The balloon is hardly used in the game, but it still exists. Why not add it to CTF and spice up the general chaos? :fridge:

In general, the ball is a little useless without an open ceiling (which is why this pull request depends on https://github.com/transhumandesign/kag-base/pull/2333), as well as the ability to launch kegs and bombs directly from it. There is code for this on Gruhsha CTF, but it's a bit bad and part of it is taken from another mod (Territory Control Classic), so I don't know if I can reuse it for this pull request. So for now, the pull request is marked as **IN DEVELOPMENT**, but everything can be changed if necessary.

## Showcase

Balloon in the shop:
image

Crate with balloon:
image
image


GELGELAT:
no.

PR #2360, opened 2025-09-15
[new] Enable active sync on a few blobs

Vam-Jam:
The TLDR is that the blob is synced whilst the box2d body is active (you can check this with g_debug 1). You only want to use this on blobs where knowing where the blob position is very important (e.g. players, projectiles, explosives etc). The trade off is that this will consume more bandwidth.

Disabling this/not setting it will use the default sync method, where every tick, we check to see if our blob position has moved in any direction far enough for the sync to be worth it.

Currently enabled for:
- Anything that uses `RunnerDefault` (Archer, Builder, Knight, Migrant, Necromancer, Princess, Trader)
- Arrows
- Bombs
- BallistaBolt
- Keg
- Mine

### DIFF
--- a/Entities/Characters/Scripts/RunnerDefault.as
+++ b/Entities/Characters/Scripts/RunnerDefault.as
@@ -22,6 +22,8 @@ void onInit(CBlob@ this)
 	this.SetChatBubbleFont("hud");
 	this.maxChatBubbleLines = 4;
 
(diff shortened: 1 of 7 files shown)


Vam-Jam:
There's an engine PR that fixes that. The issue was caused by some out of screen optimizations. 


We could enable active sync on that since it's a short lived projectile anyhow.

Issue #1813, opened 2023-12-30
Ferrezinhre_hardened has the tents switched with one another


mugg91:
## Description

On `Ferrezinhre_hardened` the red tent is on the left and the blue tent is on the right, whereas all the other maps have it the other way around. 

This causes the issue, that the team switch buttons will not accommodate for this.

https://i.imgur.com/4oVrRr2.png

## Suggestions

Either the team switch buttons logic should check where the tents are on the map and line up the buttons accordingly, or the map `Ferrezinhre_hardened` should switch the tent colors.

PR #2364, opened 2025-09-17
Mapcycle Update

### LINKED ISSUE #2312: clean up and fix the stupid maps

These two maps just don't have trees, its a living hell.
Ferrezinhre_Golden_Melancholy
Ferrezinhre_Submissive_Transfiguration
34d2a2976178a85b

These maps just lead to dead ends.
Fellere_FourRivers_v2
Fellere_FourRivers
Ferrezinhre_Sunrinhelm
Ferrezinhre_Sunrinhelm
**Proposed Change**
Add trees to the first two maps. (I had TDM on)
Ferrezinhre_Golden_Melancholy_v2
Ferrezinhre_Submissive_Transfiguration_v2
6dfb86e48ea7b8ed
3769a5fb37ec9348

Other second two cards are removed from the loop.

### LINKED ISSUE #2363: Flower in Biurza_Hop_Hazard doesn't grow
## Description

The flower on the right side in the TDM map `Biurza_Hop_Hazard` doesn't grow.

mehwaffle10:
## Status

Work-in-Progress

## Description

Remake of map pools for all game modes. Includes maps from the current pool, old maps previously in the pool, and new maps submitted in discord. Also reorganizes the map folder structure to be much more streamlined. in `Base/Maps` there is a folder per gamemode. All maps in the top level of their gamemode directory is currently in the pool. There are also `Dormant` and `Retired` folders in each gamemode's folder for maps that are not currently in the pool and maps that should never be added to the pool. Removes duplicate maps in different directories or multiple versions of the same map


mugg91:
`Coroz_floating_rifts` has flowers on stone tiles that don't grow.

mugg91:
I guess it is nice that someone is cleaning up maps. Although I don't like cropping on some of the maps (e.g. `Ferrezinhre_Vertigo`). Mappers made some nice designs and playing games should also be about discovering things, like it makes the map feel more alive if you know what I mean. I don't think any of the maps were that big that players were suffering slowdowns and cropping maps doesn't really save that much file size.

mehwaffle10:
Files are trimmed to make maps load faster and less load intensive on weaker clients, not to reduce file size. Very impactful with how often maps are loaded in TDM. I have done what I can to maintain as much detailing as possible while optimizing. I do what I can not to impact gameplay at all either. The only thing cut for detailing in vertigo is some small dirt protrusions that hardly anyone would notice. Plus now it looks like a hamburger which is a W in my books

mugg91:
Maybe you can incorporate https://github.com/transhumandesign/kag-base/pull/1823 to fix https://github.com/transhumandesign/kag-base/issues/1813

Vam-Jam:
Merge conflicts will need to be sorted, I'd imagine you just want to override any changes with yours.

.gitignore
```diff
@@ -0,0 +1,4 @@
+/Maps/CTF/test
+/Maps/TDM/New
+/Maps/SmallCTF/New
+/Maps/CTF/New
```
Vam-Jam: Is this .gitignore intentional?

mehwaffle10: No, it can be removed. Those were subfolders I dumped all the new maps into and pulled into the main area as they were completed. I also believe I changed Git to be case sensitive at one point

PR #2366, opened 2025-09-18
Do not load runner textures on server

asumagic:
## Description

Small optimization to server memory usage. AFAIK there isn't ever a point in having those textures be loaded serverside so let's not.


TerminalHash:
we played several matches with that pr, and it seems that no problems were noticed.

Vam-Jam (approved):

PR #2367, opened 2025-09-20
[change] Mine does not explode when touching a static enemy arrow or ballista bolt

### LINKED ISSUE #2362: Mine explodes when touching a stuck enemy arrow
## Description

When a mine falls into or touches a stuck enemy arrow, it explodes.

I think this shouldn't happen.

mugg91:
## Description

Fixes https://github.com/transhumandesign/kag-base/issues/2362

This makes it so a Mine does not get triggered by an enemy projectile (arrow, ballista bolt) after it has become static.

Tested in offline, works.

I found this to be odd behavior so I made this change. A primed Mine will still explode on collision with a standstill enemy vehicle (dinghy, mounted bow, catapult, etc.)

### DIFF
--- a/Entities/Items/Explosives/Mine/Mine.as
+++ b/Entities/Items/Explosives/Mine/Mine.as
@@ -178,7 +178,7 @@ void onThisRemoveFromInventory(CBlob@ this, CBlob@ inventoryBlob)
 bool explodeOnCollideWithBlob(CBlob@ this, CBlob@ blob)
 {
 	return this.getTeamNum() != blob.getTeamNum() &&
-	(blob.hasTag("flesh") || blob.hasTag("projectile") || blob.hasTag("vehicle"));
+	(blob.hasTag("flesh") || blob.hasTag("vehicle") || (blob.hasTag("projectile") && !blob.getShape().isStatic()));
 }
 
 bool doesCollideWithBlob(CBlob@ this, CBlob@ blob)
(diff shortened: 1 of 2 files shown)


Skayland6909:
Alternatively arrows also get a `"collided"` tag that could be used instead of `getShape().isStatic()` unless there is a reason not to.

TerminalHash:
wdym it's a strange behavior? if you look closely at the mine's code, everything is quite logical + mechanism of mine, as implied by the sprite, means that it will explode in any case if something hits it.

asumagic:
I don't really think it's an inconsistency gameplay-wise at the end of the day, it does make sense for vehicles to trigger it. They can bounce off terrain without triggering, so making it trigger on static projectiles seems kind of unnecessary. An example would be dropping a mine with doors from the top of a tower, it's a bit lame if it explodes on a random arrow.

TerminalHash:
If it exploded on the arrow during the fall, then it's a skill issue of the person who threw mine :fridge:

mugg91:
> Alternatively arrows also get a `"collided"` tag that could be used instead of `getShape().isStatic()` unless there is a reason not to.

Ballista Bolt has a boolean "static" and Arrow has a tag "collided".
Both become a static shape so I used that info.
Alternative solution would be to use a tag or boolean for both projectiles and check that instead.

@TerminalHash 
No problem, I'm just saying it is strange behavior _to me_, since the Arrow isn't moving at all.

Skayland6909:
My understanding is that `hasTag()` is likely slightly cheaper than `getShape()`, and since arrows are more common I thought it would make sense to do something like:
`blob.hasTag("projectile") && (!blob.hasTag("collided") || !blob.getShape().isStatic())`
so that it passes _ever so slightly_ faster for arrows, either that or give the tag to ballista bolt too.

But either way I agree that mines exploding on stuck arrows is silly.

asumagic:
Oh wait, I interpreted the PR as the exact opposite of what it does. I don't know how I've never noticed this happen.

mugg91:
Pushed a commit that changes the check to `hasTag("collided")`. 
Changed instances of `get_bool("static")` in `BallistaBolt.as` to `hasTag("collided")`.
Tested with arrow and ballista bolt in offline Sandbox, works the same as initially.

Issue #2147, opened 2024-10-30
TDM Spawn should not show a change class button


mugg91:
## Current in-game behaviour.

When pressing "Use" while at a TDM spawn, the change class menu shows, but if you are too far away, a green up arrow button appears which you have to click on first, which wastes time having to deal with if enemies are approaching.

## Suggestion

Show the menu even when holding "Use" from afar, with the menu being 0.5 oppacity (to indicate it is not active) if you are too far.
Clicking on the menu while it is not active should not close the menu.

Reduces needless menuing.

Related: https://github.com/transhumandesign/kag-base/issues/1730
The game needs better ways of changing class fast in TDM.

Issue #1730, opened 2023-10-24
Please make it so that when in TDM a player quickly presses the interaction button (E), the player's class changes to the opposite one


NikitaVNsv:
Please make it so that when in TDM a player quickly presses the interaction button (E), the player's class changes to the opposite one.
It will be do game more comfortable : )

mugg91:
I can attest to the fact that trying to click on the Knight button usually fails because I miss 5 times and the enemy Knight kills me in the meantime.

So a quick simple button press would be cool. I'd prefer pressing something like "1" for Knight, "2" for archer, maybe customizable in the settings.

PR #2369, opened 2025-09-20
[added] can change class by key press

mugg91:
## Description

Copied code from `Shop.as` to `ClassSelectMenu.as`.
When the class select menu is opened, you can press number keys to switch to classes.

Tested in offline and dedicated, works.

Addresses https://github.com/transhumandesign/kag-base/issues/1730 and https://github.com/transhumandesign/kag-base/issues/2147 in that a fast way to switch class is introduced.


mugg91:
Added a commit https://github.com/transhumandesign/kag-base/pull/2369/commits/c02308d2cc5e1424e690d731ba51eda588965cd9 that makes it so you can press number keys at Tent and TDM Ruins to switch to a class directly, without having to open the switch class menu first. Works as intended in offline and dedicated. 

Due to https://github.com/transhumandesign/kag-base/issues/2399 , I resorted to using `this.isOverlapping(myblob)`.

I looked at the possibility of simply re-enabling the boolean `enable_quickswap` in `StandardRespawnCommand.as`, but its behavior seems odd and wonky so I left it.

PR #2372, opened 2025-09-24
[changed] Chest color will not change on rejoin

### LINKED ISSUE #2365: Remember chest color
## Current in-game behaviour.

Chests will have different color between clients or when reloading the map.

## Proposed Change

Make chests have the same color across clients and remember it when reloading the map.

mugg91:
## Description

Fixed https://github.com/transhumandesign/kag-base/issues/2365

A chest's color will stay the same when rejoining and between clients.

## Testing

On dedicated server, I spawned 10 chests and rejoined 3 times. The colors didn't change.

Without this PR, the colors will be different every time.

### DIFF
--- a/Entities/Items/Containers/Chest/Chest.as
+++ b/Entities/Items/Containers/Chest/Chest.as
@@ -34,11 +34,14 @@ void onInit(CBlob@ this)
 	CSprite@ sprite = this.getSprite();
 	if (sprite !is null)
 	{
-		u8 team_color = XORRandom(5);
-		this.set_u8("team_color", team_color);
+		u8 col = XORRandom(5);
+		if (this.exists("color"))
+			col = this.get_u8("color");
+		else
+			this.set_u8("color", col);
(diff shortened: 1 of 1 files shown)


Entities/Items/Containers/Chest/Chest.as
```diff
@@ -137,7 +140,7 @@ void onDie(CBlob@ this)
 		1.0f,                               // scale?
 		0,                                  // ?
 		"",                                 // sound
-		this.get_u8("team_color"));         // team number
+		this.get_u8("color"));              // team number
```
Vam-Jam, line 143: I feel like if you could just replace this with `this.getNetworkID() % 5` would remove the need for storing "color"

mugg91: But then the chest won't have a randomized color. Chests would have a predetermined color in TDM maps.

Also this is the same logic used in `Fishy.as`

PR #2374, opened 2025-09-24
[changed] Catapult launches dinghy less far

### LINKED ISSUE #2351: Boat launcher bug
https://github.com/user-attachments/assets/0b837729-7f78-4ca9-b693-4ac215551bd6
https://github.com/user-attachments/assets/74c5d616-cda4-458f-ab96-3ab351262bde
everything is as shown in the video
delete this crap quickly!!!

mugg91:
## Description

Fixes https://github.com/transhumandesign/kag-base/issues/2351

This makes dinghy to get launched the same as players (modifier `0.75f` instead of `1.1f`) and impose the same cooldown as if a player was launched, hopefully fixing the issue with players sitting in dinghies getting shot across the whole map, but not nerfing it too hard so players will still get launched at the velocity they would have got as if they sat in the catapult's mag alone.

Needs testing in match.

### DIFF
--- a/Entities/Vehicles/Catapult/Catapult.as
+++ b/Entities/Vehicles/Catapult/Catapult.as
@@ -93,7 +93,7 @@ class CatapultInfo : VehicleInfo
 			vel += (Vec2f((_r.NextFloat() - 0.5f) * 128, (_r.NextFloat() - 0.5f) * 128) * 0.01f);
 			vel.RotateBy(this.getAngleDegrees());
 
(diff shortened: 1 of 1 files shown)


mugg91:
It looks like the trick can also be done with crates (breaking after launching or by using "get out" right after launching).
Perhaps we need a slightly different solution.

Maybe use `1.1f` for an empty dinghy/crate and something between `0.75f` ~ `1.1f` for a dinghy with people in it and a crate with a player inside.

PR #2383, opened 2025-09-28
[changed] Water explosion will wake up sleeping players

### LINKED ISSUE #2092: Stun doesn't detach you
## Description

When getting shield stunned or water bombed while riding a catapult or while sleeping in a quarters, you will not get detached.
When sleeping and getting waterbombed, the character even makes a grunt sound but just keeps sleeping.
I think stun should detach runners.

mugg91:
## Description

Fixes https://github.com/transhumandesign/kag-base/issues/2092

This makes it so waterbomb and water arrow will wake up players who are sleeping in quarters.

Previously, water bomb could explode, even in the sleeping player's inventory, and the player would make a grunt sound but still keep sleeping.

Works in testing.

### DIFF
--- a/Entities/Industry/CTFShops/Quarters/WakeOnHit.as
+++ b/Entities/Industry/CTFShops/Quarters/WakeOnHit.as
@@ -1,6 +1,7 @@
 // WakeOnHit.as
 
 #include "KnockedCommon.as"
+#include "Hitters.as"
 
 void onHealthChange(CBlob@ this, f32 oldHealth)
 {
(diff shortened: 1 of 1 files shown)


mugg91:
<s>I can also make shield stun to not hit sleeping players **OR** hit and wake up sleeping players, if wanted.</s>

mugg91:
Nevermind my previous comment. Turns out when you attach to a Quarters, `attached.getShape().getConsts().collidable = false;` happens so you actually can't shield drop on sleeping players. There was no problem with that.

PR #2385, opened 2025-10-01
[fixed?] Drill cannot be used by Knight/Archer outside of TDM

### LINKED ISSUE #2346: Knight and Archer can activate Drill
## Description

For some reason, I'm able to use the Drill as Knight and Archer in vanilla official servers.
The Drill will not break anything. It can sometimes heat up but it will not cause the character to drop it or to take damage.
However, it will cause the Drill to not be pickable for a short while.

If the Drill was heated up by Knight or Archer and you switch to builder and try to use the drill, the heat meter will reset to 0.

In previous patches Knight or Archer couldn't use the Drill by pressing `action_1`.

mugg91:
## Description

(Hopefully) Fixes https://github.com/transhumandesign/kag-base/issues/2346

The variable `sv_gamemode` seems to be bugged. When playing on a server with one gamemode and then switching to another server with a different gamemode, chances are `sv_gamemode` will not update.

Drill checks for `sv_gamemode` to determine if it can be used. If you are builder or if you are in TDM (according to `sv_gamemode`), you can use it.
When joining TDM and then a different gamemode, it is possible to use the Drill even in Sandbox/CTF, although with glitchy effects (it will heat up but not destroy tiles or deal heat damage to you).

To fix this, I exchanged the check for `sv_gamemode` with a check for `getRules().gamemode_name`. Some other script files also check for the latter, so it should be fine. As far as I tested, the latter seems to update correctly when switching servers.

### DIFF
--- a/Entities/Industry/Drill/Drill.as
+++ b/Entities/Industry/Drill/Drill.as
@@ -207,7 +207,7 @@ void onTick(CBlob@ this)
 			sprite.PlaySound("DrillOverheat.ogg");
 		}
 
-		if (holder.getName() == required_class || sv_gamemode == "TDM")
+		if (holder.getName() == required_class || getRules().gamemode_name == "Team Deathmatch")
 		{
 			if (!holder.isKeyPressed(key_action1) || isKnocked(holder))
 			{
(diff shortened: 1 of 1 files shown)


Vam-Jam:
This looks sane but I would rather fix this engine side if `sv_gamemode` is the issue here. There's a few other places where sv_gamemode is used which would also need fixing.

I'll keep it opened for now.

PR #2432, opened 2025-10-29
[changed] Grass/Plants won't grow near a burning fireplace

mugg91:
## Description

This is part of https://github.com/transhumandesign/kag-base/pull/1465

Makes it so grass and plants won't grow near a burning fireplace.

### DIFF
--- a/Rules/CommonScripts/RegrowPlants.as
+++ b/Rules/CommonScripts/RegrowPlants.as
@@ -246,11 +246,39 @@ void onTick(CRules@ this)
 			{
 				f32 random_grow = XORRandom(10000) * 0.0001f;
 
+				// GRASS
 				if (random_grow - tinfo.grassLuck() <= grass_grow_chance && !map.isTileGrass(tile_above.type))
 				{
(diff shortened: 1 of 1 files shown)


Rules/CommonScripts/RegrowPlants.as
```diff
@@ -246,11 +246,39 @@ void onTick(CRules@ this)
...
+					if (map.getBlobsInRadius(tile_position, 8.0f, blobs))
+					{
+						for (int a = 0; a < blobs.size(); a++)
+						{
+							CBlob@ blob = blobs[a];
+							if (blob is null)
+							{
+								continue;
+							}
+							if (blob.getName() == "fireplace" && blob.getSprite().isAnimation("fire"))
```
Vam-Jam, line 267: I'm not entirely sure that `getSprite().isAnimation` would work on a linux dedi server? You should generally avoid doing any type of sprite checking on the server

mugg91: The same check is used in `Fireplace.as` and `Arrow.as`.

https://github.com/transhumandesign/kag-base/pull/2290 introduces a new tag `"extinguished"` which could be used for checking instead of using `isAnimation()`.

PR #2451, opened 2025-11-09
[change] neutral archer can grapple in red barrier

### LINKED ISSUE #2450: Allow grapple on red barrier if archer is team "-1"
## Description

Grapple gets canceled when touching red barrier.
But in case archer is in team -1 (usually spectator team or neutral/animal team), he should have the privilege of grappling through red barrier.

mugg91:
## Description

Fixes https://github.com/transhumandesign/kag-base/issues/2450

This lets neutral team archer grapple in red barrier.

### DIFF
--- a/Entities/Characters/Archer/ArcherLogic.as
+++ b/Entities/Characters/Archer/ArcherLogic.as
@@ -751,7 +751,7 @@ bool checkGrappleBarrier(Vec2f pos)
 
 bool checkGrappleStep(CBlob@ this, ArcherInfo@ archer, CMap@ map, const f32 dist)
 {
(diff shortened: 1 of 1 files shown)


mugg91:
## Before

https://github.com/user-attachments/assets/1f084676-9416-4ee4-8a9e-c0a8dd6668b6

## After

https://github.com/user-attachments/assets/e9f9a96d-74cb-4261-85f0-bd3912f319a3

Vam-Jam (approved):
Not sure what the use case of this is, removing red barrier can be done via other commands, but its small enough of a change that it's fine as is

PR #2453, opened 2025-11-12
[added] Ambient music when being near or inside water

mugg91:
## Description

This adds ambient music and slightly improves the music script files.

- Don't determine booleans beforehand, determine them on the go instead.
- Use the same logic when determining if we are in a cave or up in the sky across `CTFMusic`, `WARMusic` and `ChallengeMusic`
- Night ambient added to `CTFMusic`
- "Near water" ambient added to `CTFMusic`, `WARMusic` and `ChallengeMusic`
- Underwater ambient added to `CTFMusic`, `WARMusic` and `ChallengeMusic`
- Use map tilesize rather than hard-set `8` when checking booleans

`ChallengeMusic` checks if it should fade to an ambient, but only in warmup for some reason. I didn't touch this script further but might need looking into.

`TDMMusic` is not touched, doesn't support ambient music (yet).

"Near water" ambient taken from https://pixabay.com/sound-effects/flowing-water-at-buttsbury-wash-59302/
> Water, River, Field-recording sound effect. Free for use.

Underwater ambient taken from https://pixabay.com/sound-effects/underwater-ambiencewav-14428/
> Ambience, Underwater, Constant sound effect. Free for use.

Tested in offline.


mugg91:
Video demonstrating the ambient music.

https://github.com/user-attachments/assets/4869b581-4c8f-4d8c-84f1-c02d0b876f7c

Please be aware of bug https://github.com/transhumandesign/kag-base/issues/2458

PR #2454, opened 2025-11-16
[fixed] Scoreboard stuck bug

mugg91:
## Description

When many people are online and you scroll down the scoreboard, then some people leave and you bring up the scoreboard again, it is possible that you could find yourself unable to scroll up the scoreboard.

It is possible to do the bug by entering CTF tutorial and using `AddBot()` a bunch of times. The Bots would enter Red team but then switch over the Blue team. If you add so many bots the scoreboard reaches the bottom of your screen, and scroll down, and then the bot switches to blue (so Red Team's field on the scoreboard doesn't get rendered anymore), the bug could happen and the scoreboard gets stuck.

After applying the change in this PR, the scoreboard can be scrolled back up in cases where it would have gotten stuck. So the scoreboard doesn't get stuck anymore. I haven't seen any odd behavior with this change. The scrolling and scoreboard offsetting looks to work as intended.

### DIFF
--- a/Rules/CommonScripts/ScoreboardRender.as
+++ b/Rules/CommonScripts/ScoreboardRender.as
@@ -709,7 +709,7 @@ void onRenderScoreboard(CRules@ this)
 	float screenHeight = getScreenHeight();
 	CControls@ controls = getControls();
 
-	if(scoreboardHeight > screenHeight) {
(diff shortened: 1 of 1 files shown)


mugg91:
Video demonstrating the bug and the fix

https://github.com/user-attachments/assets/e6fc49cf-7a41-4378-a49b-51978d18e70e

PR #2455, opened 2025-11-22
Red Barrier No Build

mehwaffle10:
## Description

Prevent building in the red barrier at the start of the match.

Makes the intent in how far you can build during build phase clearer, not to mention much less frustrating. Building in the red barrier is a bit of an acquired skill and is annoying to execute, amplified by the fact that less experienced teammates are likely to interrupt/block it. It's also just clunky trying to tower up while building in the barrier without ladder spamming.

Red barriers will be adjusted as need be as part of my map cycle update.

## Steps to Test or Reproduce

Enter CTF or SmallCTF. Attempt to build in the red barrier. You will be unable to place tiles or blobs.
Start the match. You will be able to build normally where the red barrier was.

You can still break blocks in the red barrier.
The user is given feedback identically to a no build zone (the flag room for example). Their cursor will not align to the tiles, the selected tile or blob will be red (and wiggle if it's a tile), and left clicking will play the error sound `NoAmmo.ogg`.

## Implementation Discussion

I utilized the actual red barrier rather than creating a no build sector on top of the red barrier to avoid rendering multiple red rectangles and to allow breaking blocks as normal. I added a helper function to make the checks cleaner, but notably required `CRules` for the checks. The places we needed to call the function from didn't have `CRules` on hand, so I decided to include a `getRules()` call to keep it clean in the other files. Please feel free to edit as need be if this is considered bad form, as the other functions in `RedBarrierCommon.as` all pass `CRules` in as a parameter.


mugg91:
I like this change.
Just two thoughts:
- Maybe spikes or blobs next to the barrier should not get affected by the pushing. It would make more sense for a spikes collapsing next to the barrier to fall down instead of getting pushed.
- You can still build buildings that extend into the barrier, which might get exploited to place backgrounds. Maybe should prevent building a building inside of red barrier.

TerminalHash:
It's hard to say how useful it will be in a public CTF tbh %)

mehwaffle10:
Prevented buildings from being built in the red barrier as well

mehwaffle10:
Blob interactions with the red barrier should likely be a separate PR. I believe you made one for arrows already

Vam-Jam:
If you can sort out the conflicts, I'll take a look after next week.

Vam-Jam:
Tested on staging dedi + localhost, ctf, small ctf and tth. Works fine, couldn't find any silly bypass, so it seems good. Marking it as ready for now as it functionally works.

Vam-Jam (approved):

Issue #1230, opened 2022-04-16
Organize better for releases


AsuMagic:
Release throughput is rather terrible partly because contributors (especially me) have finite motivation and time to help with releases.

Some people have been stepping up recently to help (e.g. Mazey) but the release process is currently way sub-optimal since core contributors are loaded with work that takes away time that could be spent on things others mostly can't do (*cough* staging release *cough*).

I think this is at least partly due to rather terrible organization, so this thread is written with the hope of creating a more robust procedure for KAG releases that reduces reliance on the very few "all-powerful" contributors and sparking some discussion on what could be improved. The current procedure is barely documented and was much better suited for a time where the game was not as community-driven and when core developers had usually more time to provide.

## Required work on releases

There are a couple things I generally need to do which represents most of the time I need to spend when pushing KAG releases (engine work and debugging aside):

1. Reviewing PRs. Code review is hard and annoying but it ultimately reduces delays caused by finding bugs late in testing. A few people have been (very helpfully) reviewing PRs but the massive number of PRs makes it difficult to catch up and I still end up reviewing most of them.
2. Writing changelogs. We do have automated changelogs in the CI/CD, but a. nearly nobody bothers writing good, descriptive commit messages as the `[added]`/`[dev]`/etc. convention is not enforced, b. they often don't properly represent all changes and c. they don't attribute the changes to a person, so we have to manually add that.
3. Writing posts: Discord, Steam, etc., which reuses the same changelogs mentioned before, but formatting, writing down a summary and checking the syntax is also time-consuming. Steam *requires* creating assets for announcement posts, which if helpful for visibility is *also* annoyingly time consuming.
4. Testing. Organizing test sessions, following a procedure (e.g. for mapcycle testing), scanning through merged PRs to see what needs to be tested, setting up test servers, etc. is overall time consuming.
5. Catching up with stuff that is not written in TODO lists but really should, and that sometimes go forgotten - e.g. Patreon heads.
6. Decide what to merge and what not to. This is more difficult than it appears because first, there is no consensus for some of the changes and little discussion (although CTF rebuild had a good approach to gathering feedback, it is not the norm and its changes are still controversial to some), and second, some PRs impactful on gameplay are published without prior discussion, and not merging PRs is a frustrating mess for everybody involved.
[...]

PR #2456, opened 2025-11-22
No build damage

### LINKED ISSUE #568: Destruction of shops or doors by explosive do not destroy the backwall
**Description**

When blowing up enemy shops or doors, their back tiles will generally stay intact.

**Reproduction**

1. Join singleplayer Sandbox
2. Build a nice thin tower with shops and doors
3. Swap teams `!team 1` and light a keg inside

**Expected Behavior**

Most back tiles should be destroyed, but this does not happen.

**Screenshots**

before

after

mehwaffle10:
## Description

Allow backwall to be damaged in `no build` zones and behind blobs.
 
Shops and doors protect the backwall behind them from any given instance of damage until destroyed. This can greatly impact the damage a well placed explosive or siege shot can do. For example, a keg in the middle of a tower will often do nothing besides destroy a couple of shops and/or doors. This is extremely frustrating as a player, especially considering how difficult it can be to actually land solid hits. Any significant damage often artificially requires two kegs minimum, since the first will clear the blobs out of the way. Since players can only carry one keg at a time, this leads to spam being the only consistent way to punch through buildings. Whether that be spamming from a shop on top of the enemy tower or tramp keg spam or siege spam, there is less precision play and agency from single players.

The flag room can permanently protect backwall from anything but fire. This leads to builders being able to hold a flag room with a single layer of block spam even under heavy fire from kegs or siege. This can greatly drag out a match where the losing team has practically nothing. It's miserable to play for the losing team and frustrating for the winning team (or toxic depending on how you view it). It contributes heavily to the problem of long running stalemates that we currently see often in pub matches, making it that much harder to get over the goal line.

Here is an example of a well placed keg on a dummy tower. Note, this tower doesn't even have that many doors, the problem worsens the more blobs there are, which tend to get spammed in pub matches.

Keg placement:

image

Keg effect before changes:

image

Keg effect after changes:

image

This helps move the dial of power between builder and the combat classes a bit more back in the direction of the combat classes. In the game's current balance, the combat classes heavily struggle to do anything against even the most basic of builders without a builder immediately with them or an oppressive tower on their side.

Explosives will break the backwall behind a door often before breaking the door itself. While the door will prevent a tower from collapsing, the lack of backwall prevents support for building above. While a smaller detail, this does have an impact on the efficacy of direct attacks especially on taller towers and again slightly shifts the balance of power back to the combat classes.

image

Note: Explosives will be blocked from hitting the next block over unless the blob is destroyed. Example with one bomb, which would normally hit a much larger radius:

image

image

These changes have been in place on the US captains mod for 3.5 years. They are well received and have improved gameplay over years of testing.

That said, they are not without fault. The biggest issue in my eyes is that it can sometimes be confusing/hard to see whether there is backwall under a blob, especially large opaque blobs like closed stone doors or especially tunnels. I suspect this was the original reason for the current implementation.

Despite this, I believe this relatively small change code wise has a large net positive effect on play.

P.S. This is also the cause of the "bug" of invincible backwall at the top of the map that is frequently exploited by sky towers, especially in SmallCTF. This PR fixes that, but my PR for the map ceiling changes also already addresses this by simply preventing building backwall that high.

P.S.S. I *believe* the change I made in `BallistaBolt.as` doesn't actually have any effect unless there are wood or stone tiles in a `no build` zone, which is impossible currently as far as I know. I wanted to make the change for consistency/cleanliness regardless.

Closes #568

## Steps to Test or Reproduce

For each of the following cases:
- Backwall in the flag room
- Backwall behind shops
- Backwall behind doors
- Backwall at the top of the map

Do each of the following:
- Use any explosive (bomb, keg, mine, bomb bolts)
- Launch a boulder out of a catapult
- Launch normal catapult rocks

The backwall behind the blobs or in the flag room will be destroyed instead of surviving.

### DIFF
--- a/Entities/Items/Boulder/Boulder.as
+++ b/Entities/Items/Boulder/Boulder.as
@@ -97,10 +97,6 @@ void Slam(CBlob @this, f32 angle, Vec2f vel, f32 vellen)
 	Tile tile = map.getTile(pos);
 	if (map.isTileBackgroundNonEmpty(tile))
 	{
-		if (map.getSectorAtPosition(pos, "no build") !is null)
-		{
-			return;
-		}
 		map.server_DestroyTile(pos + Vec2f(7.0f, 7.0f), 10.0f, this);
 		map.server_DestroyTile(pos - Vec2f(7.0f, 7.0f), 10.0f, this);
 	}
@@ -123,14 +119,8 @@ bool BoulderHitMap(CBlob@ this, Vec2f worldPoint, int tileOffset, Vec2f velocity
 
 	if (map.isTileCastle(t) || map.isTileWood(t))
 	{
-		Vec2f tpos = this.getMap().getTileWorldPosition(tileOffset);
-		if (map.getSectorAtPosition(tpos, "no build") !is null)
-		{
-			return false;
-		}
-
 		//make a shower of gibs here
-
+		Vec2f tpos = this.getMap().getTileWorldPosition(tileOffset);
 		map.server_DestroyTile(tpos, 100.0f, this);
 		Vec2f vel = this.getVelocity();
 		this.setVelocity(vel * 0.8f); //damp
--- a/Entities/Items/Explosives/Explosion.as
+++ b/Entities/Items/Explosives/Explosion.as
@@ -260,7 +260,7 @@ void Explode(CBlob@ this, f32 radius, f32 damage)
 
 							if (canHit)
 							{
-								if (canExplosionDamage(map, tpos, tile))
+								if (canExplosionDamage(tile))
 								{
 									if (!map.isTileBedrock(tile))
 									{
@@ -369,7 +369,7 @@ void LinearExplosion(CBlob@ this, Vec2f _direction, f32 length, const f32 width,
 				}
 				else if (t != CMap::tile_empty && t != CMap::tile_ground_back)
 				{
-					if (canExplosionDamage(map, tpos, t))
+					if (canExplosionDamage(t))
 					{
 						if (!justhurt)
 							damaged = true;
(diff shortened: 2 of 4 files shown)


Vam-Jam (approved):
Works on staging dedi + locally, will mark as ready + discussion

GELGELAT:
ADD IT YES!

Dredonkey:
I like the changes, would like to see this merged. My only concern is that it may buff siege too much, especially since siege is significantly more cheesy in vanilla than in US captains. A bit unrelated, but I would really like to have a faster release cycle after the update this is included in, so that if siege does get too op some stop gap changes can be implemented. I think reworking siege before the next update would be bad since we really need to a get a release out to fix annoying bugs that have been in the game for over a year at this point... which is just ridiculous.

TerminalHash:
we tested that change on Gruhsha CTF some time ago (two years ago, huh) and removed it because some people didnt like it, so idk how it will be good for vanilla explosions (btw current keg explosion is cringe, change it pls). that change would be epic, if code for damaging tiles and another blobs outside the blobs like building, doors, shops, etc. will be greatly improved, because that mechanic is cringe and allows cases where a player can protect themselves from an explosion behind any blob, if explosive dont using a BombermanExplosion.

Vam-Jam:
> [snip] A bit unrelated, but I would really like to have a faster release cycle after the update this is included in, so that if siege does get too op some stop gap changes can be implemented. I think reworking siege before the next update would be bad since we really need to a get a release out to fix annoying bugs that have been in the game for over a year at this point... which is just ridiculous.

It would be better to discuss this in #1230

mehwaffle10:
> My only concern is that it may buff siege too much, especially since siege is significantly more cheesy in vanilla than in US captains

Mind the fact that map heights are being drastically reduced across the board and players will be able to go above max build height, which will both be significant nerfs to siege spam from towers. If anything, these changes combined would make siege feel less "cheesy" (not sure that's the best way to describe it). I would argue that instead of fearing changes due to existing balance, it should rather act as impetus for other changes that encourage healthier gameplay.

> we tested that change on Gruhsha CTF some time ago (two years ago, huh) and removed it because some people didnt like it, so idk how it will be good for vanilla explosions

I don't think the explosion implementation would matter for this change. I think it is a step in the right direction regardless of if explosions are changed or a better implementation is developed in the future. In general, invincible/heavily reinforced backwall *should not be a thing* as it causes many issues in gameplay that can be quite frustrating to deal with.

Introducing changes is always going to lead to "some people not liking it". People hate change. It's often necessary to bite the bullet and accept that you're going to receive pushback as long as the changes promote healthier gameplay.

TerminalHash:
> Introducing changes is always going to lead to "some people not liking it". People hate change. It's often necessary to bite the bullet and accept that you're going to receive pushback as long as the changes promote healthier gameplay.

Based.

PR #2459, opened 2025-11-22
[fixed] Music/Ambient music will fade out when turning off game music in the settings

### LINKED ISSUE #2458: Ambient music will often not fade out
## Description

Ambient music will often not fade out when it is supposed to.

Noticeable after applying https://github.com/transhumandesign/kag-base/pull/2453
- When approaching water and the water ambient plays, then moving away from the water, there is a 50% chance the water/underwater ambient will keep playing.
- When the ambient is playing and you set game music off, there is a chance the ambient music will keep playing

mugg91:
## Description

Fixes https://github.com/transhumandesign/kag-base/issues/2458

Previously, it would just check if the music engine is active and volume is above 0, then would enter the function `GameMusicLogic()` and early out when checking `!s_gamemusic` so it didn't fade out the music and music would keep playing until reaching the end.

Now, when turning off game music in the settings, music and ambient music will fade out immediately since `s_gamemusic` is checked beforehand.

Tested with `WARMusic.as`, should work the same with the other scripts.

### DIFF
--- a/Entities/Meta/CTFMusic.as
+++ b/Entities/Meta/CTFMusic.as
@@ -40,7 +40,7 @@ void onTick(CBlob@ this)
 	if (mixer is null)
 		return;
 
-	if (s_soundon != 0 && s_musicvolume > 0.0f)
+	if (s_soundon != 0 && s_gamemusic && s_musicvolume > 0.0f)
 	{
 		if (!this.get_bool("initialized game"))
 		{
@@ -90,7 +90,7 @@ bool gameStarted = true;
 
 void GameMusicLogic(CBlob@ this, CMixer@ mixer)
 {
-	if (mixer is null || !s_gamemusic)
+	if (mixer is null)
 		return;
 
 	if (mixer.isPlaying(world_battle))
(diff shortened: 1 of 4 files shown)


mugg91:
## Before

https://github.com/user-attachments/assets/0b84e663-497e-4ac1-a81f-e068a8ce879f

## After

https://github.com/user-attachments/assets/b095f0cd-9ce2-4147-94b4-f27373513aee

Vam-Jam (dismissed):
Tested CTF, TDM, TTH and it all seems to work, Sandbox also works but for whatever reason, the fanfare sound will play after you unmute for the first time, regardless of how long you've been in the server for. Not big enough of an issue to block the PR however

mugg91:
> (...) for whatever reason, the fanfare sound will play after you unmute for the first time, regardless of how long you've been in the server for. Not big enough of an issue to block the PR however

It's an issue unrelated to this PR as it happens in vanilla, too.
I will try to make a fix though.

mugg91:
I submitted new code but nevermind, it does not take into account match transitioning from warmup to running.
I will try to find a different solution. If I can't find a solution, I will revert the code to how it was.

mugg91:
Reverted for now as I couldn't find any easy fix to the fanfare problem.

PR #2466, opened 2025-11-29
[fixed] Scenery now gets destroyed by unpacked or spawned in buildings

### LINKED ISSUE #2464: Unpacking an outpost doesn't call destroyScenary()
## 

Grass gets removed (due to backwalls getting placed) but bushes will keep intact.
Since building a building regularly destroys plants, unpacking a building should, too.

mugg91:
## Description

Fixes https://github.com/transhumandesign/kag-base/issues/2464

This PR fixes that spawned in or unpacked buildings didn't destroy bushes.
Done by moving a `DestroyScenary()` call from `BuilderCommon.as` to `AddTilesBySector.as`.

## Testing
Move your character to a bush.
Type in `/spawn building`.
Notice the bush will still be there.
**After this PR, the bush wouldn't be there.**

Type in `/crate outpost` or buy a crate with outpost from vehicle shop.
Take the crate and move your character to a bush.
Unpack the crate inside the bush.
Notice the bush will still be there.
**After this PR, the bush wouldn't be there.**

### DIFF
--- a/Entities/Characters/Builder/BuilderCommon.as
+++ b/Entities/Characters/Builder/BuilderCommon.as
@@ -136,7 +136,6 @@ CBlob@ server_BuildBlob(CBlob@ this, BuildBlock[]@ blocks, const u32 &in index)
 
 		// take inv here instead of in onDetach
 		server_TakeRequirements(inv, b.reqs);
-		DestroyScenary(tl, br);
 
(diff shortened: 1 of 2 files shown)


Vam-Jam (approved):
Worked on staging dedi, could not spot any side effects from this change

PR #2481, opened 2025-12-04
Spectators heard lose fanfare on game over

longwoodbow:
## Description

Spectators heard lose fanfare on game over.
Now they can hear nothing like tie game.

## Steps to Test or Reproduce

1. Join game with 3 players (TDM is good)
2. Join Blue, Red team and Spectator
3. Make game over
4. Check the sound of spectator


mugg91:
It would make sense to do this since spectators aren't in the losing team.

However, with this, there won't be an audible indicator that a match has ended, e.g. if you have the game minimized as spectator.

asumagic (approved):

PR #2486, opened 2025-12-06
[added] can travel tunnel by key

mugg91:
## Description

This lets you select a tunnel in the tunnel travel menu by num key, similar to how you can do it in shops.

The keys are 1, 2, 3, 4, 5, 6, 7, 8, 9, 0, from left to right for both teams.
The originating tunnel that says "You are here" is omitted.

I didn't find a way to put a small number on the CGridButtons to quickly see the number to press to travel there. 
It would be nice. Otherwise let me know how to implement it or how to otherwise improve this.

Tested in offline, works.

## Test

Place at least 3 tunnels.
Open tunnel travel menu:
- Assuming you start from the middle tunnel, you can press 1 to go to the left tunnel or 3 to go to the right tunnel.
- Assuming you start from the left tunnel, you can press 2 to go to the middle tunnel or 3 to go to the right tunnel.
- Assuming you start from the right tunnel, you can press 1 to go to the left tunnel or 2 to go to the middle tunnel.

## Related PR

https://github.com/transhumandesign/kag-base/pull/2369

### DIFF
--- a/Entities/Industry/Tunnel/TunnelTravel.as
+++ b/Entities/Industry/Tunnel/TunnelTravel.as
@@ -99,6 +99,12 @@ void onCommand(CBlob@ this, u8 cmd, CBitStream @params)
 		if (tunnel is null) return;
 
 		client_Travel(this, caller, tunnel);
+
+		CGridMenu@ menu = getGridMenuByName(getTranslatedString("Pick tunnel to travel"));
+		if (menu !is null)
+		{
+			menu.kill = true;
+		}
 	}
 }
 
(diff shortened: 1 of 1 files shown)


mugg91:
Note: Right team can just spam 1 to go to the leftmost tunnel, while Left team needs to find the correct key first...
Perhaps the keys should be mirrored depending on team. As said, having numbers on the buttons would be nice.

mugg91:
Perhaps tunnels should get assigned a number on creation, regardless of the order they appear in in the travel menu.
That way players can figure out a tunnel's number and will remember which tunnel is which number.
It would be same behavior for both teams as well, solving the above-mentioned problem.

But I think numbers on buttons is a must.

Issue #1456, opened 2022-11-06
Slash/Jab often goes one tile beyond unwantedly


mugg91:
## Description

1) [Stand here and aim. White rectangle shows up.](https://i.imgur.com/QGhBpim.png)
2) [But a different tile, further beyond, is attacked instead.](https://imgur.com/6KM9noi.png)

Different example:
1) [Stand here and aim. White rectangle shows up.](https://imgur.com/r1NPgdF.png)
2) [Again different tile is attacked.](https://imgur.com/PcA6dMv.png)

Commonly happens when you try to attack dirt as knight.

mugg91:
This issue was introduced by build 4137-hotifx.
```
Build 3900		Dec 21 2021 	Everything still good, Knight hits nearest block
Build 4137-hotfix	Nov 04 2022	Knight hits furthest block
```

PR #2507, opened 2025-12-29
Improve map hit of knight(remake pr)

longwoodbow:
## Description

This pr is READY, but may need improve more.

Made knight hit aimed, if it's impossible hit nearest block with sword.
First, use current hit solve system to get nearest hit.
And second, use ray hit to find aimed block.

## Steps to Test or Reproduce
1. Become knight
2. Hit blocks


mugg91:
I have not tested but it looks like this should fix https://github.com/transhumandesign/kag-base/issues/1456

Vam-Jam:
Going to mark this as discussion/testing wanted. I feel like this needs sanity checking in a competitive match just to rule out accidentally changing any gameplay mechanics/bugs

PR #2515, opened 2026-01-22
[fix] duplicate arrow bolter bug

Gingerbeard5773:
## Description

Fixed a bug where 1 magazine could shoot the same arrow out of all attached bolters for the price of only 1 ammo.

This is a ported fix from ZF
https://github.com/Gingerbeard5773/Zombies_Reborn/issues/209

## Steps to Test or Reproduce

Attach two bolters to the same magazine and fire a bomb arrow out of it. With the fix, only one of the bolters will fire the arrow.

### DIFF
--- a/Entities/Structures/Components/Load/Bolter/Bolter.as
+++ b/Entities/Structures/Components/Load/Bolter/Bolter.as
@@ -33,25 +33,25 @@ class Bolter : Component
 				CBlob@ blob = blobs[i];
 				if (blob.getName() != "magazine" || !blob.getShape().isStatic()) continue;
 
-				CBlob@ ammo = blob.getInventory().getItem(0);
+				// get the next blob in inventory
+				CInventory@ inventory = blob.getInventory();
+				CBlob@ ammo;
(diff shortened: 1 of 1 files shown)


Entities/Structures/Components/Load/Bolter/Bolter.as
```diff
@@ -33,25 +33,25 @@ class Bolter : Component
...
-				CBlob@ ammo = blob.getInventory().getItem(0);
+				// get the next blob in inventory
+				CInventory@ inventory = blob.getInventory();
+				CBlob@ ammo;
+				for (u16 i = 0; i < inventory.getItemsCount(); i++)
+				{
+					CBlob@ next_ammo = inventory.getItem(i);
+					if (next_ammo.getQuantity() > 0) @ammo = next_ammo;
+				}
+
```
Gingerbeard5773, line 44: The code above looks unnecessary but its required due to an issue where items with a quantity of 0 still technically exist. Without it, not all bolters could fire arrows even if there is available ammo in the next stack.

PR #2518, opened 2026-02-15
fix config load in CTF/TTH for mods

TerminalHash:
## Description
Vanilla code does not allow loading the configuration file for CTF/TTH from mods; this fix corrects that.

The change was made to resemble similar code from TDM.as.

## Steps to test or reproduce
Change any value in config file and try to load it from mod.

### DIFF
--- a/Rules/CTF/Scripts/CTF.as
+++ b/Rules/CTF/Scripts/CTF.as
@@ -13,14 +13,10 @@
 // no scripting required!
 void Config(CTFCore@ this)
 {
-	string configstr = "Rules/CTF/ctf_vars.cfg";
-
-	if (getRules().exists("ctfconfig"))
-	{
-		configstr = getRules().get_string("ctfconfig");
-	}
+	string configstr = "ctf_vars.cfg";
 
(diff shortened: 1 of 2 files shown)


Rules/CTF/Scripts/CTF.as
```diff
@@ -9,18 +9,19 @@
...
+	CRules@ rules = getRules();
 
-	if (getRules().exists("ctfconfig"))
-	{
+	if (rules.exists("ctfconfig")) 
 		configstr = getRules().get_string("ctfconfig");
-	}
 
 	ConfigFile cfg = ConfigFile(configstr);
+	cfg.loadFile(configstr);
```
asumagic: ?
This is only in TDM.as because it does not pass the config path to the `ConfigFile` constructor. Should not have to do this?

asumagic (changes requested):

Rules/CTF/Scripts/CTF.as
TerminalHash: removed

asumagic (approved):

PR #2523, opened 2026-04-09
backstab mechanic for archer

TerminalHash:
## Description
While working with hitters, I repeatedly noticed the hitter `stab`, which used by archers only for damaging trees. I thought, why not use this and made an additional condition for players (actually for blob classes) just for lulz. Mechanic are difficult to use (requiring good archer skill or working in a pair), but the result justifies the difficulty.

It's unlikely that this will be accepted, but why not?

## Steps to Test or Reproduce
Video example of that mechanic:
https://github.com/user-attachments/assets/3529c4b7-f900-4964-823a-009b427c3829

Also it has the killfeed icon:
Снимок экрана от 2026-04-09 22-03-19ed


mugg91:
It seems this was in the game at some point? There is code for extra blood effects if a player was stabbed. Search "throat".

Vam-Jam:
It was a feature in classic, but this is slightly different to that (iirc?). We talked about it briefly in discord but this is a bit too strong.

TerminalHash:
About the conditions (compared to the old implementation):
1. Enemy should be stunned (preferably as long as possible);
2. Player should be close to enemy (very close, actually);
3. Player should aim with cursor to the enemy blob;
4. After that, player should press key_action2 (right mouse button by default).

If everything matches, the enemy will be killed, otherwise the archer risks dying himself.

TerminalHash:
If that's "unbalanced" (high risk for high reward), then I don't know, lol.

